PR #21, opened 2013-11-14
Ignore /Yu and /Yc

mellon85:
Options /Yu and /Yc are ignored instead of blocking the compilation.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -523,7 +523,7 @@ def splitCommandsFile(line):
     # since seems it difficult to set up it to correctly parse escaped quotes.
     # A good test line to split is
     # '"-IC:\\Program files\\Some library" -DX=1 -DVERSION=\\"1.0\\"
-    # -I..\\.. -I"..\\..\\lib" -DMYPATH=\\"C:\\Path\\" /Fo"C:\\A\\B\\" /Fd"C:\\A\\B.pdb"'
+    # -I..\\.. -I"..\\..\\lib" -DMYPATH=\\"C:\\Path\\"'
     i = 0
     wordStart = -1
     insideQuotes = False
@@ -532,7 +532,7 @@ def splitCommandsFile(line):
         if line[i] == ' ' and not insideQuotes and wordStart >= 0:
             result.append(extractArgument(line, wordStart, i))
             wordStart = -1
-        if line[i] == '"':
+        if line[i] == '"' and ((i == 0) or (i > 0 and line[i - 1] != '\\')): 
             insideQuotes = not insideQuotes
         if line[i] != ' ' and wordStart < 0:
             wordStart = i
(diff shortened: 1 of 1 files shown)


jdonald:
+1

I think Incredibuild used to do something similar, back when they didn't support precompiled headers. (According to Xoreax's website, they do support precompiled headers by now.)

I am using something like this workaround in my local configuration. However, as written I don't think this works reliably (independent of the merge conflict). In our CMake-generated build config, PCH is specified via `/Yu"stdafx.h"`, not two command-line arguments.

webmaster128:
I am not a fan of cutting command line arguments. Shouldn’t be the use of clcache entirely transparent to the caller? I think `cl` with and without clcache must always produce the exact same output. Otherwise you create a new compiler similar to msvc but with other behaviour that must be documented.

This might sound a bit academic if you are facing a real world productivity issue, but isn't there any change to get rid of the options in the build system?

jdonald:
What you may be suggesting is that ultimately clcache should support precompiled headers for a real. That would provide the most reliability and performance. I'm sure that would be the preferred solution, just it may not happen anytime soon. It took years for *ccache* and Xoreax Incredibuild to add support for precompiled headers.

> Otherwise you create a new compiler similar to msvc but with other behaviour

When it comes down to it *ccache* with gcc is a new compiler in corner cases, but what matters is that those minor output differences are practically benign, and documented, and the flags that enable such are opt-in.

Another approach would be to fall back to the regular compiler whenever precompiled headers are involved, which is what *ccache* did for years. It's possible that Incredibuild didn't go that route (before they supported precompiled headers) because it would make their software appear useless. The drawback as one would expect was that sometimes you would get code that would get projects that build in the IDE but not Incredibuild, or vice versa.

> but isn't there any change to get rid of the options in the build system?

It sounds like you're recommending that Visual Studio users avoid using precompiled headers. I would rather have a documented opt-in flag that go that route.

webmaster128:
> Another approach would be to fall back to the regular compiler whenever precompiled headers are involved, which is what ccache did for years.

Yes, I think that should be the case. Isn't that what clcache does right now without this PR?

> It sounds like you're recommending that Visual Studio users avoid using precompiled headers. I would rather have a documented opt-in flag that go that route.

Yes, either use precompiled headers or enjoy the features of clcache. I mean even with this PR you don't get both. You just turn off precompiled headers in the clcache processing instead of your build system. When you need a clcache option to kick out `/Yu` and `/Yc`, you are basically saying that your build system is not able to drop those in the beginning. Is that the core issue?

jdonald:
> Isn't that what clcache does right now without this PR?

clcache currently crashes and burns on precompiled headers, which I'm sure you'll see if you test it for yourself. As @mellon85 had stated:

> if those options are enabled clcache will return an error. Ignoring them will just cause the compiler to not use the PCH and do the preprocessing each time.

webmaster128:
> clcache currently crashes and burns on precompiled headers, which I'm sure you'll see if you test it for yourself.

Confirmed. I created a test and proposed a fix in #48.

PR #23, opened 2014-05-21
Handle other possible quote scenarios

jacquesg:
I've picked up 2 small niggles

1. Even though something like ```"-Iincludedir"``` gets handle correctly, ```-I"includedir"``` doesn't.
2. When a file with the compiler options is specified with an ```@``` marker the options are not always parsed correctly. The file may have its options specified on multiple lines. The solution I've come up with is to replace ```\r\n```` with a space.


webmaster128:
Could you rewrite your patch for the current master? Please add a generous amount of tests first before you change the implementation. Sample tests are in [tests.py](https://github.com/frerich/clcache/blob/master/tests.py) and can easily be extended.

frerich:
@jacquesg ping? It would be great if you could update your patch so that it applies to current master (and please also extend the tests for `extractArgument` and `splitCommandsFile` in `unittests.py`). Thanks!

frerich:
This is superseded by #122 .

PR #26, opened 2014-12-21
Support python3

cemehehko:


clcache.py
```diff
@@ -30,7 +30,12 @@
 from ctypes import windll, wintypes
 import codecs
 from collections import defaultdict, namedtuple
-import cPickle as pickle
+
+try:
+  import cPickle as pickle
+except ImportError:
+  import pickle
```
webmaster128: On my Linux system, Python 2.7.10 needs ~0.25ms to load cPickel and ~1ms to load pickel. Trying a missing import followed by a pickel import takes max. 0.1ms longer than pickel alone. Given that *cPickle can be up to 1000 times faster than pickle*, those 0.1ms are worth it I guess.

On a Windows cloud machine running Python 3, importing pickel always takes 30 ms, no matter if I checked for cPickel first or not.

Confirm [pickelimporttest.py](https://gist.github.com/webmaster128/f557a5173ab5953f49b9)

webmaster128:
LTGM, +1 for Python 3 support

PR #27, opened 2015-01-31
Fix crash when using CLCACHE_NODIRECT

McMartin:
@frerich: after hearing about `ccache`, I looked for an equivalent that could work with `MSVC`, and thanks to you I found it!

Just by using a Python Linter, I spotted a bug when using `CLCACHE_NODIRECT`, as `processNoDirect()` is actually missing its `outputFile` parameter. So this PR fixes this bug, and improves `clcache.py` by following the `PEP8` rules.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -36,7 +36,7 @@ import json
 import os
 from shutil import copyfile, rmtree
 import subprocess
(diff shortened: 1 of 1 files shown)


frerich:
Looks good, thanks! I never knew that you could write

    a not in b

PR #35, opened 2015-12-25
Ensure the destination folder exists

webmaster128:
With this change, destination folders are created.

When a destination file is e.g. `build\obj\lib\asn1_oid_lookup_default.obj` and I remove the entire directory `build\obj`, the script will now create `build\obj\lib` before copying the object.

This is very useful when you have build folders that can be deleted entirely or you build in-tree and perform a git clean (like our CI does). Necessary for re-building Botan.


frerich:
Looks good, thanks!

PR #36, opened 2015-12-26
Set default cache size to 1 GiB

webmaster128:
This helps readers to understand the default value of the cache size in the README


frerich:
I guess the old default was a little lame, it was neither 10^9 nor 2^30 but rather 2^20 * 10^3. So this seems like a sane improvement to me. Thanks!

PR #39, opened 2015-12-27
Make clcache compatible with Python 3

webmaster128:
This PR makes clcache compatible with Python 2.7 and Python 3.

The changes are minimal and can be considered good style even for Python 2. Most work was done using the 2to3 tool (fixing `print` statements and stuff like `dict.items()`).

For the `wintypes.c_int` there is a public alias which is can be used in both Python 2 and Python 3.

Some changes are required because Python 3 strictly differentiates between unicode strings and byte arrays. E.g. string must be encoded before they can be hashed.

I change the fallback of command line file encoding to UTF-8, which is used in my JOM+CMake setup. I think it is a better fallback than just pass raw bytes because it handles UTF-8 and ASCII files correctly. 

Credit goes to @cemehehko for the pickel block.

Tested using Python 2.7 and Python 3.4

Closes #26

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -3,6 +3,7 @@
 # clcache.py - a compiler cache for Microsoft Visual Studio
 #
 # Copyright (c) 2010, 2011, 2012, 2013 froglogic GmbH <raabe@froglogic.com>
+#               2016 Simon Warta (Kullo GmbH)
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
@@ -27,10 +28,18 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
+
+# All string literals are unicode strings. Requires Python 3.3+
+# (http://python-future.org/faq.html#which-versions-of-python-does-python-future-support)
+from __future__ import unicode_literals
+
 from ctypes import windll, wintypes
 import codecs
 from collections import defaultdict, namedtuple
-import cPickle as pickle
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
 import errno
 import hashlib
 import json
@@ -80,9 +89,9 @@ class ObjectCacheLock:
     def __init__(self, mutexName, timeoutMs):
         mutexName = 'Local\\' + mutexName
         self._mutex = windll.kernel32.CreateMutexW(
-            wintypes.c_int(0),
-            wintypes.c_int(0),
-            unicode(mutexName))
+            wintypes.INT(0),
+            wintypes.INT(0),
+            mutexName)
         self._timeoutMs = timeoutMs
         self._acquired = False
         assert self._mutex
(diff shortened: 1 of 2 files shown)


clcache.py
```diff
@@ -177,7 +180,7 @@ def removeObjects(self, stats, removedObjects):
                 rmtree(dirPath)
             stats.setCacheSize(currentSize)
 
-    def getManifestHash(self, compilerBinary, commandLine, sourceFile):
+    def getManifestHash(self, compilerBinary, commandLine, sourceFilePath):
```
frerich: Why is this argument renaming part of the commit - is `sourceFile` some sort of identifier in Python 3.x?

clcache.py
```diff
@@ -448,15 +451,18 @@ def save(self):
...
 
 
-def getFileHash(filePath, additionalData=None):
+def getFileHash(filePath, additionalDataString=None):
     hasher = HASH_ALGORITHM()
     with open(filePath, 'rb') as inFile:
         hasher.update(inFile.read())
-    if additionalData is not None:
-        hasher.update(additionalData)
+    if additionalDataString is not None:
```
frerich: I'm not really fan of hungarian notation, I'd rather not merge this renaming work (which doesn't seem to be needed to fulfill the purpose of this commit, implementing support for Python 2.7 and Python 3.3).

clcache.py
```diff
@@ -448,15 +451,18 @@ def save(self):
...
+        # Encoding of this additional data does not really matter
+        # as long as we keep it fixed, otherwise hashes change.
+        # The string should fit into ASCII, so UTF8 should not change anything
+        hasher.update(additionalDataString.encode("UTF-8"))
```
frerich: I don't understand this part - how does encoding the data as UTF-8 improve anything as far as the hash is concerned? Can you give an example?

clcache.py
webmaster128: You're right, the renaming is not necessary at all. I just found it helping since we are dealing with a lot of "file" variables within this Python script. A file may refer to a [file handle](https://docs.python.org/2/library/stdtypes.html#bltin-file-objects) when reading, the file's content in binary format, the file's content as a unicode string or the path to a file. Pythons type system does not make it easy to differentiate those on the first glance.

But at the end of the day, it's a matter of taste. Just tell me which one you would like to have :)

clcache.py
webmaster128: `additionalData` is a string, i.e. a list of unicode code points. Those unicode strings are not a sequence of bytes but the hasher can only consume byte sequences. There are multiple ways to "encode" a unicode string into a sequence of bytes, for example latin1, UTF-8 or UTF-16. All of them will generate different binary representations for the same string. Python 2 does not strictly differentiate between unicode strings and byte sequences, but Python 3 does.

Without explicitly encoding, Python 3 tell us exactly that:

```
File "[...]\clcache.py", line 465, in getFileHash
    hasher.update(additionalDataString)
TypeError: Unicode-objects must be encoded before hashing
```

clcache.py
webmaster128: Right, in this case it is not helping as much as in the file case. I'll revert this renaming. But let me know what you think of the file case above.

webmaster128:
Reverted all unnecessary renamings, added Python 3 support for `CLCACHE_NODIRECT=1`, added myself as contributor, specified minimal version of the Python 3.x branch.

frerich:
Thanks! It appears the AppVeyor hook works. :-)

frerich:
In particular, thank you for taking the time to clean up the patch so that it's only the Python 3 specific stuff! Alas, I don't get around to hacking clcache as much as I'd like to, so I really depend on a clean history. :-)

webmaster128:
Sure, thanks for review, feedback and merge. I'll try to keep pull requests as small and specific as possible.

PR #40, opened 2016-01-03
Add initial AppVeyor syntax check for multiple Python versions

webmaster128:
This AppVeyor configuration allows us to perform a basic syntax check using all supported Python versions by running `clcache.py --help` and `clcache.py -s`.

More sophisticated testing like compiling a sample file could be added later on.

Sample results testing my branch: https://ci.appveyor.com/project/webmaster128/clcache

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,26 @@
+environment:
+  matrix:
+    # AppVeyor installed Python versions
+    # http://www.appveyor.com/docs/installed-software#python
+
+    - PYTHON_INSTALL: "C:\\Python27"
+    - PYTHON_INSTALL: "C:\\Python33"
+    - PYTHON_INSTALL: "C:\\Python34"
+    - PYTHON_INSTALL: "C:\\Python35"
+
+install:
+  # Make compiler available (use MSVC 2013, 32 bit)
+  - call "%ProgramFiles(x86)%\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" x86
+
+  # Check compiler version
+  - cl
+
+  # Prepend Python installation to PATH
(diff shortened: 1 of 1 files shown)


frerich:
I guess I'm not quite up to date here - I See that AppVeyor is some dort of
CI system. What dies it do for US, where can one See results?
Am 03.01.2016 9:43 nachm. schrieb "Simon Warta" <notifications@github.com>:

webmaster128:
Right, AppVeyor is a CI system, like Travis or CircleCi. In contrast to the last two, it runs Windows machines.

On a CI system, you can run all kinds of tasks and get green light when all commands returned with exit status 0 and red light otherwise. The advantage for us is that it can test multiple system configurations at once. When for example a Python 3 developer makes a change that is not compatible with Python 2, he does not recognice it. The CI does. Later we can add direct/nodirect mode to test even more code paths.

The cool thing about CI systems with Github integration is that pull requests are checked before merge. So part of the review is done automatically without anyone looking at the patch. Check the Botan PR list for visualization of CI status of all pull requests: https://github.com/randombit/botan/pulls

I added a result page to my original description (thus not in your notification email): https://ci.appveyor.com/project/webmaster128/clcache You can see that the current syntax is not compatible with Python 3, which is the expected result. As soon as my other patch is in, everything should get green.

frerich:
Interesting, let's give it a try! Thanks a lot!

webmaster128:
Cool, thanks. You (as a repository owner) now need to go to https://ci.appveyor.com/, login with your Github account and add clcache as a project.

webmaster128:
Great. Check https://ci.appveyor.com/project/frerich/clcache/history: master is red as expected and my PR branch is green.

PR #42, opened 2016-01-04
Run tests in CI

webmaster128:
We already have a test framework, great! Let's run it in the CI.


frerich:
Thanks! Yeah, our tests are not much but I guess you have to start somewhere. The command line splitting on Windows is notoriously difficult.

PR #44, opened 2016-01-20
Add initial compile tests

webmaster128:
This is a basic version of testing the compilation through clcache, starting with a simple C and a simple Cpp program.

I need the CI to work on that. Feel free to comment at any point but please don't merge. I'll add a comment when I am done.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,3 +27,8 @@ build_script:
 
 test_script:
   - "python tests.py"
+
+after_test:
+  # Get a rough overview about what happened in the tests.
+  # This is not part of the actual testing but pure convenience.
+  - "python clcache.py -s"
--- a/tests.py
+++ b/tests.py
@@ -1,5 +1,7 @@
 import unittest
 import clcache
+import subprocess
+import sys
 
 class TestSplitCommandsFile(unittest.TestCase):
     def _genericTest(self, fileContents, expectedOutput):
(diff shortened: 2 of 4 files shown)


frerich:
The man after which the famous sequence is named was called Fibonacci. ;-)

webmaster128:
Good point, thanks

webmaster128:
Okay, I am happy with this for now. This already covers a lot of additional clcache codepaths. Next project would be to test for cache hits, but that is another topic.

frerich:
Thanks, this looks interesting! A couple of questions:

1. Can you please mention (in the commit log message) why you added 'python clcache.py -s' as an after-test step)?
2. I think it would be nice to make the tests perform some 'tear down', too - they should at least delete any generated files (e.g. .obj and .exe files), no?
3. Instead of assuming that `python` is in the Path, it seems to me that using `sys.executable` would be better (this gives the path to the current Python interpreter). I suspect that this will fix running the tests when not having `python` in the path.

frerich:
Also, I wonder - are there any nice code coverage solutions for Python? I suppose ti would be good to set that up early to be able to tell what needs testing and whether any two tests have significant overlap.

webmaster128:
Alright:

1. I added a comment in the config file itself to make that clear for any future reader.
2. Best practice is to leave test artefacts untouchted in order to be able to better debug testing. But it might be useful to have a specific test output directory the can be gitignored. I'd leave it as it is now and find a solution that works well when more artefacts are added into the mix.
3. Good point, applied

> Also, I wonder - are there any nice code coverage solutions for Python?

I don't know of any but I am not a Python expert. For now I see a lot of low hanging fruits and existing bugs that can be tested away before we really need coverage.

Issue #31, opened 2015-04-11
clcache.py can not deal with args with double \


billy0920:
when dealing with some args like /Fo"Release\\", clcache.py will faild to regnizing the 2nd quote.

jdonald:
This is the first roadblock I ran into when trying to build our project with clcache. We're using CMake-generated Visual Studio 2013 proj/sln files. The closing doublequote gets misinterpreted as an escaped doublequote, then argument parsing is completely broken after that point.

Oddly the `splitCommandsFile` description says we should handle this case :
```
    # A good test line to split is
    # '"-IC:\\Program files\\Some library" -DX=1 -DVERSION=\\"1.0\\"
    # -I..\\.. -I"..\\..\\lib" -DMYPATH=\\"C:\\Path\\"'
```

(i.e. the C:\Path\ dir has a trailing slash). Yet the logic that follows looks like it's trying to implement UNIX-style doublequote escaping :

```
        if line[i] == '"' and ((i == 0) or (i > 0 and line[i - 1] != '\\')):
            insideQuotes = not insideQuotes
```

So my local fix was to remove that backslash-check, and simply predicate on `if line[i] == '"':` for inverting insideQuotes.

It looks like this code was intended for the rare case of command-line macros that require escaped doublequotes, like `-DVERSION=\\"1.0\\"`, should a project be unfortunate to have those. However, the rules for escaping doublequotes in the Windows command prompt seem to be [complicated](http://stackoverflow.com/questions/7760545/cmd-escape-double-quotes-in-parameter) and nothing like bash, so I don't think this code as written works. I likewise had to remove the `replace('\\"', '"')` step of `extractArgument`.

vchigrin:
@jdonald, Part of splitCommandsFile code was written by me, sorry if it does not work on your project. Handling things like ```-DVERSION=\\"1.0\\"``` really should be kept - I am certainly can remember that this issue were met during building some versions of chromium browser. I am not sure about case
```-DMYPATH=\\"C:\\Path\\"'```, though, we still use clcache on our project and it works fine... 
Could you attach example of command-line that cause wrong behavior of splitCommandsFile ? Is it something like ``` /Fo"Release\"```, as mentioned above?

**Comment by jdonald** [comment]
> Could you attach example of command-line that cause wrong behavior of splitCommandsFile

We have pretty long command lines, but here's the relevant part :
```
[...]

PR #45, opened 2016-01-21
Improve command line argument handling

webmaster128:
* use argparse library for cleaner argument parsing
* auto-generate help/usage message
* Force integer value for `-M`
* Add aliases `--show-stats`, `--zero-stats`, `--max-size` known from ccache
* Block values less or equal to 0 for `-M`

This is work in progress, which should not be merged because I found another bug that should be handled first.


webmaster128:
Rebased onto current master. I am happy with this for now, but I guess there is room for comments.

frerich:
The command line parsing is really a tricky part of `clcache.py`. There have been numerous patches back and forth affecting this. And there's still one issue open about this, #31 .

If I remember correctly, one source of complexity was the fact that the optional command file (`@foo`) is parsed with slightly different rules than what's documented on e.g. https://msdn.microsoft.com/en-us/library/17w5ykft.aspx

frerich:
One particular case I remember which excluded using any kind of ready-made command line parsing library: an argument like `/Fo"Release\"`.

webmaster128:
The idea here is to handle clcache arguments only. Everything else is passed into `compiler_args`.

But unfortunately, this does only work as long as cl.exe arguments do not start with `-`. Damn.

PR #46, opened 2016-01-21
Fix bug when object file was specified without directory

webmaster128:
Recompiling the same file currently fails on my machine, so let's add a test for that


webmaster128:
Okay, this finds and fixes a bug. Ready for merge.

frerich:
Thanks! Odd that this didn't get hit before...

webmaster128:
I introduced the bug myself in c6914de5d33a25b924388b06c02327c

PR #47, opened 2016-01-21
Random thoughts on #31

webmaster128:
This removed the replacement from `\"` to `"` in extractArgument, which should solve #31.

Note that Python strings having the prefix `r` no not use the backslash notation to escape a special char, so `r'\t'` is `\` and `t`, not a tab. Github messes up with syntax highlighting here.


frerich:
It's a bit late (or early) where I am, so I hope I didn't misread anything -- but the tests all seem sane. Thanks!

PR #48, opened 2016-01-22
Test precompiled headers with clcache

webmaster128:
This passes all calls including `/Yc` and `/Yu` to the real compiler, allowing the use of clcache in a mixed environment, where compilation units without precompiled headers use the cache and compilation units with precompiled headers don't.

I was able to reproduce the bug described in #21 using a `nmake`, `nmake clean`, `nmake` on the sample project. Only the second nmake build failed because the .pch file was missing. This is because clcache used the object from cache, which did not restore the .pch file. With the patch in place, the test passes.

Closes #21

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -681,7 +681,7 @@ def analyzeCommandLine(cmdline):
     # copy the generated .pdb files into/out of the cache.
     if 'Zi' in options:
         return AnalysisResult.ExternalDebugInfo, None, None
-    if 'Yu' in options:
+    if 'Yc' in options or 'Yu' in options:
         return AnalysisResult.CalledWithPch, None, None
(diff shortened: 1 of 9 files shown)


frerich:
Can you please split this commit, or make the commit log more descriptive about what's going on? 161 added lines and 9 changed lines is a pretty hefty change for the commit log message 'Test precompiled headers with clcache'.

webmaster128:
Jap. I'll update and rebase this as soon as #49 is in. There might be merge conflicts in tests.py because I sorted the modules alphabetically.

webmaster128:
Check. Rebased and updates the commit message. In this case everything belongs together (bugfix and a small test project).

Issue #32, opened 2015-04-11
clcache.py run too many jobs 


billy0920:
when the cl.exe with many source files, clcache.py make one process for one source files. I get a project that call cl.exe with more than 2000 source files, and in that case, clcache.py run more than 2000 processes and the memory run out.

jdonald:
I see about 7 MB per clcache py2exe process. When cl.exe is invoked with multiple source files it does spawn itself for each one. This could be made more efficient by spawning a thread rather than a new process for each source.

But in my dev environment, cl.exe possibly uses more than 7 MB per source file as a baseline, so this isn't a dealbreaker. Even when the codebase is thousands of files, Visual Studio or nmake does not run run all those compilations at once anyhow. If it did, it would run out of memory even without clcache.

What I have seen when debugging parsing issues is that I was able to mess up the logic such that it did an infinite recursion of cl.exe calling itself. If such a bug happens, yes one can lock up the machine very quickly. You can identify this case by running with CLCACHE_LOG and see if it happens to report "multiple source files" without ever converging.

frerich:
Maybe we should indeed not fork one clcache program per source file but rather maintain a pool of processes. I imagine {{cl.exe}} itself has some (configurable?) limit on how many files to build in parallel. The same limit could be used to decide on a maximum number of parallel {{clcache}} processes.

webmaster128:
Why would you call a compiler with multiple source files? Doesn't every build system on earth generate one compiler command per source file?

frerich:
Visual Studio itself or Qt's "qmake" both call "cl" with multiple source
files if possible.

2016-01-22 9:42 GMT+01:00 Simon Warta <notifications@github.com>:

webmaster128:
In the README we have the restriction

> There must be exactly one source file present on the command line.

which differs from what the code tries to do. Is the documentation outdated?

frerich:
Indeed, it is - I overlooked that the patch which introduced support for
multiple source files (somebody contributed that feature) didn't update the
README file. :-/

2016-01-23 12:04 GMT+01:00 Simon Warta <notifications@github.com>:

webmaster128:
Works fine for me:

* Compile 500 files using `clcache.bat /c *.cpp` runs 1 process
* Compile 500 files using `clcache.bat /c /MP *.cpp` runs 4 process (on my 4 cores)
* Compile 500 files using `clcache.bat /c /MP3 *.cpp` runs 3 process

[...]

PR #49, opened 2016-01-23
Fix /MP[processMax] parsing

webmaster128:
Fixes a bug in the regular expression for `/MP[processMax]`. Before the integer was not optional, so `/MP` was not interpreted as an mp switch and thus

```
    # /MP, but no count specified; use CPU count
    try:
        return multiprocessing.cpu_count()
    except:
        # not expected to happen
        return 2
```

was dead code.

This renames `mp_switch` to `mp_switches` in the implementation because it is a list of switches and should not be confused with the chosen switch `mp_switch`.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -797,7 +797,7 @@ def jobCount(cmdLine):
 
     switches.extend(cmdLine)
 
-    mp_switch = [switch for switch in switches if re.search(r'^/MP\d+$', switch) is not None]
+    mp_switch = [switch for switch in switches if re.search(r'^/MP(\d+)?$', switch) is not None]
     if len(mp_switch) == 0:
         return 1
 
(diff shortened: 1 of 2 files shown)


webmaster128:
Okay I think this can be merged. This fixes a bug in testJobCount(), so it is somehow related to #32. But I don't think it the the core of the issue. That would only be the case if the user called `cl /c /MP2000 /MP sources...`, which is unlikely.

frerich:
Thanks! However, it seems the commit should maybe be split into two, or get a more elaborate log message. As it is, it almost seems unintentional that `clcache.py` got modified because the log just talks about adding tests. :-}

I think in general it would be best to have explanatory texts such as your first comment right in the Git history, so that one does not need to check GitHub for *why* something was done.

webmaster128:
Thanks for review. You're right. The original patch only added tests but than I found the bug and did not update the commit message.

Now we have bugfix+tests and then the renaming.

frerich:
Looks nice, thanks!

PR #51, opened 2016-01-26
Move PYTHON_BINARY and CLCACHE_SCRIPT to per file scope

webmaster128:


frerich:
Thanks! I was about to ask 'Why?' but I suppose the diff speaks for itself here.

PR #52, opened 2016-01-27
Add initial hit test

webmaster128:
This adds a first test that ensures that an object which should be in the cache is hit when compiled again.


frerich:
Don't worry. I think the fact that neither you nor me knows the semantics of `with` exactly is a good argument against trying to be clever. :-]

PR #53, opened 2016-01-27
Add test for changes in header

webmaster128:
This is WIP. We discovered that in direct mode, changing header files are not detected properly. Let's find out what is going on here.


frerich:
I'm curious: who is 'we'? I'm always interested in hearing about use cases for clcache. :-)

webmaster128:
We are [Kullo – Secure Communication](https://www.kullo.net/)

webmaster128:
Okay I found the bug by reviewing the code. I'll update this until I have a test setup to cover the bug and fix it then.

frerich:
Thanks!

PR #54, opened 2016-01-27
Replace command 'rm' by 'del'

webmaster128:
'rm' is not a Windows command. It works in the CI but not in a usual
developer console.


frerich:
Thanks!

PR #55, opened 2016-01-27
Cleanup tests

webmaster128:
2d1e5c4e6e134a (Github #52) and 12cc496bbb (Github #51) bite each other
a bit. This cleans it up.


frerich:
Thanks for taking care of this!

PR #56, opened 2016-01-27
Rename parseIncludesList() and add tests

webmaster128:
WIP. Let's see if this works as expected ...


webmaster128:
Alright, looks like there is no bug in here. So I am happy with this now: some minor refactorings and a test.

webmaster128:
So, I hope I got everything. Anything else?

frerich:
The only thing left would be a glass of $FAVORITE_BEVERAGE to celebrate the merge, I guess. ;-)

webmaster128:
Fine. $FAVORITE_BEVERAGE comes with after fixing this direct mode recompile bug, I am hunting down with all of those current PRs.

PR #57, opened 2016-01-27
Use some methods in a static way

webmaster128:
The machods getManifestHash(), computeKey(), getKeyInManifest(),
getDirectCacheKey() never touch the cache instance, so they can me made
static. Calling then in a static way makes the caller code much more
readable, since it is clear where ther state lives.

Move getHash() from global scope to ObjectCache where all its callers
are as well.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -187,7 +187,8 @@ class ObjectCache:
                 rmtree(dirPath)
             stats.setCacheSize(currentSize)
 
-    def getManifestHash(self, compilerBinary, commandLine, sourceFile):
+    @staticmethod
+    def getManifestHash(compilerBinary, commandLine, sourceFile):
         stat = os.stat(compilerBinary)
         # NOTE: We intentionally do not normalize command line to include
         # preprocessor options. In direct mode we do not perform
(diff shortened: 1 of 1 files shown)


webmaster128:
A lot of python people say that the better way is to have those static functions in the global scope. Python 2.7 does not even support static functions without this magic decorator.

Ideally one could move them into a module. But I don't think that can be done without adding a second file, which I want to avoid.

So the question is, do those functions really belong to ObjectCache or should they go into global scope.

frerich:
My understanding is that static functions (aka "class methods") are only syntactically different from global free functions.

I'm a functional programmer at heart (I'm a big fan of Haskell) and as such, to me, there is very little difference between passing the `self` (or `this`, or whatever) argument implicitly (as is the case with object methods) or explicitly (as would be the case for free functions).

In fact, I argue that the latter provides better scoping since it can only access global identifiers and its arguments (as opposed to an object method, which can also access private variables & methods of the class it is associated with).

So I very much agree with decoupling code from classes, but I have no strong feeling on whether to use a 'staticmethod' in Python or a free function.

PR #58, opened 2016-01-27
 Fix recompile bug in direct mode when header changed

webmaster128:
This does not yet include a fix, but I am going to add one.

The issue is that `processNoManifestMiss()` does not add `/showIncludes` to a given command line, so `listOfIncludes` is always empty and thus the header content is never checked in direct mode.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1010,8 +1010,13 @@ def processHeaderChangedMiss(stats, cache, outputFile, manifest, manifestHash, k
 
 def processNoManifestMiss(stats, cache, outputFile, manifestHash, baseDir, compiler, cmdLine, sourceFile):
     stats.registerSourceChangedMiss()
-    stripIncludes = '/showIncludes' not in cmdLine
-    returnCode, compilerOutput, compilerStderr = invokeRealCompiler(compiler, cmdLine, captureOutput=True)
+    if '/showIncludes' in cmdLine:
+        realCompilerCmdLine = cmdLine
+        stripIncludes = False
+    else:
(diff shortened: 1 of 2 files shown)


webmaster128:
Oh yeah – here we go!

frerich:
Good catch! Thanks for tracing this down - the commits seem fine to me, except for the first one which actually implements the fix (but I just added my thoughts on that).

webmaster128:
In the last commit (bugfix): I was avoiding the one-liner if statement because it is too complex for the *conditional* appending.

Question is, if *unconditional* appending should be used. As far as I can see, the `realCompilerCmdLine` is not visible anywhere. It is constructed after the `manifestHash`, which includes the user's command line has been fixed. It is also not content of the manifest itself because there only the list of includes is stored.

So I guess as long as cl.exe excepts multiple `/showIncludes`, this is a matter of taste. I don't find any argument pro or contra.

frerich:
No problem. I don't see any technical issue with your commit, and whether or not a particular code is "simpler" than some other version is of course somewhat subjective. I just wanted to suggest an alternative definition with a smaller diff, but I don't really mind your version either.

frerich:
Sorry, my finger is somewhat trigger-happy today.

webmaster128:
Great, thanks! Thanks for quick and precise code review.

Now it is really time for a glass of `$FAVORITE_BEVERAGE`!

I guess this is worth a bugfix release if you are planing to have something like releases in the future. For Kullo, releases do not matter since I work with a local git checkout. Is this py2exe thing still maintained? I guess clcache binary releases would be a good idea to attract new users – just a random thought.

PR #61, opened 2016-01-30
Improve versioning

webmaster128:
This makes clcache versioning clearer, adds a changelog and uses the clcache version as part of the compiler hash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,19 @@
+clcache changelog
+=================
+
+## clcache 3.0.2 (2016-01-29)
+
+ * Python 3 compatibility
+ * Add new env variable to control clcache lock timeout
+ * Bugfix: Fix recompile bug in direct mode when header changed
+ * Bugfix: Fix compile error when clcache is used with precompiled headers
+ * Bugfix: `/MP[processMax]` was not properly parsed when `processMax` was unset
+ * Bugfix: Properly handle arguments including `\"` (e.g. `/Fo"Release\"`)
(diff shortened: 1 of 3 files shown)


frerich:
I guess this can go in - your patches clearly improve things, and I'd rather not keep them out only because there *maybe* are even better approaches.

Crazy experiments with automatically generated changelogs can still be pursued in the future, should we feel the need for it.

Now I just need to make a habit of reminding people to extend the ChangeLog when mergeing pull requests. Oops. :-}

webmaster128:
Thanks for merging this. To answer the open comment related to cache invalidation, which is a non-trivial topic worth discussing:

I think it is a good mentality to say: it is better to invalidate the cache a thousand times than restore a broken object once. This is because we cannot estimate which impact broken executable code has – could be a security issue deployed to millions of people.

I encountered the problem when I compiled the code of a compilation unit, which had a broken manifest file in the cache (stored from clcache 3.0.1 and restored by clcache 3.0.2). My I case I was lucky and got a compile error. But than I had no chance to clear the cache via any command line interface so I had to manually remove the cache files from the hard disk. 

With the release of 3.0.2 we did not see such situations coming and I think it is very hard to decide which bugfix would need to manually increment a CACHE_REVISION_NUMBER in order to invalidate what is in the cache.

At Kullo we need the cache because the CI builds our software multiple times a day and clears the build directory in between. It would be perfectly alright for us to clear the cache once a week. Given that the expected release time is somewhere between 2 years and 1 month, I don't think it hurts anyone.

PR #63, opened 2016-02-06
support for py2exe on python-3.4

timblechmann:


frerich:
Thanks! I was worried for a moment that making `myExecutablePath` a Unicode string even with Python 2.7, but since the CI system doesn't complain I guess it doesn't harm. :-)

timblechmann:
@frerich well, unfortunately i had to go back to py2exe/python-2.7 due to some character decoding issues, probably related to compiler output in german locale :(

PR #73, opened 2016-04-11
output files: compress compiler output via gzip

timblechmann:
(still untested)

compiler output is rather verbose and can be compressed very well. the output files are now also properly flushed to disk while still holding the lock

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -253,10 +253,10 @@ class ObjectCache:
                 copyOrLink(objectFileName, self.cachedObjectName(key))
 
             if cfg.compressedOutput():
-                with GzipFile(self._cachedCompilerCompressedOutputName(key), 'w') as outFile:
+                with GzipFile(self._cachedCompilerCompressedOutputName(key), 'w', 1) as outFile:
                     outFile.write(compilerOutput)
                 if compilerStderr != '':
-                    with GzipFile(self._cachedCompilerCompressedStderrName(key), 'w') as outFile:
+                    with GzipFile(self._cachedCompilerCompressedStderrName(key), 'w', 1) as outFile:
                         outFile.write(compilerStderr)
             else:
                 with open(self._cachedCompilerOutputName( key ), 'w') as outFile:
--- a/clcache.py
+++ b/clcache.py
@@ -242,20 +242,28 @@ class ObjectCache:
     def hasEntry(self, key):
         with self.lock:
             return os.path.exists(self.cachedObjectName(key)) \
-                   or os.path.exists(self._cachedCompilerOutputName(key)) \
-                   or os.path.exists(self._cachedCompilerLegacyOutputName(key))
+                   or os.path.exists(self._cachedCompilerCompressedOutputName(key)) \
+                   or os.path.exists(self._cachedCompilerOutputName(key))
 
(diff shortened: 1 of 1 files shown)


webmaster128:
How does this migrate existing caches where output.txt is not compressed?

timblechmann:
it doesn't ... do you think that a fallback mechanism is necessary?

webmaster128:
Reading an existing object from cache and having no compiler output leads at least to incorrect data or worse to exceptions because `output.txt` is missing.

In general I don't think it's worth adding extra complexity here because `output.txt` make up very little space from the entire cache. In my case this is 125513 bytes `output.txt` 984607894 bytes `object`, i.e. 0,012 % output.txt in the entire cache.

Scripts to reproduce run in the bash that comes with Git for Windows:

```
find . -name "output.txt" -print0 | du --files0-from=- -bc
find . -name "object" -print0 | du --files0-from=- -bc
```

timblechmann:
this is quite different for my codebase: 1.640.260.694 of `output.txt` compared to 41.548.668.525` of objects, which is about 4% of compiler output

webmaster128:
Is this because objects are smaller or because STDOUT is larger? All the `output.txt`s I opened contained not more than the source filename, which makes these files bigger when using gzip.

Could you add an example output?

frerich:
I suppose it's not too hard to imagine C++ programs making heavy use of C++ templates which cause hideous amounts of warnings. Maybe that's it?

timblechmann:
no warnings, just thousands of lines with include paths, probably because of compiling via `/showIncludes`

timblechmann:
some numbers: 

* output.txt: 474429 byte
* output.txt.gz: 25119 byte

42xx lines of: `Hinweis: Einlesen der Datei: ` with include path ...

frerich:
I can imagine that this switch may make sense in some cases, but `clcache` in general values speed over disk space. It happens that in some cases this goes hand in hand (e.g. reading/writing less files is faster), but in this particular case I'm not sure about it.

A couple of benchmarks would be in order, but my gut feeling is that this functionality should be configurable via a switch which defaults to `off`. We may change the default of this switch in the future once it was shown that the reduced disk space usage is a bigger gain than the (hopefully small) decrease in runtime.

timblechmann:
hmm, according to many filesystem benchmarks, transparent compression often outperforms un-compressed file access, since fewer data needs to be transferred from the device and the throughput of lightweight compression is larger than disk throughput. not sure how filesystem caching would benefit form it on windows, though. so compression might actually improve performance.

frerich:
Sure, it might improve performance. Building a couple of typical projects with and without the feature would surely help judging the effect.

For what it's worth, I also found that compression did not actually help since the hard disk driver was reading in larger chunks anyway - so even requesting 8k data would end up reading 256k or so. I wouldn't be surprised if performance wise, it's equal (which would be a good argument to include the feature because it would reduce disk space usage without having a significant performance impact).

Another thing which might be relevant here is that if you're dealing with small files, you may be able hold more files in the file system caches at once (which may have a noticeable effect).

Anyway, it's all somewhat hand wavy at this point, I guess

frerich:
Closing this; the reduction in disk size might be nice, but there does not seem to be a clear improvement in runtime speed (which is what clcache is all about), so I think this idea can be dropped for now.

Issue #70, opened 2016-04-10
cache directory and stats go out of sync


timblechmann:
in my configuration, the size of the cache dir is configured to be 200GB, but after some time, the cache directory exceeds the stats.txt file: the file reports something like 160GB usage, but the real disk utilisation is about 1TB.

timblechmann:
one possible reason for the mismatch: the `output.txt` files are not taken into account for computing the cache size (some 100kb per translation unit in my codebase)

webmaster128:
Confirmed: my 400 MB cache takes 950 MB on disk.

Given that `output.txt` is typically 1/50 to 1/1000 the size of the actual object, I think there must be something bigger going wrong.

webmaster128:
Can that happen from running multiple clcache processes? E.g.

1. clcache.py A is called and reads stats.txt with `CacheSize = 1000`
1. clcache.py B is called and reads stats.txt with `CacheSize = 1000`
1. clcache.py A starts the real compiler to compile a new object of 100 bytes
1. clcache.py A writes `CacheSize = 1100` to disk
1. clcache.py B starts the real compiler to compile a new object of 200 bytes
1. clcache.py B writes `CacheSize = 1200` to disk

Now the cache size in stats.txt is lower than the file size on disk.

frerich:
It seems to me that the `Configuration` class is inherently racy in its API: it does use a lock file to avoid that there are concurrent read/write accesses to the file, but since the API is defined in terms of getters and setters, there's always the possibility that the configuration is changed on disk without a second `Configuration` object noticing the change (and thus overwriting it when saving). So right now, the burden of maintaining consistency is on the caller.

This is not even specific to the `CacheSize` field, the same race condition would probably manifest when updating e.g. `CallsForLinking` or `CallsWithoutSourceFile`.

It would be possible to change the `Configuration` API such that it is not defined in terms of getter/setters but rather `update` methods (which atomically change a value). However, that's probably going to be inefficient (and certainly incorrect!) when updating multiple values, because the code would acquire and release the lock twice when updating e.g. two fields.

webmaster128:
I guess the most reliable solution here is to use a SQLite database, where we have one connection per clcache.py process. To this database we can send updates like (pseudo code) `CacheSize += 200` or `CallsForLinking += 1`. These are atomic operations that SQLite is able to serialize. This avoids any locks in clcache.

[...]

PR #74, opened 2016-04-14
WIP: stats: read and modify stats only when holding the cache lock

timblechmann:
* untested
* we might want to separate locks for stats and cache

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -148,29 +148,28 @@ class ObjectCache:
         return self.dir
 
     def clean(self, stats, maximumSize):
-        with self.lock:
-            currentSize = stats.currentCacheSize()
-            if currentSize < maximumSize:
-                return
+        currentSize = stats.currentCacheSize()
+        if currentSize < maximumSize:
+            return
 
-            # Free at least 10% to avoid cleaning up too often which
-            # is a big performance hit with large caches.
-            effectiveMaximumSize = maximumSize * 0.9
+        # Free at least 10% to avoid cleaning up too often which
+        # is a big performance hit with large caches.
+        effectiveMaximumSize = maximumSize * 0.9
 
-            objects = [os.path.join(root, "object")
-                       for root, folder, files in os.walk(self.objectsDir)
-                       if "object" in files]
+        objects = [os.path.join(root, "object")
+                   for root, folder, files in os.walk(self.objectsDir)
+                   if "object" in files]
 
-            objectInfos = [(os.stat(fn), fn) for fn in objects]
-            objectInfos.sort(key=lambda t: t[0].st_atime)
+        objectInfos = [(os.stat(fn), fn) for fn in objects]
+        objectInfos.sort(key=lambda t: t[0].st_atime)
 
-            for stat, fn in objectInfos:
-                rmtree(os.path.split(fn)[0])
-                currentSize -= stat.st_size
-                if currentSize < effectiveMaximumSize:
-                    break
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -148,29 +148,28 @@ def cacheDirectory(self):
         return self.dir
 
     def clean(self, stats, maximumSize):
-        with self.lock:
-            currentSize = stats.currentCacheSize()
-            if currentSize < maximumSize:
-                return
+        currentSize = stats.currentCacheSize()
```
webmaster128: Why did you pull this outside the lock? currentCacheSize() is a read operation that should be protected.

timblechmann: `clean` is called under the cache lock (i've guarded each top-level instant of `stats` with the lock)

webmaster128:
I don't think that this is going to end up in a robust solution as it is now because every caller has to take care to properly lock the stats in places where you want to have a higher level of abstraction.

Having a single lock for stats and cache is very problematic because you'll end up with 7 waiting cores that just need to read or write a tiny bit from/to stats.txt because 1 core is writing a big object file on a 8 core machine. Even in not so bad cases, waiting will sum up easily.

As @frerich suggested in #70, we should use CacheStatistics's methods to lock something.

What I have in mind is the following:

* Get rid of CacheStatistics.save()
* Create a separate lock for stats.txt: "CacheStatisticsLock" like ObjectCacheLock width other mutex name
* Add a generic private getter and setter `_get(key)` and `_put(key, newValue)` that *lock, read/write file, unlock*
* Make existing getters use the new getter, e.g.
```
    def numCallsWithoutSourceFile(self):
        return self._get("CallsWithoutSourceFile")
```
* Make existing setters use the new setter, e.g.

```
    def registerCallWithMultipleSourceFiles(self):
        old = self._get("CallsWithMultipleSourceFiles")
        self._put("CallsWithMultipleSourceFiles", old+1)
```

* Now we don't want 2 locks for increase operation, so add `_increase(key, ammount)` to `_get` and `_put` that does *lock, read, increase, write, unlock* and refactor the method above to

```
    def registerCallWithMultipleSourceFiles(self):
        self._increase("CallsWithMultipleSourceFiles", 1)
```

* Since every data comes from the stats.txt file now, we can remove the in memory copy of values `self._stats`

This reduces the locking time to a minimal, keeps the code simple because only 3 methods have locking logic in it, solves #70 because we don't need to keep a copy of the current stats value in memory, can be improved by using SQLite later on.

frerich:
Thanks for the patch! Except for minor cosmetic issues (you introduced spaces within `()` in a couple of places, e.g. `CacheStatistics(cache)` became `CacheStatistics( cache )` in a couple of places), I couldn't spot any obvious problems, i.e. I suspect it does the Right Thing(tm).

At the same time, that's my main concern with the patch. I find it hard to tell whether it's correct, which is not you to blame but rather a problem with the current code base. Hence, I'd feel a bit better if we could change the code such that instead of having no obvious bugs (with respect to concurrency), it obviously has no bugs.

My idea would be to devise a series of small refactorings, each of which being reasonably easy to verify, so that we end up in a state which has all operations to be guarded against concurrency happening in a single place. My hope is that the patch which then resolves the concurrency issue(s) is then a lot more obvious in how and why it helps. I.e. it not only does the right thing, it also does the Thing Right. :-)

timblechmann:
#75 and #77 make this obsolete

PR #75, opened 2016-04-16
Fix stats locking

### LINKED ISSUE #70: cache directory and stats go out of sync

webmaster128:
as described in #74.

Closes #70.


clcache.py
```diff
@@ -373,11 +373,14 @@ def save(self):
...
-            self._stats = PersistentJSONDict(os.path.join(objectCache.cacheDirectory(),
-                                                          "stats.txt"))
+    def __init__(self, cacheDirectory):
+        lockName = cacheDirectory.replace(':', '-').replace('\\', '-') + "_stats"
+        timeoutMs = 10 * 1000
+        self._lock = SystemSharedLock(lockName, timeoutMs)
+
+        with self._lock:
+            self._stats = PersistentJSONDict(os.path.join(cacheDirectory, "stats.txt"))
+
```
frerich: Using two locks seems very dubious to me. Does this mean that the actual cache contents might now actually be different than what's printed in the statistics?

clcache.py
```diff
@@ -376,101 +376,111 @@ def save(self):
...
+            storage[key] = value
+            storage.save()
+
+    def _increase(self, key, ammount):
+        with self._lock:
+            storage = PersistentJSONDict(os.path.join(self._cacheDirectory, "stats.txt"))
+            if not key in storage:
+                storage[key] = self._defaultValue
+            storage[key] += ammount
+            storage.save()
```
frerich: `_set` and `_increase` are notably similiar. I think you could fix the code duplication here by introducing a generic `_modify` method which is parametrised on the code using which the key in the dictionary is modified:

    def _modify(self, key, fn):
        with self._lock:
            storage = PersistentJSONDict(os.path.join(self._cacheDirectory, "stats.txt"))
            storage[key] = fn(storage.key(key) or self._defaultValue)
            storage.save()

(Writing this code makes me think that it should be possible to use `PersistentJSONDict` with `with`...).

With this methods at hand, `_set` and `_increase` are just:

    def _set(self, key, value):
        self._modify(key, lambda _: value)

    def _increase(self, key, amount):
        self._modify(key, lambda x: x + amount)

frerich:
Sorry for the delayed response, I was on vacation for the last couple of days to escape the cold April in Hamburg. :-)

I mostly like this change. There are just two things I'm unsure about:

1. Having two separate, independent, locks for accessing the object cache (one used for the statistics, one for the actual data storage) seems dubious. It sounds like the statistics and the actual data might get out of sync?

2. I'm somewhat wary of doing locking at such a low level. In my experience, this tends to be not only inefficient but also incorrect because when you want to do multiple actions as one atomic step then you can't just call those two functions in sequence because there's a small time window inbetween during which the cache is not locked.

One crazy idea came to my mind. Maybe it's possible to collect all the modifications to be done to the cache (i.e. files to be added, cleanups to be performed, stats to be updated) as a list of callables and then perform them all in one go in a single place. I hope that this could reduce the number of places where locking needs to be considered, so it not only simplifies locking but also avoids unnecessary locking.

The inspiration for this idea is software transactional memory, in particular the Haskell implementation (in which you can compose STM actions and execute them all in one go, atomically). Another similiar idea is implemented in SQL transactions. Maybe we can do something like this?

frerich:
Sorry, I forgot the most important bit: Thanks! That's quite some work you did there. Much appreciated! 👍 

One further thing came to my mind: I think it's notoriously hard to test that code is free of concurrency issues - that's why I tend to be somewhat pedantic in this respect. I'd like the code to be such that it's obviously correct.  :-)

clcache.py
```diff
@@ -142,7 +144,7 @@ def __init__(self):
             os.makedirs(self.objectsDir)
         lockName = self.cacheDirectory().replace(':', '-').replace('\\', '-')
         timeout_ms = int(os.environ.get('CLCACHE_OBJECT_CACHE_TIMEOUT_MS', 10 * 1000))
-        self.lock = ObjectCacheLock(lockName, timeout_ms)
+        self.lock = SystemSharedLock(lockName, timeout_ms)
```
frerich: Makes a lot of sense. By now, I think it would be senible to consider splitting the `clcache.py` file, there are a couple of generic utility classes by now. Both `SystemSharedLock{Exception}` as well as `PersistentJSONDict` seem to be good candidates.

webmaster128:
Thanks for review. Let me go through your points giving them a number such that we can keep track, starting with the more trivial ones:

1. **file splitting:** this is an option we can discuss independent of this particular PR. Just open a issue for it to start a discussion. Let's keep this PR branch a manageable size.
2. **self._modify():** Implementing `_mofidy()` using lambda functions looks like a huge overkill just to deduplicate 3-5 lines of code. In my opinion it more important to have code that you can read and understand at the first glance then do extreme [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself). E.g. I do not know what the `_` parameter means without reading additional documentation. But sure, we can easily deduplicate that later. Feel free to open a Github issue and assign it to me if you want it to be done.
3. ***with* operator for PersistentJSONDict:** Since the semantic is not clear here, I would not do that. PersistentJSONDict has file reading from disk, in-memory updates and storing to disk in this class. In some cases you just want to read and sometimes you want to write and update. In contrast to `cd` or `open` I do not see an obvious default behavior that we want to have in almost all use cases.
4. **Having two separate, independent, locks:** The question is, what those locks are made for in the first place. Any kind of locking is evil because it prevents other processes from working while one process is holding a lock. So we should lock only when absolutely necessary. Locking objects makes sense because we do not want process B to read object XYZ before process A is done writing object XYZ. Locking stats.txt makes sense because we do not want file system exceptions on process B because process A has opened the exact same file. File object actions tend to be long lasting (e.g. copy 2 MB .obj file) while stats.txt tent to be very short lasting (read/write some bytes). So locking both with the same lock is VERY expensive. We don't need to bother about any runtime overhead in the entire clcache code when we accept that 7 clcache processes that just want to touch `stats.txt` have to wait because one clcache process is copying an object file (on a 8 core build machine). Having separate locks improves runtime significantly but comes at a minimal cost: during one clcache call, objects and stats diverge. An object can be written to cache but not registered in `stats.txt` when another process reads that stats. However the diverge is minimal in terms of time frame and limited to the amount (size and hits count) that is currently processed. At the point when all processes stopped building, cache and stats get back in sync again. Given the current design of the cleanup process, this is not an issue. As long as a user does not use --stats as high precision realtime data source, I don't see any practical impact.
5. **locking at such a low level:** Locking at this level was my only chance to really understand locking. Before the locks where acquired at different levels of abstraction making it impossible to check by review where not enough or too much was locked. During my work I found a lot of places where the cache was locked multiple times in the same process at different places in the call stack. What I do is *reduce the number of places where locking needs to be considered* by making all `lock` objects private to ensure that the code that need to care about locking fits on one sceen page and no callee has to bother about locking at all.
6. **collect all the modifications:** This is what SQLite does for us when we using in [write ahead log](https://en.wikipedia.org/wiki/Write-ahead_logging) mode, at least for stats. Given the design of this PR, this can be used as a backend for stats.

frerich:
1. **file splitting:** Sure, I was mostly thinking out loud here. It just became more evident to me that separate files would be appropriate, now that the new class name is independent of clcache.

2. **self._modify**: It's not so much the amount of code which made me suggest this but rather the desire to the number of places where locking is done. My hope is that (in the course of some incremental revisions of the code) we can get this number down a lot more. Doing locking in something called `_modify` also resonated with me because it matched my thinking ("*locking needs to be done when `get`ing something from the cache or when `modify`ing it*"). I'm not actually dogmatic about this, but I do have a clear preference here. As for the `_` -- I'm sorry if that obfuscated things. It's a common way to denote unused values in Python and many editors (e.g. the Eclipse IDE) or other tools (e.g. pylint) won't issue "unused variable" warnings if the name is just `_`.

3. **`with` support for PersistentJSDONDict:** Maybe I wasn't clear here - I didn't intend to change the semantics of the class at all. This was -- much like item 1. -- more a thinking out loud thing: if `PersistentJSONDict` defined `__enter__` and `__exit__` then clients *could* use it with `with` if they wanted (and thus make sure that their changes actually end up on the disk, i.e. they are persistent, even in the face of exceptions). They don't need to.

4. **Having two separate, independent, locks:** The purpose of the locks was to ensure that the externally visible state is always consistent, i.e. the statistics printed by clcache actually match what's on the disk. I very much agree with your *So we should lock only when absolutely necessary.* sentiment, and I'm a strong proponent that correctness makes this necessary. In my local tests, the performance difference actually wasn't measurable, but a) I'm using a SSD and b) I have a lot less contention (since I have only two CPU cores). Can you maybe provide some benchmarks to support your position? That would make me feel much better when merging the change. :-) As a side-note: maybe this optimization (that's what it is, right?) is a bit out of scope for this PR. Maybe we shouldn't let this PR get distracted by prolonged discussions of the merits and downsides of the change but rather do so in a separate PR?

I don't actually have any useful to contribute to 5. and 6.. :-)

akleber:
I did a quick benchmark with and without this PR and I could not see any significant time differences. I used a cmake based ninja build with 8 cores and about 1730 cache entries. The wall time for building the project with an empty cache and with a full cache was basically the same with and without this PR. By the way the stats without this fix display only 200 entries compared to the 1730 actual entries. The size is way off too of course.

webmaster128:
Did you use copy or hardlink to store/restore objects? This should have a significant impact on locking durations.

akleber:
Objects were copied.

frerich:
I think if this patch has no measurable impact on the runtime but a very measurable impact on the correctness of the cache statistics, then that's a strong enough argument for inclusion of this patch already.  ;-)

akleber:
I have two more observations: I tried to use a samba share for the cache. When using the warm cache over LAN the build time decreases by about 25% compare to a warm cache on the same local SSD. But I assume that this is a measurement error. Also I saw a python exception mentioning a SystemLock? But I was not able to reproduce it to save the exact stack trace. I will try again tomorrow.

webmaster128:
This email reply message somehow got lost (after *Objects were copied.*)

---------------

On what kind of storage are you using?

Cloud-VMs using a NAS can be incredible slow compared to local hard drives or even local SSDs. 

I did not yet find the time to run a test myself but I'll try to make it happen.

An alternative implementation I have in my mind is locking per directory (first two character of hash) in order to be able to lock objects and stats together. This means having one stats file per folder. This reduces the likelihood of one lock waiting for another to 1/256 [0-9a-f]{2}. If I understood correctly, this is what ccache does.

akleber:
I did some more testings and I could not reproduce the exception I saw once yesterday. I think I can attribute it to some network problem. Also today with a more clear test setup I could not see any significant time differences between local and network cache. I did all test on my local machine on a SSD. When having the cache on the network share the network is a 1Gbit LAN and the cache is located on a NSA with some fast HDD.
To summarize: I do not see and time penalty of this PR compared to clcache without it or any other problem. For my specific project the build time decreased to 1/3 when using clcache on a "clean" build.
I will now deploy clcache with the PR to three nodes of our CI build cluster with a common cache on a network share and monitor its performance.

webmaster128:
> three nodes of our CI build cluster with a common cache on a network share

Are you aware of the fact that clcache does not support sharing a a cache across multiple machines? Since the locking is not shared across multiple systems, this me lead to corrupt builds. Should we support that or at least document it?

webmaster128:
I just ran a bunch of benchmarks on a 4 core Azure Cloud machine (usually hard disk performance) with `CLCACHE_HARDLINK` disabled in order to find whether or not locking times matter.

I compared #74 and this PR which both aim to fix the consistency issue in #70. In both cases I tested when the project is not in cache and when it is in cache:

* A: Single lock (empty cache)
* B: Single lock (project in cache)
* C: Separate locks (empty cache)
* D: Separate locks (project in cache)

Each configuration ran four times in a row.

### A: Single lock (empty cache)

```
Run A1: 684 seconds
Run A2: 705 seconds
Run A3: 709 seconds
Run A4: 713 seconds
```

### B: Single lock (project in cache)

```
Run B1: 166 seconds
Run B2: 164 seconds
Run B3: 167 seconds
Run B4: 165 seconds
```

### C: Separate locks (empty cache)

```
Run C1: 708 seconds
Run C2: 711 seconds
Run C3: 701 seconds
Run C4: 696 seconds
```

### D: Separate locks (project in cache)

```
Run D1: 165 seconds
Run D2: 167 seconds
Run D3: 166 seconds
Run D4: 165 seconds
```

Turns out both approaches perform equally (A vs. C, B vs. D). What surprised me is the consistency of the stats: https://gist.github.com/webmaster128/db8268748ffdf3c5c1012d1f26eb0a2a Where A and B report different stats per run, C and D show 100 % the same output in all 4 runs.

This leads to the theory that the only way to get locking done in a solid maintainable way is to have it in a single low level place of the code.

The question whether one uses one or two different locks seems to be irrelevant in terms of performance. But when we use the low level approach in this PR, we can only guarantee consistency after all clcache processes are done, because a lock is released and acquired again in one process leaving a short timeframe of inconsistency. Given that, we can have two locks without further drawback.

frerich:
Those are some interesting timings! To present yet another approach, I now submitted #77 to show how I envisioned the locking strategy to be. The guiding principle is that low-level locks are avoided (since they do not compose) but instead locking is pulled to the topmost level such that all updates of the cache + stats are transactional. I believe the PR implements a minimal number of lock operations and hopefully improves concurrency while still fixing #70 .

Would it be possible to perform a benchmark of that patch, too?

akleber:
@webmaster128 : I was not aware of the sharing problem, thanks for the heads up. I will try it nevertheless with its problems in mind, as I am also interested if clcache can make a difference for our project/environment.
If I understand the code correctly, currently clcache uses a mutex from the win api and thus the locking does not work between different machines. Do you think it might be possible to use a lock-file instead? I think i could do a quick check if the lock-file based interprocess lock of the "fastener" package might work as a replacement for ObjectCacheLock.

webmaster128:
@akleber 

> If I understand the code correctly, currently clcache uses a mutex from the win api and thus the locking does not work between different machines.

Yes.

> Do you think it might be possible to use a lock-file instead? I think i could do a quick check if the lock-file based interprocess lock of the "fastener" package might work as a replacement for ObjectCacheLock.

Yes I think that can be done. If I remember correctly, ccache (no "L" here) switched from POSIX file based locks to some symlink mechanism because it was more reliable. You might want to search the ccache changelog for any interesting insights. I do not have any additional knowledge about how it can be done.

frerich:
@webmaster128 is correct - clcache currently uses a system-wide mutex to guard against concurrent invocations. The good thing about this is that it's extremely efficient, and very easy to get right.

For what it's worth, clcache used to use file-based locking prior to commit c88c6417 -- but it was dropped because there were issues with the lock file: it did not get deleted in all cases in case clcache was terminated abnormally (e.g. terminated via the task manager). Mutexes don't have this issue, they are released automatically once the process terminates.

frerich:
Superseded by PR #77 (which was merged by now).

PR #77, opened 2016-05-13
Fixed locking strategy to avoid corrupting cache stats

frerich:
Concurrent invocations of clcache.py would easily break the cache
statistics. The reason for this was that in some cases, a
read/modify/write session on the statistics was not guarded by a single
lock, so two clcache.py processes A and B which try to bump some value
(e.g. number of cache entries)might have executed a sequence like

  A read value 5
  B read value 5
  A write value 6
  B write value 6

Fix this by rewriting the locking strategy. Locking is no longer done at
the lowest level (since such locks do not compose) but at the toplevel
in all relevant areas:

- When printing the cache statistics
- When resetting the cache statistics
- When changing the cache configuration

Furthermore, when processing a compiler request, the lock is acquired at
least once in order to:

- Check if the invocation can be cached at all
- Check if there is a cache entry

Then (without the lock being held!) the compiler is invoked. Finally,
depending on the invocation and the configuration settings, a
post-processing function is invoked to update the bookkeeping. This
function also takes care of holding the lock as needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ clcache changelog
 
  * Bugfix: Support use of py2exe with Python 3.4
  * Bugfix: Support includes parsing when cl language is not English (GH #65)
+ * Bugfix: Fix bug causing statistics to get corrupted after concurrent invocations
 
 ## clcache 3.0.2 (2016-01-29)
 
--- a/clcache.py
+++ b/clcache.py
@@ -148,47 +148,45 @@ class ObjectCache:
         return self.dir
 
     def clean(self, stats, maximumSize):
-        with self.lock:
-            currentSize = stats.currentCacheSize()
-            if currentSize < maximumSize:
-                return
+        currentSize = stats.currentCacheSize()
+        if currentSize < maximumSize:
+            return
 
-            # Free at least 10% to avoid cleaning up too often which
-            # is a big performance hit with large caches.
-            effectiveMaximumSize = maximumSize * 0.9
+        # Free at least 10% to avoid cleaning up too often which
+        # is a big performance hit with large caches.
+        effectiveMaximumSize = maximumSize * 0.9
 
-            objects = [os.path.join(root, "object")
-                       for root, folder, files in os.walk(self.objectsDir)
-                       if "object" in files]
+        objects = [os.path.join(root, "object")
+                   for root, folder, files in os.walk(self.objectsDir)
+                   if "object" in files]
 
-            objectInfos = [(os.stat(fn), fn) for fn in objects]
-            objectInfos.sort(key=lambda t: t[0].st_atime)
+        objectInfos = [(os.stat(fn), fn) for fn in objects]
+        objectInfos.sort(key=lambda t: t[0].st_atime)
 
-            for stat, fn in objectInfos:
-                rmtree(os.path.split(fn)[0])
-                currentSize -= stat.st_size
-                if currentSize < effectiveMaximumSize:
-                    break
+        for stat, fn in objectInfos:
+            rmtree(os.path.split(fn)[0])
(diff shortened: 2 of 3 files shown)


frerich:
This is another alternative to #74 and #75 to fix #70 .

webmaster128:
I like the idea. Is there a way to assert that certain methods are only called when a lock was acquired above them in the call stack (e.g. constructor of `CacheStatistics`)?

When the CI of this turns green, I'll try to benchmark it as well.

frerich:
I hope the CI will turn green shortly, just fixing two last issues.

One way to enforce that certain methods are only called when a lock was acquired was to expose those methods on an object returned by the function which acquires a lock, i.e. acquiring a lock means getting a 'cache' object, along the lines of

    with lockCache() as cache:
        ...

webmaster128:
Nice idea. Having the cache object only available when locked is a huge step for correctness. The stats could than be a property of the cache:

```
with lockCache() as cache:
    # copy object into cache
    with cache.stats() as stats:
        # Read stats file
        # modify stats
        # save stats
 ```

**Comment by webmaster128** [comment]
ps.: but I'd leave this as a follow-up refactoring ;)

**Comment by webmaster128** [comment]
TLDR: all timings good, same output as my PR; I prefer this one

-----------------------------

Got the timings (same setup as posted before, added E and F). The results are very stable given that the machine runs in a shared cloud environment, so I think everything in the +/- 10 % could be ignored.

### A: Single lock (empty cache)

```
Run A1: 684 seconds
Run A2: 705 seconds
Run A3: 709 seconds
Run A4: 713 seconds
```

### B: Single lock (project in cache)

```
Run B1: 166 seconds
Run B2: 164 seconds
Run B3: 167 seconds
Run B4: 165 seconds
```

### C: Separate locks (empty cache)

```
Run C1: 708 seconds
Run C2: 711 seconds
Run C3: 701 seconds
Run C4: 696 seconds
```

### D: Separate locks (project in cache)

```
Run D1: 165 seconds
Run D2: 167 seconds
Run D3: 166 seconds
Run D4: 165 seconds
```

#### E: Global locks (empty cache)

```
Run E1: 695 seconds
Run E2: 696 seconds
Run E3: 699 seconds
Run E4: 696 seconds
```

#### F: Global locks (project in cache)

```
Run F1: 176 seconds
Run F2: 172 seconds
Run F3: 172 seconds
Run F4: 189 seconds
```

Restoring from cache might be slightly slower but I tried to choose a worst case environment (4 cores, slow NAS storage and no hardlinks). So the real world impact is negligible.

All stats are 100 % the same with what resulted from my PR: https://gist.github.com/webmaster128/db8268748ffdf3c5c1012d1f26eb0a2a

So 100 % support for merging this. Good job and thanks all for the constructive discussion @frerich @akleber @timblechmann :thumbsup:

timblechmann:
i guess contention would be an issue on a many-core buildserver with hot caches ... not necessarily on a slow machine ... but then those machines will usually have fast drives as well ...

thanks a lot for looking into it!

frerich:
Thanks y'all for providing your feedback, and thanks @webmaster128 for timing things again!

I'm curious - how did you conduct those benchmarks, is there some easily available cloud computing platform available which we may want to perform further tests in the future?

webmaster128:
> I'm curious - how did you conduct those benchmarks, is there some easily available cloud computing platform available which we may want to perform further tests in the future?

I compiled libkullo from Kullo on our main Windows build server, which we got from Microsoft Azur via the Bizspark program. There I have everything setup and could try different clcache branches. The 19 "called for external debug" should be a bundled google-test that somehow uses /Zi.

webmaster128:
Great. Right now I am working on a test that covers this fix (i.e. fails with the old code). Looks good so far.

PR #79, opened 2016-05-15
Add test that covers #77

webmaster128:
This tests fails before a34dc9976 becaus stats have not been counted
properly when multiple clcache processes ran at the same time.

buildall.py builds 10 .cpp files at once to penetrate locking behavior.


frerich:
Good idea! Alas, I now remember that I wanted to create a test as part of #77 but then failed to actually do so. I always just did manual tests. :-(

Do you really need the `buildall.py` script though? Wouldn't it be easier to just launch a couple of processes in parallel, something along the lines of:

    processes = []
    for sourceFile in glob.glob('parallel/*.cpp'):
      processes.append(subprocess.Popen([PYTHON_BINARY, CLCACHE_SCRIPT, sourceFile]))

    for p in processes():
      p.wait()

In fact, you could even reuse the support for compiling multiple source files in clcache itself, e.g.:

    subprocess.check_call([PYTHON_BINARY, CLCACHE_SCRIPT] + glob.glob('parallel/*.cpp'))

...should make clcache invoke itself for each of the source files in parallel.

webmaster128:
> Do you really need the buildall.py script though? Wouldn't it be easier to just launch a couple of processes in parallel [...]

The idea is to separate things that do not necessarily belong together. My test code should read something like "build the C++ project in folder tests/parallel using compiler X and flags Y". buildall.py is basically a replacement for "make --jobs=999", which is unfortunately is not possible using nmake and I wanted to not have the dependency on [jom](https://wiki.qt.io/Jom). Having the build script for the test project separately has the advantage that one can build the project independent of the test suite and can play around with compiler and compiler flags.

If necessary, buildall.py can easily be generalized in oder to be reused, e.g. using glob instead of sources list or having a --jobs argument. But for now, I did not see the benefit.

frerich:
I think keeping the code of the test clean of such implementation details to avoid that the focus of the test is diluted makes a lot of sense.

`buildall.py` however seems to be more complex than it needs to be: instead of forking a couple of threads, each of which starting a new process synchronously, it could just start a couple of processes asynchronously, i.e.:

```python
processes = []
for fn in sourceFiles:
  processes.append(subprocess.Popen([os.environ['CXX'], os.environ['CXXFLAGS'].split(), '/c', fn]))

for p in processes():
  p.wait()
```

By this time, you might find that a separate script doesn't really pull its own weight, so I think you could also move this code into a

```python
def buildall(sourceFiles):
  ...
```

function right in `tests.py` and then call that from `TestRunParallel.testParallel`. That way, you can also define the list of source files (i.e. `fibonacci*.cpp`) right in `TestRunParallel.testParallel`, i.e. the test is no longer spread over two files. I think it would be less code, keep the (very good!) readability of the test itself and prepare reuse by future tests.

For command-line runs (for experimentation), I personally just run a specific test case individually, e.g.

```shell
$ python tests.py TestRunParallel.testParallel
```

webmaster128:
Implemented comments and rebased onto current master. I was not aware of the fact that subprocess.Popen() starts a process asynchronously, thanks.

frerich:
Merged, Thanks!

PR #80, opened 2016-05-16
Pull out processDirect()

webmaster128:
to make processCompileRequest() more readable.


frerich:
Thanks! I'm somewhat unsatisfied with the overall structure of the code and I think simplifying processCompileRequest() is a step in the right direction.

PR #81, opened 2016-05-19
speed test harness for cached builds of openssl

inorton:
Hiya, I've created a py.test test that does timed builds of openssl, no clcache, cold clcache, warm clcache which is useful for benchmarking cache performance.

You need the following for this to be of use:

1. visual studio 11, 12 or 13 installed
2. perl on PATH (windows perl will do)
3. internet access so it can get the openssl source
4. a python installation with py.test included

It can either be run directly or under an IDE like pycharm


speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,253 @@
...
+    :return:
+    """
+    for _ in range(30):
+        # antivirus might be busy in here..
+        try:
+            shutil.rmtree(path)
+            return
+        except WindowsError:
+            time.sleep(2)
+        if os.path.exists(path):
```
webmaster128: Shouldn't this come after the loop?

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+        self.started = time.time()
+        print "\n..begin {} .. ".format(self.message)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        result = "OK"
+        if exc_val is not None:
+            result = "ERROR"
+        print "\n..end {} {}.. ({}sec)".format(self.message, result,
+                                               time.time() - self.started)
+
```
frerich: Using `contextlib.contextmanager`, this class could be simplified to the generator

```python
@contextmanager
def blockMessage(message):
    started = time.time()
    print "\n..begin {} .. ".format(message)
    try:
        yield
        result = "OK"
    except:
        result = "ERROR"
    print "\n..end {} {}.. ({}sec)".format(message, result, time.time() - started)
```

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+
+def download_openssl():
+    """
+    Get the openssl zip and unpack it
+    """
+    if not os.path.exists(OPENSSL_ZIP):
+        getoslsrc = urllib.URLopener()
+        with BlockMessage("download openssl"):
+            getoslsrc.retrieve(OPENSSL_URL, OPENSSL_ZIP + ".part")
```
frerich: Why not just use `urllib.urlretrieve` here?

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+    with zipfile.ZipFile(OPENSSL_ZIP, "r") as unzip:
+        folder = unzip.namelist()[0]
+        if os.path.exists(folder):
+            with BlockMessage("delete old openssl folder"):
+                retry_delete(folder)
+
+        with BlockMessage("unzip openssl"):
+            unzip.extractall()
+
+        if len(SOURCES) == 0:
```
frerich: Regarding testing sequences for emptyness, [PEP 8|https://www.python.org/dev/peps/pep-0008/] says:

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
>
> Yes: if not seq:
>      if seq:
>
> No: if len(seq):
>     if not len(seq):

Based on that, I think `if not SOURCES:` would be better here.

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+            SOURCES.append(folder.rstrip("/"))
+
+
+def find_visual_studio():
+    """
+    Attempt to find vs 11 or vs 12
```
frerich: There seems to be `13` missing here.

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+import time
+
+
+OPENSSL_ZIP = "OpenSSL_1_0_2-stable.zip"
+OPENSSL_URL = "https://codeload.github.com/openssl/openssl/zip/OpenSSL_1_0_2-stable"
+
+THISDIR = os.path.dirname(os.path.abspath(__file__))
+DISTDIR = os.path.join(os.path.dirname(THISDIR), "dist")
+CLCACHE = os.path.join(DISTDIR, "cl.exe")
+SOURCES = []
```
frerich: Why is `SOURCES` a list? It appears the code always only accesses the first element of it?

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+    if not os.path.isfile(CLCACHE):
+        pytest.fail("please build the exe first")
+
+    _, vcvars = find_visual_studio()
+    with BlockMessage("getting vc envs"):
+        getenvs = subprocess.check_output([os.path.join(THISDIR, "get_vc_envs.bat"), vcvars])
+        for line in getenvs.splitlines():
+            if "=" in line:
+                name, val = line.split("=", 1)
+                ENVS[name.upper()] = val
```
frerich: I think you could drop the `get_vc_envs.bat` file in favor of just doing:

```python
output = subprocess.check_output([vcvars, '>', 'NUL', '&&', 'set'], shell=True)
ENVS = dict(line.split('=', 1) for line in output.splitlines())
```

I.e. just run the batch file and then run `set` to print the environment, then parse that. This clearly assumes that every line in the `set` output has at least one `=` sign -- I hope that's true.

speedtest/test_performance_openssl.py
```diff
@@ -0,0 +1,247 @@
...
+    if not pdbs:
+        replace_wipe_cflags(os.path.join(SOURCES[0], nmakefile))
+
+    if addpath is not None:
+        envs["PATH"] = addpath + os.pathsep + envs["PATH"]
+
+    try:
+        with BlockMessage("running nmake"):
+            subprocess.check_output(["nmake", "-f", nmakefile],
+                                    shell=True,
```
frerich: Why do you need `shell=True` here?

frerich:
Thanks for this test! I think good and stable tests for measuring the performance of a caching systems are very much needed. The Perl dependency is a little bit annoying though. Do you have some idea on how to run the test automatically, e.g. as part of AppVeyor (which we already use)?

speedtest/test_performance_openssl.py
inorton: Yes! funnily enough I've been looking at a bug because of this :)

speedtest/test_performance_openssl.py
inorton: Ooh! I rather like that!

speedtest/test_performance_openssl.py
inorton: Ok, that makes sense

speedtest/test_performance_openssl.py
inorton: that might work yes!

speedtest/test_performance_openssl.py
inorton: Actually it doesn't here, but I found that check_call() doesn't process any supplied env vars unless shell=True, this used to be a check_call()

inorton:
If AppVeyor can't run the perl configure step I'm not sure how we could cope with that.

webmaster128:
Good job!

How long does this run? I don't think it is a great idea to blow up tests from 3 minutes to 30 just to get an idea of how things perform.

Regarding Perl: Could we switch from OpenSSL to Botan as a test project? Botan has a Python (2.6–3.5) configure script and can be build using nmake or jom from there.

frerich:
I have no strong opinion on any particular project to test `clcache` on, but I very much agree that having a real-life project is very desirable, and having one which doesn't introduce too many extra dependencies in addition to those we already have is preferable. With that in mind, "Botan" sounds interesting (though I didn't look at it).

inorton:
On my (not very good) PC it takes 4 minutes to build openssl without
clcache.

This could of course use a smaller set of test sources but I wanted a
fairly representative test.

You don't have to run the entire suite, individual bits can be selected on
the pytest cmdline.

inorton:
Don't worry, I've not forgotten, just settling down to push some fixes to this

inorton:
Re openssl, it wouldn't be too hard to add a "light" test and use Botan or some other thing for that yet keep openssl in (we build openssl alot where I work so it gets a good hammering)

inorton:
I hope you don't mind, I think it would be easier to look at if I closed this pull request and made another (i've moved some funcs about)

frerich:
Sure, go wild. :-)

inorton:
turns out botan needs GNU Make, i'll look for another small project

inorton:
moved to another PR

webmaster128:
Nope. We're buildung Botan with nmake or jom on Windows.  
> turns out botan needs GNU Make
>

PR #82, opened 2016-05-23
adds lockfile based locking strategy

akleber:
This PR adds an additional lockfile based locking mechanism. To enable the lockfile based locking, "LockingStrategy" has to be set to "File" in the cache config file. A command line parameter to set the locking mechanism is not yet provided. In contrast to an earlier implementation that was removed some time ago, this implementation does not rely on a specific file being present or not, but instead uses a win32 api to explicitly lock the file. This mechanism is based on a file handle of the actual lock file and should release the lock in case the process dies unexpectedly.
This PR is still rough around the edges, but I wanted to provided a first look.
My reason for implementing this is, that I want to share the cache via a smb share to multiple build machines.
I ran some benchmarks (unfortunately not with the new test harness). It ran on a 12 core machine with SSD and 1Gibt network. All numbers are averages over 4 runs. The cache contains 1740 objects and is 1.5 Gb in size.

Lockfile, local cache, cold cache: 480s
Mutex, local cache, cold cache: 482s

Lockfile, local cache, warm cache: 270s
Mutex, local cache, warm cache: 272s

Lockfile, cache on smb, cold cache: 525s
Lockfile, cache on smb, warm cache: 361s

So it seems that the performance of the lockfile compared to the mutex is not significant different. I attribute the differences to inaccuracies in the measurement.
Placing the cache on a smb share has however a noticeable penalty.
I am going to test this implementation on a build cluster with a "real world" work load and have a look how it performs and will report back.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -51,6 +51,9 @@ from subprocess import Popen, PIPE
 import sys
 import multiprocessing
 import re
+import time
+import msvcrt
+import random
 
 VERSION = "3.0.3-dev"
 
@@ -84,10 +87,9 @@ class LogicException(Exception):
         return repr(self.message)
 
 
-class ObjectCacheLock:
+class ObjectCacheLockMutex:
     """ Implements a lock for the object cache which
     can be used in 'with' statements. """
-    INFINITE = 0xFFFFFFFF
 
     def __init__(self, mutexName, timeoutMs):
         mutexName = 'Local\\' + mutexName
@@ -126,6 +128,80 @@ class ObjectCacheLock:
         self._acquired = False
 
 
+class ObjectCacheLockFile:
+    """ Implements a lock file lock for the object cache which
+    can be used in 'with' statements.
+    Inspired by
+    https://github.com/harlowja/fasteners
+    https://github.com/benediktschmitt/py-filelock"""
+
+    def __init__(self, lockfileName, timeoutMs):
+        self._lockfileName = lockfileName
+        self._lockfile = None
+        self._timeoutMs = timeoutMs
+
+    def __enter__(self):
+        self.acquire()
+
+    def __exit__(self, typ, value, traceback):
+        self.release()
+
+    def __del__(self):
+        self.release()
+
+    def _acquire(self):
+        try:
+            lockfile = open(self._lockfileName, 'a')
+        except OSError:
+            lockfile = None
+        else:
+            try:
+                msvcrt.locking(lockfile.fileno(), msvcrt.LK_NBLCK, 1)
+            except (IOError, OSError):
+                lockfile.close()
+            else:
+                self._lockfile = lockfile
+        return None
+
+    def _release(self):
+        if self._lockfile is not None:
+            lockfile = self._lockfile
+            self._lockfile = None
+            msvcrt.locking(lockfile.fileno(), msvcrt.LK_UNLCK, 1)
+            lockfile.close()
+
+            #The following might fail because another instance already has locked the file.
+            #This is no problem because the existence of the file does not provide the
+            #locking but win32 api base file locking mechanism.
+            try:
+                os.remove(self._lockfileName)
+            except OSError:
+                pass
+
+        return None
+
+    def is_locked(self):
+        return self._lockfile is not None
+
+    def acquire(self):
+        start_time = time.time()
+        while True:
+            if not self.is_locked():
+                self._acquire()
+
+            if self.is_locked():
+                break
+            elif self._timeoutMs >= 0 and time.time() - start_time > self._timeoutMs/1000:
+                raise ObjectCacheLockException("Timeout waiting for file lock")
+            else:
+                poll_intervall = random.uniform(0.01, 0.1)
+                time.sleep(poll_intervall)
+
+    def release(self):
+        if self.is_locked():
+            self._release()
+
+
 class ObjectCache:
     def __init__(self):
         try:
@@ -142,7 +218,13 @@ class ObjectCache:
             os.makedirs(self.objectsDir)
         lockName = self.cacheDirectory().replace(':', '-').replace('\\', '-')
         timeout_ms = int(os.environ.get('CLCACHE_OBJECT_CACHE_TIMEOUT_MS', 10 * 1000))
-        self.lock = ObjectCacheLock(lockName, timeout_ms)
+
+        cfg = Configuration(self)
+        if cfg.lockingStrategy() == "File":
+            lockfileName = os.path.join(self.cacheDirectory(), "cache.lock")
+            self.lock = ObjectCacheLockFile(lockfileName, timeout_ms)
+        else:
+            self.lock = ObjectCacheLockMutex(lockName, timeout_ms)
 
     def cacheDirectory(self):
         return self.dir
@@ -342,7 +424,8 @@ class PersistentJSONDict:
 
 
 class Configuration:
-    _defaultValues = {"MaximumCacheSize": 1073741824} # 1 GiB
+    _defaultValues = {"MaximumCacheSize": 1073741824, # 1 GiB
+                      "LockingStrategy": "Mutex"}
 
     def __init__(self, objectCache):
         self._objectCache = objectCache
@@ -357,6 +440,9 @@ class Configuration:
 
     def setMaximumCacheSize(self, size):
         self._cfg["MaximumCacheSize"] = size
+
+    def lockingStrategy(self):
+        return self._cfg["LockingStrategy"]
 
     def save(self):
         self._cfg.save()
@@ -888,6 +974,7 @@ clcache statistics:
   current cache dir         : {}
   cache size                : {:,} bytes
   maximum cache size        : {:,} bytes
+  locking strategy          : {}
   cache entries             : {}
   cache hits                : {}
   cache misses
@@ -905,6 +992,7 @@ clcache statistics:
         cache.cacheDirectory(),
         stats.currentCacheSize(),
         cfg.maximumCacheSize(),
+        cfg.lockingStrategy(),
         stats.numCacheEntries(),
         stats.numCacheHits(),
         stats.numCacheMisses(),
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+        self.release()
+
+    def __del__(self):
+        self.release()
+
+    def _acquire(self):
+        try:
+            lockfile = open(self._lockfileName, 'a')
+        except OSError:
+            lockfile = None
```
frerich: I think this assignment is unneeded: `_acquire` is only called if `self.lockfile == None` holds, which is still `True` in case `open` raises an exception. I'd say you can just `return` here.

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+        else:
+            try:
+                msvcrt.locking(lockfile.fileno(), msvcrt.LK_NBLCK, 1)
+            except (IOError, OSError):
+                lockfile.close()
+            else:
+                self._lockfile = lockfile
+        return None
```
frerich: For [stylistic reasons|http://stackoverflow.com/questions/15300550/python-return-return-none-and-no-return-at-all], I think it would be fine to just emit the `return` statement completely here: the method never returns any value in any of the code paths.

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+
+    def _release(self):
+        if self._lockfile is not None:
```
frerich: This `if` is both unneded an asymmetric: `_release` is only called if `is_locked()` evaluates to `True`, which is only the case if `self._lockfile is not None` holds. Furthermore, it's asymmetric, because `_acquire` doesn't check `self._lockfile` itself but rather relies on the caller to do so (i.e. `_acquire` maintains it as a precondition). I'd vote for removing this `if` to be able to reduce the indentation of the following code block.

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+
+            #The following might fail because another instance already has locked the file.
+            #This is no problem because the existence of the file does not provide the
+            #locking but win32 api base file locking mechanism.
+            try:
+                os.remove(self._lockfileName)
+            except OSError:
+                pass
+
+        return None
```
frerich: Same remarks regarding `return None` apply here which I mentioned in line 164 above.

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+            msvcrt.locking(lockfile.fileno(), msvcrt.LK_UNLCK, 1)
+            lockfile.close()
+
+            #The following might fail because another instance already has locked the file.
+            #This is no problem because the existence of the file does not provide the
+            #locking but win32 api base file locking mechanism.
+            try:
+                os.remove(self._lockfileName)
+            except OSError:
+                pass
```
frerich: Given that the existence of the file doesn't matter, I guess it could be argued that there's no point in removing the file - how about just dropping this code completely?

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+
+    def acquire(self):
+        start_time = time.time()
+        while True:
+            if not self.is_locked():
+                self._acquire()
+
+            if self.is_locked():
+                break
+            elif self._timeoutMs >= 0 and time.time() - start_time > self._timeoutMs/1000:
```
frerich: I think it would be nicer to enforce that `self_timeoutMs >= 0` holds right in the constructor such that you don't have to repeatedly check it here.

clcache.py
```diff
@@ -1165,7 +1253,11 @@ def processCompileRequest(cache, compiler, args):
 
 def processDirect(cache, outputFile, compiler, cmdLine, sourceFile):
     manifestHash = ObjectCache.getManifestHash(compiler, cmdLine, sourceFile)
-    with cache.lock:
+    postProcessing = None
+
+    try:
```
frerich: Not sure I get this change: why couldn't you go with `with cache.lock` anymore but needed to explicitly `acquire`/`release` the lock? Something to do with scoping?

frerich:
Great PR, it looks very good! The idea to not rely on the sheer existence of a file for the locking but rather locking file regions (and then relying on those regions being released automatically when the file handle is closed) is really nice, that didn't occur to me. It nicely takes care of the main gripe we had with file-based locks (which is: not cleaning up in case of abnormal exists). I really only have a few stylistic concerns here.

I'd like to hear what you think about my comments, but in general I'm ready to merge this.

webmaster128:
I guess this non-trivial change justifies an entry in the copyright paragraph, right? What do you think @frerich?

Does this file based locking require any specific file system? Or can I do that on a FAT or FTP mount? Could you extend the README to tell people everything about File locking they need to know?

Regarding configuration: everything except cache size is configured via an environment variable and my feeling is that it would be better to move cache size there too. Since we're passing to cl, I think we should keep the clcache parameter parsing logic as minimal as possible. So for now I'd suggest configuring the locking method via an environment variable. Having everything in variables allows us to migrate to a more general configuration approach [as seen in ccache](https://ccache.samba.org/manual.html#_configuration) Frerich?

clcache.py
```diff
@@ -126,6 +128,80 @@ def release(self):
...
+        while True:
+            if not self.is_locked():
+                self._acquire()
+
+            if self.is_locked():
+                break
+            elif self._timeoutMs >= 0 and time.time() - start_time > self._timeoutMs/1000:
+                raise ObjectCacheLockException("Timeout waiting for file lock")
+            else:
+                poll_intervall = random.uniform(0.01, 0.1)
```
webmaster128: camelCase; is this really an interval? I think pollDelay would be a better name indicating that the delay before the next poll – a single float number – is stored.

clcache.py
```diff
@@ -142,7 +218,13 @@ def __init__(self):
             os.makedirs(self.objectsDir)
         lockName = self.cacheDirectory().replace(':', '-').replace('\\', '-')
         timeout_ms = int(os.environ.get('CLCACHE_OBJECT_CACHE_TIMEOUT_MS', 10 * 1000))
-        self.lock = ObjectCacheLock(lockName, timeout_ms)
+
+        cfg = Configuration(self)
```
webmaster128: This one is tricky ;) We need to hold a lock before we're allowed to access the configuration. At least this is what the current code suggests: https://github.com/frerich/clcache/blob/dc63c22f2d36fc624ea83f0187ef045e09c2c0e4/clcache.py#L1104

So this might be the point where moving to environment variables makes life easier.

frerich:
@webmaster128 Wether or not some configuration knob is done as part of the cache configuration or via environment variables is - to me - mostly an issue of whether it's something you typically turn on and off for certain projects and/or builds (in which case I'd go for an environment variable), or whether it's a property of the cache itself which is independent of any particular build (e.g. the maximum cache size) in which case I'd go for a cache configuration switch.

To me, @akleber 's decision seems perfectly sensible.

As for the copyright notice -- if @akleber doesn't mind getting some of the bl^H^Hfame, I'd be happy to add him to the list of contributors.

clcache.py
frerich: Good catch! There's a hen and egg problem here: the `Configuration` constructor reads the cache configuration. To do so, you first need to acquire a lock. However, to know what kind of lock to acquire you currently need to look at the cache configuration. Oops.

Quite frankly, given that file-based locking doesn't seem to have any downsides so far (at least not timing-wise), I think we could just as well make it the default for the next version and see how it goes.

webmaster128:
+1 for making this the default behavior (override via environment variable?) as soon as we have the file system requirements in the table.

clcache.py
akleber: I agree.

clcache.py
akleber: Sorry some temporary change. I revert it. I wanted to try if it is possible to fall back to just invoke the compiler in case we hit a timeout waiting for the lockfile instead of aborting the build.
Before adding the random pollDelay I observed some race conditions, where 12 clcache process were fighting for the lockfile and no one made progress.

akleber:
Thanks for the thorough review so far.

Regarding the config location: My thinking was that the type of locking is specific to the cache, so every process that wants to use a particular cache has to use the appropriate locking mechanism.
However I perfectly see the hen and egg problem.
I think changing the locking mechanism configuration to an env variable is fine in so far, as one who sets up a shared cache should also be able to properly configure the locking setting on the participating clcache processes.

@frerich I am fine with being added to the list of contributors.

Also I am adding relevant infos to the README regarding the lockfile requirements and usage.

webmaster128:
I was especially thinking of the copyright header in the clcache.py file. I think this should be extended with every non-trivial change to that file.

akleber:
I updated this PR with changes concerning the reviews. However I also observed some race-condition when using a network share for the cache so please do NOT merge this PR yet. I am still investigating...

clcache.py
```diff
@@ -186,7 +186,7 @@ def acquire(self):
             elif time.time() - startTime > self._timeoutMs/1000:
                 raise ObjectCacheLockException("Timeout waiting for file lock")
             else:
-                pollDelay = random.uniform(0.01, 0.1)
+                pollDelay = random.uniform(0.1, 1.0)
```
frerich: I'm curious: what does this fix? SMB connections don't like polling too often for some reason?

frerich:
@akleber ping? Any new insights recarding that race condition when using a network share? I saw that you bumped the retry interval when polling SMB shares, did that help with anything? Do you understand why?

akleber:
I currently do not have much time to work on this, but I will soon hopefully. Unfortunately I was not yet able to fix the race condition when using the lockfile on a smb share and I even do not yet understand what the problem might be. Using the lockfile on a local filesystem works fine however as far I could see.

frerich:
Thanks for the update, @akleber ! Pity that the cause for the race condition is still unknown, because non-local file systems were the prime reason for using a file-based locking mechanism in the first place. I guess it doesn't hurt to keep this PR around, I suspect it's "almost there".

PR #83, opened 2016-05-24
Speedtest curl

inorton:
Cache performance test harness using curl and/or openssl for representative tests.

Curl builds are slightly faster than openssl ( 90sec uncached vs 4 minutes)

The curl build test only requires that you have visual studio installed and have built the clcache py2exe executable.


webmaster128:
I am working on a similar test using Botan, because

curl

* Makefile.vc mode: cannot be build in parallel
* Makefile.vc mode: Cannot use custom CC before the upcoming release (see [1](https://github.com/curl/curl/pull/949), [2](https://github.com/curl/curl/pull/950), [3](https://github.com/curl/curl/pull/952))
* ccache: Cannot use custom CC before the upcoming release (see [set CC](https://github.com/curl/curl/blob/82617e7499138077db55a0bc2878e17edd16dd06/CMakeLists.txt#L1135))

openssl

* depends on perl

Botan on the other hand

* Custom CC (so `CC="python path/to/clcache.py"` can be used – no environment changes)
* is configured using Python
* Makefile can be build using nmake or jom (thus multiple jobs)
* no dependencies apart from Python 2.6–3.6, jom and MSVS 2013+
* module system allows reducing the build time (full build on a 2 core local VM: 24 min cold cache, 1 min hot cache)

I just wait for https://github.com/randombit/botan/pull/586 or https://github.com/randombit/botan/pull/587 in order to get proper cache hitting.

inorton:
I did manage the get botan running in this pull request. Good luck though :)

webmaster128:
May I reuse code from this PR for the Botan version? I am especially interested in the Python implementation of Download+unzip to keep dependencies low.

inorton:
sure, feel free

PR #50, opened 2016-01-23
Add license to tests

webmaster128:
This adds the same 3-clause BSD license that we have in clcache.py to tests.py.

I used the [Github generated version](https://github.com/webmaster128/github-bsd-3-clause/blob/master/LICENSE) of the license to have useful witespace and linebreaks.

Additionally I changed

> Neither the name of [repository_name] nor the names of its contributors

to

> Neither the name of the copyright holder nor the names of its contributors

which I took from https://opensource.org/licenses/BSD-3-Clause

Now the file was created by @vchigrin, so I'd like to ask you, if it is in your interest to license your work under these conditions. I you want, I can add your real name, company or email address. Just leave a comment, what you prefer.

PR #85, opened 2016-05-31
Parsing option file with line breaks

IUdalov:
Hi.
This pull request adds support for multiline command file.
I faced with this problem during integration clcache into gradle.

P.S.
Also minor fix for decode compiler output when CLCACHE_NODIRECT enabled.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -206,7 +206,7 @@ class ObjectCache:
         (preprocessedSourceCode, pperr) = preprocessor.communicate()
 
         if preprocessor.returncode != 0:
-            sys.stderr.write(pperr)
+            sys.stderr.write(pperr.decode("utf-8"))
             sys.stderr.write("clcache: preprocessor failed\n")
             sys.exit(preprocessor.returncode)
 
@@ -581,7 +581,7 @@ def extractArgument(argument):
     return argument.strip()
 
 
-def splitCommandsFile(line):
+def splitCommandsLine(line):
     # Note, we must treat lines in quotes as one argument. We do not use shlex
     # since seems it difficult to set up it to correctly parse escaped quotes.
     # A good test line to split is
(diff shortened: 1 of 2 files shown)


clcache.py
```diff
@@ -206,7 +206,7 @@ def computeKey(compilerBinary, commandLine):
         (preprocessedSourceCode, pperr) = preprocessor.communicate()
 
         if preprocessor.returncode != 0:
-            sys.stderr.write(pperr)
+            sys.stderr.write(pperr.decode("utf-8"))
```
webmaster128: How can we know that cl's output is UTF-8? As far as I know it is not and there is proper documentation missing about cl's encoding.

webmaster128:
Thanks! Could you please split this in separate PRs or at least separate commits?

Regarding tests.py: can we wait for a test re-write as discussed in #50 before merging this? I am going to give this a priority and think I can do my job (#50) this week. Then we're going to have a solid base for additions like this one in tests.py.

clcache.py
IUdalov: I'll try to fix it.
Probably method from `expandCommandLine` will work.

IUdalov:
I have opened PR #86 with changes related to multiline options file. 
I am okey with waiting #50.

clcache.py
webmaster128: Could you please open an issue for this decoding issue? Even if you don't have a perfect solution. Maybe we can find out what the problem is together.

IUdalov: Promlem is in sys.stderr  API.
sys.stderr.write(perr) works for python 2.7 compatible. 
sys.stderr.writebuffer(perr) for python 3 (or somthing like it).
I just do bit know how to forward bytes to stderr for both versions without explicity decoding.

webmaster128: I see. What do you think about this implementation? https://github.com/frerich/clcache/pull/117/commits/5fd8a4142418ac4c27db3548fe9e8f3b8e44b82c

webmaster128: The commit I deeplinked was altered. Just check https://github.com/frerich/clcache/pull/117

PR #86, opened 2016-06-01
Parsing multiline options file

IUdalov:
This PR adds support for cl options file with multiple lines.
Usually these files generated by gradle.

Separation of #85.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -581,7 +581,7 @@ def extractArgument(argument):
     return argument.strip()
 
 
-def splitCommandsFile(line):
+def splitCommandsLine(line):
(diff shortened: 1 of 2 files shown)


webmaster128:
Looks good to me. Could you rewrite that for the current master?

The test case TestSplitCommandsFile was removed entirely due to unclear license. But you are free to write a similar test on your own.

Tests are split in unit tests and integration tests now. Since you are calling a clcache method directly, this belongs in unittests.py.

frerich:
PR #102 supersedes this: it's the same code, just rebased on top of current master.

PR #88, opened 2016-06-02
command line: implement cleanCache and clearCache

timblechmann:
implement explicit cleanup/clear as command line arguments (as in `ccache -c` or `ccache -C`)

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -925,6 +925,19 @@ def resetStatistics(cache):
     stats.save()
     print('Statistics reset')
 
+def cleanCache(cache):
+    cfg = getConfiguration()
+    stats = CacheStatistics(cache)
+    cache.clean(stats, cfg.maximumCacheSize())
+    stats.save()
+    print('Cache cleaned')
+
+def clearCache(cache):
+    stats = CacheStatistics(cache)
+    cache.clean(stats, 0)
+    stats.save()
+    print('Cache cleared')
+
 
 # Returns pair - list of includes and new compiler output.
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -925,6 +925,19 @@ def resetStatistics(cache):
     stats.save()
     print('Statistics reset')
 
+def cleanCache(cache):
+    cfg = getConfiguration()
+    stats = CacheStatistics(cache)
+    cache.clean(stats, cfg.maximumCacheSize())
+    stats.save()
+    print('Cleaning cache')
```
frerich: I think these print statements should be shown *before* doing the action (e.g. cleaning the cache).

frerich:
Looks good - can you adjust the code such that the information messages are printed earlier? If cleaning the cache takes longer, the current code seems to just sit there doing nothing and then it prints the "Cleaning the cache" message after the fact. Otherwise, I think this is good to go!

webmaster128:
Wouldn't it be consistent with 'Statistics reset' to have a final print() feedback when done? In this case it must be past "Cleaned/Cleared". As a user I'd expect a process to do work when I call -c or -C without any output.

frerich:
Right, either the tense of the message should be changed to simple past, or the message should come earlier.

I don't care too much into which direction this is changed, it just seemed odd to start a destructive operation like "clear my cache" and then it sits there without me knowing whether it's doing something. :-]

frerich:
Merged this, thanks! I think the only minor things which were raised during code review don't warrant not including this useful feature in the mainline tree.

PR #89, opened 2016-06-02
clean: use scandir if possible

timblechmann:
`scandir.walk` should be faster than `os.walk`. compare https://www.python.org/dev/peps/pep-0471


frerich:
Thanks! Seems straightforward enough - the comment doesn't seem to add any useful information, but that's not a biggie. IMHO it can be removed in a subsequent commit.

PR #90, opened 2016-06-02
robust cache handling

timblechmann:
the current implementation completely relies on the correct entries of the stats file. that might not reflect the real-world e.g. if parts of the cache were deleted manually by a user.

the cache also does not correctly reflect the number of entries as removing objects from the cache won't update the stats

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -163,6 +163,9 @@ class ObjectCache:
         objectInfos = [(os.stat(fn), fn) for fn in objects]
         objectInfos.sort(key=lambda t: t[0].st_atime)
 
+        # compute real current size to fix up the stored cacheSize
+        currentSize = sum(x[0].st_size for x in objectInfos)
+
         for stat, fn in objectInfos:
             rmtree(os.path.split(fn)[0])
             currentSize -= stat.st_size
--- a/clcache.py
+++ b/clcache.py
@@ -166,17 +166,22 @@ class ObjectCache:
         # compute real current size to fix up the stored cacheSize
         currentSize = sum(x[0].st_size for x in objectInfos)
 
+        removedItems = 0
         for stat, fn in objectInfos:
             rmtree(os.path.split(fn)[0])
+            removedItems += 1
             currentSize -= stat.st_size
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -163,6 +163,9 @@ def clean(self, stats, maximumSize):
         objectInfos = [(os.stat(fn), fn) for fn in objects]
         objectInfos.sort(key=lambda t: t[0].st_atime)
 
+        # compute real current size to fix up the stored cacheSize
+        currentSize = reduce( lambda x, y: x + y[ 0 ].st_size, objectInfos, 0 )
```
frerich: There's a common name for the function which reduces a sequence of values using `+`: `sum`. :-)

So how about just

    currentSize = sum(x[0].st_size for x in objectInfos)

frerich:
Looks good - can you please merge the last two commits? The last one seems to be a fixup for the second one. Other than that, I think this can go in.

clcache.py
```diff
@@ -163,17 +163,25 @@ def clean(self, stats, maximumSize):
...
         for stat, fn in objectInfos:
             rmtree(os.path.split(fn)[0])
+            removedItems += 1
             currentSize -= stat.st_size
             if currentSize < effectiveMaximumSize:
                 break
 
         stats.setCacheSize(currentSize)
+        stats.setNumCacheEntries(len(objectInfos) - removedItems)
+
```
webmaster128: I think there should be only one empty line between two methods in a class

clcache.py
```diff
@@ -163,17 +163,25 @@ def clean(self, stats, maximumSize):
...
 
     def removeObjects(self, stats, removedObjects):
         if len(removedObjects) == 0:
             return
+
         currentSize = stats.currentCacheSize()
```
webmaster128: `currentSize` can die then, right?

webmaster128:
Beautiful!

Are you aware that there is no notification when you push new commits to your PR branch? So you might want so add a short comment when you implemented review feedback.

frerich:
Merged, thanks!

PR #91, opened 2016-06-06
Feature/failsafe cleanup

timblechmann:


clcache.py
```diff
@@ -164,33 +164,43 @@ def clean(self, stats, maximumSize):
...
-                   for root, folder, files in walker(self.objectsDir)
-                   if "object" in files]
+        objectInfos = []
+        for x in xrange(16):
+            try:
+                objects = [os.path.join(root, "object")
+                       for root, folder, files in walker(self.objectsDir)
+                       if "object" in files]
+
+                objectInfos = [(os.stat(fn), fn) for fn in objects]
```
timblechmann: we had some issues when `stat` threw. possibly this happened when files were removed by a user (or parallel clcache process).

trying 16 times is an pragamtic heuristic. not sure if we should end up with a more verbose error message.

clcache.py
```diff
@@ -164,33 +164,43 @@ def clean(self, stats, maximumSize):
...
-        objectInfos = [(os.stat(fn), fn) for fn in objects]
         objectInfos.sort(key=lambda t: t[0].st_atime)
 
         # compute real current size to fix up the stored cacheSize
         currentSize = sum(x[0].st_size for x in objectInfos)
 
         removedItems = 0
         for stat, fn in objectInfos:
-            rmtree(os.path.split(fn)[0])
+            rmtree(os.path.split(fn)[0], ignore_errors=True)
```
timblechmann: again: `rmtree` may throw when the directory has been deleted from elsewhere. better gracefully ignore the error than fail (especially on buildservers)

clcache.py
frerich: If it's `stat` which threw, why bother re-trying? And why re-iterate the `object` directory? Wouldn't it be more straightforward to just skip files for which `os.stat` fails, like

```python
objectInfos = []
for o in objects:
    try:
        objectInfos.append((os.stat(fn), fn))
    except:
        pass
```

As for parallel `clcache` processes causing this: I would hope that this cannot be true because there's a lock file to prevent this kind of issue...

timblechmann: if `stat` throws by deleting the cache, the `object` directory most likely has changed. but i don't have any strong opinion on this, so just skipping is fine as for me.

i had been using this patch for a while (when parallel processes were still an issue). these days, it is mainly that users can delete the cache without causing builds to fail.

webmaster128: > it is mainly that users can delete the cache without causing builds to fail.

I don't think we should support that because when we start caring here, we must handle all cases where some random outside process does anything with the cache. Especially since we have `clcache -C` now, there should not be a reason to manually touch the cache while build processes are running.

timblechmann: no matter if there is a reason to manually delete the cache or not ... people will do it :/

this patch won't add a lot of overhead, but will make the code more robust.

webmaster128: Sure they will and I am a heavy abuser of this myself (since there was no -C for a long time). But when I do so, I wait until my compilations are done and otherwise live with build errors. When you're deleting the cache while building, you risk restoring corrupted object files without noticing it, which is something you should avoid.

The patch does not add runtime overhead, right. But it adds code complexity and WTF moments when trying to understand the code. This can be done, but I don't see that it is worth it here.

timblechmann: corrupted cache files ... shouldn't deleting be atomically? i'm a bit afraid that people who are not aware of clcache (like sys admins) will just log onto build servers, see the huge folder and delete it. (i'm not worrying so much about local builds).

webmaster128: I am not exactly sure how hard the mess can be, but

 1. Find that cache entry exists
 2. Restore from cache
 3. Update stats

are not atomic and the deletion might happen anywhere in between.

> people who are not aware of clcache (like sys admins) will just log onto build servers, see the huge folder and delete it.

Isn't that an issue of company structures, responsibilities and competences? Sounds like you have much bigger issues then build fails when you run in this kind of situations. I mean either this guy knows the software that is running or he/she should not be allowed to delete stuff.

timblechmann: > I mean either this guy knows the software that is running or he/she should not be allowed to delete stuff.

difference between theory and reality :)

frerich:
Looks good, merged. Thanks!

webmaster128:
Has this ever been tested? There is no such thing as `fn` …

frerich:
Indeed, a good opportunity to add an integration test for this, which I'm doing in this very moment. ;-)

webmaster128:
ps.: I found this one by playing around with pylint, not by looking at the code ;)

frerich:
I also started enjoying AppVeyor quite a bit, submitting WIP PR's to my own repo just to get the integration test glory with different Python versions. And it sounds like pylint will make a great addition to the toolchain. :-)

PR #92, opened 2016-06-06
Add initial pylint configuration

webmaster128:
This is work in progress

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -141,7 +141,7 @@ inlinevar-rgx=[A-Za-z_][A-Za-z0-9_]*$
 variable-rgx=[a-z]+([A-Z][a-z0-9]+){0,5}$
 
 # Regular expression matching correct class attribute names
-class-attribute-rgx=([A-Za-z_][A-Za-z0-9_]{2,30}|(__.*__))$
+class-attribute-rgx=([A-Za-z_][A-Za-z0-9_]{1,30}|(__.*__))$
 
 # Regular expression which should only match function or class names that do
 # not require a docstring.
--- a/.pylintrc
+++ b/.pylintrc
@@ -269,7 +269,10 @@ dummy-variables-rgx=_$|dummy
 
 # List of additional names supposed to be defined in builtins. Remember that
 # you should avoid to define new builtins when possible.
-additional-builtins=
+#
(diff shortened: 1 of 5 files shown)


webmaster128:
As far as I can see, this is done. In my opinion, Pylint is a great tool that already detected serious bugs. It also checks a lot of codestyle, which I found useful, e.g. that all names are `camelCase` and that local variables must not start with the private prefix `_`.

Other warnings pointed out serious misunderstandings of me about how Python works, most notably: [dangerous default value](http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) fixed in 7def38df096d9237a1cac2d18ad933fa9b0cca20.

I did increase the max line length to 120 (default was 100) because I'd consider a line length 100 as good practice but 120 is still okay in edge cases. Most of the time we needed longer lines, the architecture is a bit messy.

frerich:
Looks nice, but to me the main benefit would be if pylint actually makes the CI build fail in case an error (IIRC those are denoted by a leading `E` in the console output?) is detected, such as an unknown identifier being used. Does it do that yet (maybe via a non-zero exit code of the `pylint` program)?

webmaster128:
It does. pylint returns a non-zero exit code for lint errors and warnings (different code per type). I added 3 pylint commands for the 3 .py files to the AppVeyor config. In fact a lot of those stylistic commits have been made because of failing CI runs for this PR.

webmaster128:
Here you can see a failing CI run caused by pylint because of import order rules:

https://ci.appveyor.com/project/frerich/clcache/build/1.0.251

frerich:
Ah, that looks good - thanks!

frerich:
I suppose with this stuff being merged, all the other PRs should maybe get rebased such that they are automatically checked by pylint, too.

webmaster128:
Well, I would not bee that hard. Rebasing is not trivial and not everyone is familiar with it. It is too easy to destroy hours of work just for a beautiful history.

If PR can be automatically merged, you'll see pylint checks as soon as master is tested. The fixups necessary will be none to minimal.

When one minor change is done to a PR, the PR will be tested again. That is the result of merging the PR into master. So every CI run from now on will include pylint because it is on master.

PR #93, opened 2016-06-06
Fix variable name in LogicException.__init__

webmaster128:
This was probably copy/pasted from different python examples. The member that holds the exception text should be called `self.message`.


frerich:
Thanks! Scary, how long this simple Problem went unnoticed.

PR #94, opened 2016-06-06
Added initial test for clearing the cache

frerich:
As an initial test, verify that cleaning the cache is idempotent, i.e.
repeatedly cleaning the cache should work and maintain the same
postcondition.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -171,7 +171,7 @@ class ObjectCache:
         objectInfos = []
         for o in objects:
             try:
-                objectInfos.append((os.stat(fn), fn))
+                objectInfos.append((os.stat(o), o))
             except WindowsError:
                 pass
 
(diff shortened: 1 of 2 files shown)


webmaster128:
How about using the `clcache -C` call in the integration tests? That would be a more real world scenario and cover [issues like this one](https://github.com/frerich/clcache/commit/ffbe4a65cf9926df05639c910161ddb4bb0cbe59) as well.

frerich:
Good idea!

frerich:
I just noticed that I also got the terminology wrong, what I'm testing here is clearing the cache, not cleaning it. Ahem...

webmaster128:
Very nice :+1: Still the terminology in `TestCleaning` and commit message.

frerich:
True, the commit log needs updating. As for the class name, I actually wanted to use this code both for cleaning and clearing already and thus thought about calling the class `TestCleansing`, but maybe that really just confuses things to much and I should have a separate class.

webmaster128:
I would split it because cleaning and clearing are very different tasks. The fact that clearing is implemented using clean is an implementation detail that might change at any time.

PR #95, opened 2016-06-06
Whitelist branches in AppVeyor config

webmaster128:
to avoid testing feature branches twice (as branch and as pull request)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,8 @@
+# whitelist branches to avoid testing feature branches twice (as branch and as pull request)
+branches:
+  only:
+    - master
+
 environment:
   matrix:
     # AppVeyor installed Python versions


frerich:
It's probably my ignorance about how AppVeyor actually works, but: what's the effect of this commit? Does it mean that only changes to the `master` branch cause the tests to be executed, i.e. submitting PRs doesn't trigger changes anymore, but merging to master does? OTOH, I can see that this PR itself triggered appveyor builds (at least one...?), so something seems to be working. :-]

webmaster128:
Feature branches in this repo that as PRs are tested **twice**. See https://github.com/frerich/clcache/pull/94. This feeds the job queue without adding value. With this config, the feature branches are only tested as PRs.

frerich:
Aha, I see - so by default, AppVeyor is triggered for any commits in any branch *and* for PRs for some branch. With your commit, it's only triggered for commits to master and PRs for master - did I get that right?

frerich:
Does the .yaml file override the configuration done on the appveyor web page (in the project settings)?

frerich:
...aha, indeed: http://www.appveyor.com/docs/build-configuration explains:

> It's worth noticing that both appveyor.yml and UI configuration are mutually exclusive. It's always either YAML or UI - the settings from each are not merged. If you have appveyor.yml in your repo it will override all settings made on the UI unless explicitly disabled by Ignore appveyor.yml.

frerich:
A shorter job queue means faster test results, so I'm all for it!

PR #97, opened 2016-06-06
Use r prefix for regexp string with backslashes

webmaster128:
This is a potential bug because \w and \S are unknown escape sequences
for a normal string


webmaster128:
All the credits go to pylint

frerich:
Merged, thanks to both of you. ;-)

PR #99, opened 2016-06-09
Copy objects to a temporary path an then move

gqmelo:
This lower the chances of ending up with a partial file in case any
error happens during the copy or someone kill the process


webmaster128:
looks good to me. Did you have real world problems because of this?

frerich:
Looks good, merged. Thanks! I'll take care of mentioning this in the ChangeLog myself.

gqmelo:
@webmaster128, I only suspect because on my company we had occasionally some corrupted objs and we have a lot of CI jobs that we can cancel at any time.

Issue #100, opened 2016-06-11
"no source file found" with double backslash in the path


sowsem:
Hi!

I am trying to make clcache work with my cmake-based project. I have been able to actually make bypass an error in clcache just by removing \Zi flags that looks not to be supported.

Unfortunately, looks like there is another problem when using clcache in my project: it complains about being unable to find each single source file for actually caching it for later compilations. I get tons of  "passed to the real compiler" hits in the clcache stats. The error is always the same: no source file found. 

This is a sample of the kind of log I get from clcache:

C:\dev\clcache\dist\clcache.py Cannot cache invocation as ['/nologo', u'/c', u'/I"C:\\dev\\project\\common\\include"' /Fd"Bullet2FileLoader.dir\\Debug\\vc120.pdb" /Gd /TP /wd4201 /wd4512 /analyze- /errorReport:queue "C:\\dev\\project\\src\\Bullet3Serialize\\Bullet2FileLoader\\b3Serializer.cpp"']: no source file found
  
Any idea how to fix this issue or if there is actually an issue in clcache itself?

frerich:
It looks like the quotation marks after /I and /Fd are confusing the command line parser. I'll have a look.

sowsem:
Cool!. As a side note it is being run from cygwin... just in case!

frerich:
Two things - can you please reproduce this error but with having the `CLCACHE_LOG` environment variable set? That'll make clcache print all kinds of debugging information about what it does.In particular, it should print two lines saying `Parsing given command line...` and `Expanded commandline` -- can you send those lines?

Also, it would be perfect if you could give the actual `cl` command line which is invoked, you can possibly get that by building your CMake project with `make VERBOSE=1`.

sowsem:
Sure. The log variable is already set. Will do when I get  to my laptop later today hopefully! Thx!

webmaster128:
The command line is parsed as 3 arguments only, which is an error

* `/nologo`
* `/c`
* `/I"C:\dev\project\common\include"' /Fd"Bullet2FileLoader.dir\Debug\vc120.pdb" /Gd /TP /wd4201 /wd4512 /analyze- /errorReport:queue "C:\dev\project\src\Bullet3Serialize\Bullet2FileLoader\b3Serializer.cpp"`

What I find strange is that there is exactly one single quote in this part

```
                                 v
[...]

PR #103, opened 2016-06-12
More splitCommandsFile()

webmaster128:
This introduces a generic test method for TestSplitCommandsFile and adds some simple tests.

This is done because it is likely that the methods might need changes, so it is important that current functionality will not break.


frerich:
Looks good - seems we're ready to extend this test now to cover #100 :-)

PR #104, opened 2016-06-12
Group and cleanup command line analysis

webmaster128:
This does not have a functional impact. It just replaces an impossible case with an exception and makes code a lot easier to understand.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -683,121 +683,124 @@ def expandCommandLine(cmdline):
     return ret
 
 
-def parseCommandLine(cmdline):
-    optionsWithParameter = ['Ob', 'Gs', 'Fa', 'Fd', 'Fm',
-                            'Fp', 'FR', 'doc', 'FA', 'Fe',
-                            'Fo', 'Fr', 'AI', 'FI', 'FU',
-                            'D', 'U', 'I', 'Zp', 'vm',
-                            'MP', 'Tc', 'V', 'wd', 'wo',
-                            'W', 'Yc', 'Yl', 'Tp', 'we',
-                            'Yu', 'Zm', 'F', 'Fi']
-    options = defaultdict(list)
-    responseFile = ""
-    sourceFiles = []
-    i = 0
-    while i < len(cmdline):
-        arg = cmdline[i]
+class CommandLineAnalyzer(object):
 
-        # Plain arguments startign with / or -
-        if arg[0] == '/' or arg[0] == '-':
-            isParametrized = False
-            for opt in optionsWithParameter:
-                if arg[1:len(opt) + 1] == opt:
-                    isParametrized = True
-                    key = opt
-                    if len(arg) > len(opt) + 1:
-                        value = arg[len(opt) + 1:]
-                    else:
-                        value = cmdline[i + 1]
-                        i += 1
-                    options[key].append(value)
-                    break
+    @staticmethod
+    def _parseOptionsAndFiles(cmdline):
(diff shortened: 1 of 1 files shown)


frerich:
The diff is not exactly small, so I may be missing something - but how exactly does this simplify the command line analysis conceptually, or the code a lot easier to understand?

It seems to me that this patch mostly introduces a class `CommandLineAnalyzer` which just serves as a namespace (it has no state, and only static methods) and then renames two functions

    parseCommandLine(cmdline) => CommandLineAnalyzer._parseOptionsAndFiles(cmdline)
    analyzeCommandLine(cmdline) => CommandLineAnalyzer.analyze(cmdline)

Can you explain the intention a bit?

webmaster128:
Right, the first idea is to group those functions into a namespace. This allows an easier understanding of what code belongs to *parsing the command line string into a command line argument list* and what code is for *understanding the content of the command line list*. The word "parse" was is a bit confusing here, because it was used for both.

The second point is making clear that parseCommandLine() is just a helper function for analyzeCommandLine(). So putting it together in a class allows making it private.

This was just preparation for the important 3rd commit. Sorting code that way reveals that 1. `responseFile` is unused and 2. `responseFile` must never be set because the case that an argument starts with '@' is impossible as long as there are no bugs.

webmaster128:
The core motivation comes from trying to bugfix the implementation of parseCommandLine() because it only handles a single response file but [cl supports multiple response files](https://msdn.microsoft.com/en-us/library/3te4xt0y.aspx). Turns out this function must not handle response files at all.

frerich:
I see, sounds interesting! I don't *quite* see where you plan to go, but I'm looking forward to learning about it. I'm not exactly satisfied with the current command line parsing logic. Another thing which always trips me off is the place(s) where we strip (or keep) quotation marks, e.g. the

    # Strip quotes around file names; seems to happen with source files
    # with spaces in their names specified via a response file generated
    # by Visual Studio.

at the end of `analyzeCommandLine` bothers me quite a bit. I'd love to have a well-defined pipeline like

1. Expand response file**s**
2. Normalise command line, e.g. I'd love for all of `/Ifoo` and `"/Ifoo"` and `/I"foo"` to get parsed into a tuple `('/I', 'foo')`.
3. Analyse command line

webmaster128:
This PR is quite small when you realize that the first commit is just indent, no code change. It does not touch any of the existing parsing/analyzing structure.

The only thing it does is make clear to the reader, that the code is for Step 3 in your list and has nothing to do with 1 or 2.

webmaster128:
Maybe "simplify" was a too vague title. "Group and cleanup" does fit better.

webmaster128:
#108 depends in this.

frerich:
I don't really like the idea of using classes to implement namespaces -- plain separate Python files do the job just fine. However, factoring this out is still reasonably easy and the insight that there is a condition to be asserted here is really nice, so let's merge this. Thanks a lot for your work!

I also learned about a new GitHub feature: when looking at some diff, add `?w=1` to the URL to ignore whitespace changes. Great stuff for Python code.

webmaster128:
> I don't really like the idea of using classes to implement namespaces -- plain separate Python files do the job just fine.

I am entirely with you but I did not want so start splitting the project in multiple files.

> I also learned about a new GitHub feature: when looking at some diff, add ?w=1 to the URL to ignore whitespace changes. Great stuff for Python code.

Great, thanks for sharing

PR #105, opened 2016-06-12
Add test that covers #100

webmaster128:

PR #107, opened 2016-06-13
Fix making test failures fatal

frerich:


webmaster128:
what about using power shell for upload only?

```
  - py.test --junitxml .\unittests.xml unittests.py
  - ps: >-
      $wc = New-Object 'System.Net.WebClient'
      $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\unittests.xml))

  - py.test --junitxml .\integrationtests.xml integrationtests.py
  - ps: >-
      $wc = New-Object 'System.Net.WebClient'
      $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\integrationtests.xml))
```

frerich:
I considered that, but I wasn't sure how AppVeyor works exactly if one of the commands fails - I wanted to be sure it uploads even if the test fails.

webmaster128:
Okay, looks like it stops at the first failing command: https://ci.appveyor.com/project/frerich/clcache/build/1.0.251/job/jve7bhex070l2gix There the 3rd pylint command, test_script and after_test did not run

webmaster128:
What about renaming `$lastError` to `$testsExitCode`? "last" is not really describing what the variable is supposed to do.

Thanks by the way. I am waiting for this until I continue  #105

frerich:
It eludes me why the CI still fails even though the test passes... the PowerShell script seems to work as intended when executing it locally - I'm just pasting the PS code into a file and then run

    powershell -ExecutionPolicy ByPass -File runtests.ps1

webmaster128:
From the AppVeyor output it looks like $testsExitCode is an empty string.

```
<span class="fg-15 bg-12">Command exited with code </span>
```

No clue how that comes.

webmaster128:
Could you print $testsExitCode before `exit`?

webmaster128:
See last post here: http://help.appveyor.com/discussions/problems/1322-powershell-exit-code-is-failing-build

> I was surprised that doing an exit 0 failed the build [...]

PR #109, opened 2016-06-13
[WIP] Add split test: double quoted with space

webmaster128:
This is a cherry-picked test that I guess does fail independent of #105.

I separate it to make merging #105 possible.

### DIFF
--- a/unittests.py
+++ b/unittests.py
@@ -96,6 +96,12 @@ class TestSplitCommandsFile(BaseTest):
         self._genericTest(r'/Fo"C:\out dir\\" /nologo', [r'/FoC:\out dir' '\\', r'/nologo'])
         self._genericTest(r'/c /Fo"C:\out dir\\" /nologo', ['/c', r'/FoC:\out dir' '\\', r'/nologo'])
 
+    def testDoubleQuotedWithSpace(self):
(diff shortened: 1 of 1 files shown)


unittests.py
```diff
@@ -96,6 +96,12 @@ def testBackslashBeforeQuote(self):
         self._genericTest(r'/Fo"C:\out dir\\" /nologo', [r'/FoC:\out dir' '\\', r'/nologo'])
         self._genericTest(r'/c /Fo"C:\out dir\\" /nologo', ['/c', r'/FoC:\out dir' '\\', r'/nologo'])
 
+    def testDoubleQuotedWithSpace(self):
+        self._genericTest(r'"/Fo"some thing\main.obj""', [r'/Fosome thing\main.obj'])
```
frerich: I think this should actually expand to two arguments,

1. `/Fosome`
2. `thing\main.obj`.

webmaster128:
Those assumptions don't match [the rules](https://msdn.microsoft.com/en-us/library/17w5ykft.aspx) and I do not remember if or where I found them in the wild. So this (failing) tests don't add value right now.

unittests.py
webmaster128: Right, you are

Issue #108, opened 2016-06-13
Cannot open compiler generated file: '"Debug\\".obj' with double backslash in the path


webmaster128:
User @sowsem found this error when compiling with `/Fo"Debug\\" `:

```
ClCompile:
  C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\CL.exe /c /nologo /W0 /WX- /Od /Oy- /D _MBCS /Gm- /EHsc /RTC1 /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Fo"Debug\\" /Fd"Debug\vc120.pdb" /Gd /TC /analyze- /errorReport:queue CMakeCCompilerId.c
  C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\clcache.py Found real compiler binary at 'C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\cl_real.exe'
  C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\clcache.py Parsing given commandline '['@C:\\Users\\davido\\AppData\\Local\\Temp\\tmpf22c581ab5514910a81a2a62e85395c1.rsp']'
  C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\clcache.py Expanded commandline '[u'/c', u'/nologo', u'/W0', u'/WX-', u'/Od', u'/Oy-', u'/D', u'_MBCS', u'/Gm-', u'/EHsc', u'/RTC1', u'/MDd', u'/GS', u'/fp:precise', u'/Zc:wchar_t', u'/Zc:forScope', u'/Fo"Debug\\\\"', u'/Fd"Debug\\vc120.pdb"', u'/Gd', u'/TC', u'/analyze-', u'/errorReport:queue', u'CMakeCCompilerId.c']'
  C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\bin\clcache.py Compiler output file: 'Debug\\'
[...]

PR #111, opened 2016-06-14
Add initial test for CommandLineAnalyzer

webmaster128:
This is a start to work on #108. When it passes, it is ready to merge.


webmaster128:
Looks good now.

frerich:
Very nice, thanks! Reading the tests makes one think of a ton of other interesting tests!

PR #112, opened 2016-06-14
Improve code for ensuring that required directories exist

frerich:
Reduces code duplication, streamlines way to handle existing files (EAFP) and saves on place where directory existence was enforced.


webmaster128:
:+1:

PR #113, opened 2016-06-14
Cleanup AppVeyor config

webmaster128:
Minor adjustments to make AppVeyor just a little bit nicer to read

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -46,7 +46,7 @@ build_script:
 test_script:
   # Run test files via py.test and generate JUnit XML. Then push test results
   # to appveyor.
-  - ps: >-
+  - ps: |
       & py.test --junitxml .\unittests.xml unittests.py
       $testsExitCode = $lastexitcode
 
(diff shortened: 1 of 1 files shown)


frerich:
Why did you convert the line endings? The version in the repository should always use Unix line endings, Git will then take care of converting to Windows line endings when checking it and back to Unix line endings when committing via the `core.autocrlf` setting.

webmaster128:
I tried to fix the issue that multiple lines in the .yml file ended up as one command. But it turns out this is not caused by line endings. So I can remove the commit.

Good to know (UNIX line endings). I occasionally checked in files with Windows line endings. So I'll pick up this policy.

I have `core.autocrlf` set to `false` but that is no big deal since I can use alternative Editors on Windows that work with `"\n"`

webmaster128:
Ah got it, the YAML block style was the cause :) All green here and Windows line endings reverted.

frerich:
Looks good, merged. Thanks!

PR #114, opened 2016-06-14
Improve testing for the use of output files (/Fo) in Analyzer

webmaster128:
This is a next step for dealing with #108

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -783,8 +783,7 @@ class CommandLineAnalyzer(object):
                 outputFile = ''
         else:
             srcFileName = os.path.basename(sourceFiles[0])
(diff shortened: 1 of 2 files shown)


frerich:
Very nice, thanks! Determining the correct output file name used to be a somewhat tricky part in the past - by now, we mostly seem to have it figured out. Having tests for this is really good!

webmaster128:
My pleasure. I am not really sure about what cl does with `/FoDebug/TheOutFile.obj`. Given that a non-Windows separator is used, maybe it makes a filename containing "/" out of it: ".\Debug/TheOutFile.obj". But I don't think so. Having this test setup in place lets us reason very precicely about such edge cases. More tests are about to come.

PR #115, opened 2016-06-15
Normalize relative output path in analyze()

webmaster128:
like cl.exe does as well


frerich:
Looks simple enough - does this actually fix a bug? If so, an addition to the change log might be due.

webmaster128:
No it doesnt. The difference is that outputFile is now normalized but even un-normalized it worked well.

This is just to improve coverage and one step before fixing #108

frerich:
Merged, thanks!

PR #117, opened 2016-06-21
Decode real compiler output from MBCS encoding to string

### LINKED ISSUE #64: python3: encoding error
when using clcache/master via python 3, i encountered the following error:

```
Exception in thread Thread-1:
  File "C:\Python34\lib\threading.py", line 920, in _bootstrap_inner
  File "C:\Python34\lib\threading.py", line 868, in run
  File "C:\Python34\lib\subprocess.py", line 1171, in _readerthread
  File "C:\Python34\lib\encodings\cp1252.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 161099: character maps to <undefined>

  File "clcache.py", line 1200, in <module>
  File "clcache.py", line 1100, in main
  File "clcache.py", line 1166, in processCompileRequest
  File "clcache.py", line 1026, in processNoManifestMiss
  File "clcache.py", line 769, in invokeRealCompiler
  File "C:\Python34\lib\subprocess.py", line 960, in communicate
  File "C:\Python34\lib\subprocess.py", line 1235, in _communicate
IndexError: list index out of range
```

it seems to be related to the fact that the compiler is using a german locale :(

webmaster128:
When a compiler is configured to use another language than English, it is possible that it outputs non-ASCII characters. Those characters are MBCS, an encoding that machtes ASCII for byte values [0,127] but adds
unicode functionality in the upper half [128,255].

> Windows uses a configurable encoding; on Windows, Python uses the name “mbcs” to refer to whatever the currently configured encoding is.

https://docs.python.org/2/howto/unicode.html

mbcs and dbcs are synonyms in this context: https://docs.python.org/3/library/codecs.html#standard-encodings

Here is what Windows means with DBCS: https://www.microsoft.com/typography/unicode/cs.htm

Fixes #64


frerich:
Hm, sounds plausible. I wonder whether this could be tested automatically. One funny idea came to my mind: maybe you could create a source file in a directory with non-ASCII characters and then try to compile

```c++
#error __FILENAME__
```

or so?

webmaster128:
Why not just `#error umlaut: äöü`? The question is if MSVS converts UTF-8 sources into MBCS output.

Maybe that works is a file es explicitly stored as UTF-16 including a UTF-16 BOM. As far as I know, cl oderstands that.

webmaster128:
Great idea to get that tested. It turns out the original commit was good but not sufficient. I am happy with this for now.

clcache.py
```diff
@@ -57,6 +57,16 @@
 
 HashAlgorithm = hashlib.md5
 
+
+# The codec that is used by clcache to store compiler STDOUR and STDERR in
+# output.txt and stderr.txt.
+# This codec is up to us and only used for clcache internal storage.
+# For possible values see https://docs.python.org/2/library/codecs.html
+CACHE_COMPILER_OUTPUT_STORAGE_CODEC = 'mbcs'
```
frerich: If "mbcs" is a system-dependant encoding, wouldn't it be better to use UTF-8 for the cache such that it can safely be shared between multiple systems with potentially different encodings?

clcache.py
```diff
@@ -73,6 +83,12 @@
 Manifest = namedtuple('Manifest', ['includeFiles', 'hashes'])
 
 
+def printBinary(outstream, rawData):
+    with os.fdopen(outstream.fileno(), 'wb') as fp:
+        fp.write(rawData)
+        fp.flush()
```
frerich: According to [this StackOverflow answer](http://stackoverflow.com/a/2447160/91757), closing a file implies flushing it - hence, I think the `fp.flush()` can be omitted.

frerich:
I think this is looking quite good, just some minor things which caught my eye. This stuff can probably get merged Real Soon Now(tm).

clcache.py
webmaster128: Good question. Is "mbcs" system-dependant? I thought it is the same everywhere.

As far as I can see there is no harm in changing this to UTF-8 given that all existing content must be ASCII.

webmaster128: You are right, it is. I already quoted that in the PR.

webmaster128:
Fine, thanks for review.

* Set `CACHE_COMPILER_OUTPUT_STORAGE_CODEC` to UTF-8
* removed extra flush

frerich:
Looks like a nice improvement, thanks a lot for this tedious work! Encoding issues are always a bit tricky (especially with the changes to string types between Python 2 and Python 3).

PR #118, opened 2016-06-22
Add compiler output encoding test

webmaster128:
This adds tests as suggested in #117. If everything works as expected, these tests ~~fail~~ fail with Python 3 and can be fixed by the change in #117.


webmaster128:
Fails beautifully as expected. Cherry-picked in #117.

PR #121, opened 2016-06-23
Add microsoft command line parsing examples as test

webmaster128:
A lot of tests right do not follow the Microsoft way of [Parsing C++ Command-Line Arguments](https://msdn.microsoft.com/en-us/library/17w5ykft.aspx).

Before trying to fix all of those, I think we should concentrate on making the tests from the Microsoft documentation work. Those are very simple cases that look like they cover all of the given rules.

If you, @frerich want to test your tokenizer, feel free to cherry-pick or copy this test and integrate it in your work.


frerich:
Integrated this commit into #122 .

webmaster128:
Beautiful

PR #122, opened 2016-06-24
Drop ad-hoc response file parsing in favor of structured LLVM-based tokenizer

### LINKED ISSUE #108: Cannot open compiler generated file: '"Debug\\".obj' with double backslash in the path

frerich:
Over the last couple of years, there have been numerous bug reports
about how clcache fails to parse response files correctly. This was
always handled in a somewhat ad-hoc manner: we tried to patch the code
such that the newly discovered case works, without breaking the old
ones.

Being dissatisfied with this situation, I looked at how other projects
parse response files and became aware of some code in the LLVM project
solving this exact issue.

The new 'CommandLineTokenizer' class introduced by this patch is a port
of the LLVM function 'cl::TokenizeWindowsCommandLine', slightly
exploiting the fact that functions in Python are nicer first class
citizens (so a 'state' variable is not needed).

This fixes #108 .

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -614,21 +614,6 @@ def printTraceStatement(msg):
         print(os.path.join(scriptDir, "clcache.py") + " " + msg)
 
 
-def extractArgument(argument):
-    # If there are quotes from both sides of argument, remove them
-    # "-Isome path" must becomse -Isome path
-    if not argument:
-        return ""
-
-    if argument == '"':
-        raise Exception("Invalid argument: '\"'")
-
-    if argument[0] == '"' and argument[-1] == '"':
-        argument = argument[1:-1] # Cut first and last character
-
-    return argument.strip()
-
-
 class CommandLineTokenizer(object):
     def __init__(self, content):
         self.argv = []
@@ -704,8 +689,7 @@ class CommandLineTokenizer(object):
 
 
 def splitCommandsFile(content):
-    tokenizer = CommandLineTokenizer(content)
-    return [extractArgument(arg) for arg in tokenizer.argv]
+    return CommandLineTokenizer(content).argv
 
 
 def expandCommandLine(cmdline):
--- a/clcache.py
+++ b/clcache.py
@@ -34,6 +34,7 @@ from __future__ import print_function
 # All string literals are unicode strings. Requires Python 3.3+
 # (http://python-future.org/faq.html#which-versions-of-python-does-python-future-support)
 from __future__ import unicode_literals
+from __future__ import division
 
 from ctypes import windll, wintypes
 import codecs
@@ -692,7 +693,7 @@ class CommandLineTokenizer(object):
 
         followedByDoubleQuote = self._pos < len(self._content) and self._content[self._pos] == '"'
         if followedByDoubleQuote:
-            self._token += '\\' * (numBackslashes / 2)
+            self._token += '\\' * (numBackslashes // 2)
             if numBackslashes % 2 == 0:
                 self._pos -= 1
             else:
--- a/clcache.py
+++ b/clcache.py
@@ -628,33 +628,84 @@ def extractArgument(argument):
     return argument.strip()
 
 
-def splitCommandsLine(line):
-    # Note, we must treat lines in quotes as one argument. We do not use shlex
-    # since seems it difficult to set up it to correctly parse escaped quotes.
-    i = 0
-    wordStart = None
-    insideQuotes = False
-    result = []
-    while i < len(line):
-        if line[i] == ' ' and not insideQuotes and wordStart is not None:
-            result.append(extractArgument(line[wordStart:i]))
-            wordStart = None
-        if line[i] == '"':
-            insideQuotes = not insideQuotes
-        if line[i] != ' ' and wordStart is None:
-            wordStart = i
-        i += 1
+class CommandLineTokenizer(object):
+    def __init__(self, content):
+        self.argv = []
+        self._content = content
+        self._pos = 0
+        self._token = ''
+        self._parser = self._initialState
 
-    if wordStart is not None:
-        result.append(extractArgument(line[wordStart:]))
-    return result
+        while self._pos < len(self._content):
+            self._parser = self._parser(self._content[self._pos])
+            self._pos += 1
+
+        if self._token:
+            self.argv.append(self._token)
+
+    def _initialState(self, currentChar):
+        if currentChar.isspace():
+            return self._initialState
+
+        if currentChar == '"':
+            return self._quotedState
+
+        if currentChar == '\\':
+            self._parseBackslash()
+            return self._unquotedState
+
+        self._token += currentChar
+        return self._unquotedState
+
+    def _unquotedState(self, currentChar):
+        if currentChar.isspace():
+            self.argv.append(self._token)
+            self._token = ''
+            return self._initialState
+
+        if currentChar == '"':
+            return self._quotedState
+
+        if currentChar == '\\':
+            self._parseBackslash()
+            return self._unquotedState
+
+        self._token += currentChar
+        return self._unquotedState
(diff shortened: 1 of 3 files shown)


clcache.py
```diff
@@ -613,33 +613,84 @@ def extractArgument(argument):
...
+
+    def _parseBackslash(self):
+        numBackslashes = 0
+        while self._pos < len(self._content) and self._content[self._pos] == '\\':
+            self._pos += 1
+            numBackslashes += 1
+
+        followedByDoubleQuote = self._pos < len(self._content) and self._content[self._pos] == '"'
+        if followedByDoubleQuote:
+            self._token += '\\' * (numBackslashes / 2)
```
webmaster128: As @hubx [pointed out](https://gist.github.com/frerich/eeb8ffc03a8dfd4750f96b35145efc3a), this causes a

> TypeError: can't multiply sequence by non-int of type 'float'

frerich: Ah, I didn't even know about that comment to my gist! I hope I didn't miss some notification. Fixed now such that it also works in Python 3.

webmaster128:
not too bad, 4 test failing. Would you like to fix those or should I PR test fixes into your branch?

frerich:
I don't care either way really, but I what I'd like to see is some reasoning on *why* the new behaviour is correct. I don't feel terribly comfortable with just using the new actual data as the future expected data. Also, I think *if* the new data is actually sane to expect, then I suspect some other code in clcache needs adjustment too to not bother stripping quotes anymore.

For instance, in [these test results](https://ci.appveyor.com/project/frerich/clcache/build/1.0.333/job/gy8mcnvqx1yv0lmb/tests), I see that `testDoubleQuoted` failed with

```
E   - [u'/Fosomething\\main.obj']
E   ?  -
E   
E   + ['/Fo"something\\main.obj"']
E   ?      +                   +
```

I.e. the quotes are not stripped anymore. Is this actually correct? If so, then there should maybe be another place which strips these quotes?

webmaster128:
In this case, the expected value must be `/Fosomething\\main.obj` because of rules

**1. In-argument quote stripping**

> A string surrounded by double quotation marks ("string") is interpreted as a single argument, regardless of white space contained within. **A quoted string can be embedded in an argument.**

and example 2 (`d"e f"g` -> `de fg`)

**2. Preserved backslashes**

> Backslashes are interpreted literally, unless they immediately precede a double quotation mark.

and example 2 (`a\\b` -> `a\\b`)

It's a bit annoying that [assertEqual](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertEqual) does not define which argument is the actual value and which one is expected. It just checks equality of two things, so the output is not as good as it could be. But well, that's what we get without adding dependencies for other testing frameworks.

webmaster128:
Good job! Just 2 remaining :)

frerich:
I hope it all passes now. The most difficult part was to justify the changes based on the rules. In the end, I managed to remember the rules from the top of my head though and came up with 'new' expected data just by looking at the input string.

frerich:
I think it's also evident that the `testBackslashBeforeQuote` test is a pathological case. It seems to assume that `/nologo` should have been a separate argument, but it's not.

frerich:
The only part which does not seem to be explicitly defined by the [parsing rules](https://msdn.microsoft.com/en-us/library/17w5ykft.aspx) is how to handle

```
  a"b
```

I.e. what if a double quote is not escaped but also not closed. Right now, the parser considers this to be equivalent to

```
  a"b"
```

i.e. the double quote is closed implicitly on the end of input and it is dropped from the output. However, I guess it can be argued that it may just as well be equivalent to

```
  a\"b
```

I.e. it's treated as a literal double quote in this case.

webmaster128:
Both possibilities you mentioned are plausible.

However, I think we can leave this just as undefined behavior, since it is a crazy edge case.

I would change the input value to `/Fo"C:\out dir\\"`.

frerich:
I'm not sure it's "invalid". It may be undefined though. I just tested what happens if I have a file `resp.rsp` with

```
/c foo.c /Fo"bar yo
```

and then there's an existing directory `bar yo`. Running

```
cl @resp.rsp
```

does not complain, but it also doesn't create `bar yo\foo.obj`. Instead, it creates `bar yo.obj`. :-]

PR #126, opened 2016-06-24
Add a lot of /Fo testing

webmaster128:
This adds most tests from my previous PR but respects the current state of command line splitting.

I removed all implementation changes for now to check if those are still needed. I guess some normalization of the `outputFile`  will be necessary.

~~I cherry-picked, #125 assuming it will be on master soon. Then I'll remove it from my branch.~~

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -814,13 +814,13 @@ class CommandLineAnalyzer(object):
         outputFile = None
         if 'Fo' in options:
             outputFile = options['Fo'][0]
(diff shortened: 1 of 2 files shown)


webmaster128:
Check. Those tests now pass with just a minimal implementation change thanks to the command line splitting, which handles all the quoting.

frerich:
Thanks, merged! I suppose parsing-wise, we're a lot better now. I think I'll do a 3.1.1 soon.

PR #127, opened 2016-06-25
Use the same future imports in all Python files

webmaster128:


integrationtests.py
```diff
@@ -163,17 +166,20 @@ def testSampleproject(self):
         with cd(os.path.join("tests", "precompiled-headers")):
             cpp = PYTHON_BINARY + " " + CLCACHE_SCRIPT
 
+            # Note: explicit str() because environment needs native str type in Python 2 and Python 3
```
frerich: Thanks for this comment!

tests/parse-includes/compiler_output.txt
```diff
@@ -1,5 +1,5 @@
 version.cpp
-Note: including file: C:\Users\me\test\smartsqlite\include\smartsqlite/version.h
+Note: including file: C:\Projects\test\smartsqlite\include\smartsqlite/version.h
```
frerich: This part I don't get -- why did you replace `Users\me` with `Projects` all over?

webmaster128:
When `unicode_literals` is set, `\U` followed by 8 hex digits [is interpreted as a unicode codepoint literal](https://docs.python.org/2/howto/unicode.html#unicode-literals-in-python-source-code).

In Python 2, an error is raised when the string after the `\U` is not a valid hex string:

Python 3: https://repl.it/C6BC
Python 2: https://repl.it/C6BB

I guess there is a way to escape the backslash, such that the unicode literal does not apply, but in favor of readability, I just changed the path.

frerich:
Thanks for the explanation, I think it would be great if you could mention those peculiarities of `\U` in the commit log. :-}

frerich:
Merged, thanks!

PR #128, opened 2016-06-25
Deduplicate basenameWithoutExtension()

webmaster128:


clcache.py
```diff
@@ -89,6 +89,11 @@ def printBinary(rawData, stream):
         fp.write(rawData)
 
 
+def basenameWithoutExtension(path):
+    basename = os.path.basename(path)
+    return os.path.splitext(basename)[0]
+
+
```
frerich: Seeing these use cases, how about a slightly more powerful function like

```python
def replaceInPath(path, extension=None, dirname=None):
    if extension is not None:
        path = os.path.splitext(path)[0] + extension
    if dirname is not None:
        path = os.path.join(dirname, os.path.basename(path))
    return path
```

...because you could then write...

clcache.py
```diff
@@ -816,9 +821,7 @@ def analyze(cmdline):
             outputFile = options['Fo'][0]
 
             if os.path.isdir(outputFile):
-                srcFileName = os.path.basename(sourceFiles[0])
-                outputFile = os.path.join(outputFile,
-                                          os.path.splitext(srcFileName)[0] + ".obj")
+                outputFile = os.path.join(outputFile, basenameWithoutExtension(sourceFiles[0]) + ".obj")
```
frerich:
```python
outputFile = replaceInPath(sourceFiles[0], extension='.obj', dirname=outputFile)
```

clcache.py
```diff
@@ -827,16 +830,13 @@ def analyze(cmdline):
                 if 'Fi' in options:
                     outputFile = options['Fi'][0]
                 else:
-                    srcFileName = os.path.basename(sourceFiles[0])
-                    outputFile = os.path.join(os.getcwd(),
-                                              os.path.splitext(srcFileName)[0] + ".i")
+                    outputFile = os.path.join(os.getcwd(), basenameWithoutExtension(sourceFiles[0]) + ".i")
```
frerich:
```python
outputFile = replaceInPath(sourceFiles[0], extension='.i', dirname=os.getcwd())
```

clcache.py
```diff
@@ -827,16 +830,13 @@ def analyze(cmdline):
...
             else:
                 # Preprocess to stdout. Use empty string rather then None to ease
                 # output to log.
                 outputFile = ''
         else:
-            srcFileName = os.path.basename(sourceFiles[0])
-            outputFile = os.path.splitext(srcFileName)[0] + ".obj"
+            outputFile = basenameWithoutExtension(sourceFiles[0]) + ".obj"
```
frerich:
```python
outputfile = replaceInPath(sourceFiles[0], extension='.obj', dirname='')
```

webmaster128:
I thought about the `rewritePath` signature a bit but I am not getting comfortable with it. Mostly because I don't think it improves clarity at the caller side. While every experienced programmer knows what a "basename" function is (e.g. from Unix command line), `rewritePath` could mean everything and you need to see the implementation to know what is going on. Additionally, basename does most of the deduplication that makes sense for now: 1. remove unnecessary function scope variable, 2. move irritating [0] call from `os.path.splitext` out of the way, 3. reduce three lines to one. The remaining concatenations of paths and extension in the caller code are in my opinion helping to read the code and understand which variable is a relative/absolute file/directory path.

Btw. I came across the basename issue because I think there is a lot of potential to pull out code that handles user input paths. I think `/Fi` must be handles the same way as `/Fo`, with just another extension. This may apply to other arguments as well. So this is the greater goal and the real deduplication and consistency target. But to get there, I need to be able to better understand the analyze() body.

frerich:
Sure, `rewritePath` is a fairly generic name. I intentionally invoked it with keyword arguments, because something like

```python
outputFile = rewritePath(sourceFiles[0], extension='.i', dirname=os.getcwd())
```

actually doesn't seem bad to me: some path is apparently rewritten, and an extension and a dirname are passed.

`basenameWithoutExtension` just seemed to be a somewhat long name, so I'm trying to tip-toe around it. The `rewritePath` function is actually a generalisation of another function I first wanted to propose,

```python
def replaceExtension(path, extension):
  return os.path.splitext(path)[0] + extension
```

Using this, one could change

```python
srcFileName = os.path.basename(sourceFiles[0])
outputFile = os.path.splitext(srcFileName)[0] + ".obj"
```

to

```python
outputFile = replaceExtension(os.path.basename(sourceFiles[0]), '.obj')
```

I think this would satisfy the same three benefits you mention (_1. remove unnecessary function scope variable, 2. move irritating [0] call from os.path.splitext out of the way, 3. reduce three lines to one._) but it would avoid the long `basenameWithoutExtension`. It's also somewhat readable, I think. What do you think?

webmaster128:
I was also thinking about some kind of replaceExtension method, but what I disliked was that it is not entirely obvious, which path is relative/absolute/filename-only.

In order to make the lines shorter, I propose the following signature:

```
def basename(path, newExtension=None)
    out = os.path.basename(path)
    if newExtension is not None:
        out =  os.path.splitext(out)[0] + newExtension
    return out
```

This reduces the characters by 25 compared to your idea and by 16 compared to my version in the original PR:

```
outputFile = replaceExtension(os.path.basename(sourceFiles[0]), '.obj')
outputFile = basenameWithoutExtension(sourceFiles[0]) + ".obj"
outputFile = basename(sourceFiles[0]), '.obj')
```

If called with `newExtension=None`, the extension should not be cut. So `basename(path)` works the same way as Python's `os.path.basename` and UNIX command line basename. My previous basenameWithoutExtension is then called via `basename(path, '')`.

This should cover all cases, satisfies 1./2./3., is short and leaves path concatenation out of the business.

frerich:
> I was also thinking about some kind of replaceExtension method, but what I disliked was that it is not entirely obvious, which path is relative/absolute/filename-only.

Not sure I understand. Why would it matter, whether a path is relative or absolute? I'd expect that a `replaceExtension` function simply replaces the file extension. Besides the usual ambiguity of 'extension' (is the extension of `foo.obj.pdb` just `.pdb` or `.obj.pdb`?), it seems to me that it should work independent of whether the path is relative or absolute.

frerich:
By the way, a nice little GitHub trick: if you mark your code snippets with three back tickets and then write `python`, as in

    ```python

you get nice syntax highlighting.

webmaster128:
> Not sure I understand. Why would it matter, whether a path is relative or absolute? I'd expect that a replaceExtension function simply replaces the file extension.

Well, yes. If you don't let it do the basename part. Of course one can do `replaceExtension(os.path.basename(sourceFiles[0]), '.obj')` but that does not help with code length.

So the options now are

```python

outputFile = replaceExtension(os.path.basename(sourceFiles[0]), '.obj')
outputFile = basenameWithoutExtension(sourceFiles[0]) + ".obj"
outputFile = basename(sourceFiles[0]), newExtension='.obj')

outputFile = os.path.join(outputFile, replaceExtension(os.path.basename(sourceFiles[0]), ".obj"))
outputFile = os.path.join(outputFile, basenameWithoutExtension(sourceFiles[0]) + ".obj")
outputFile = os.path.join(outputFile, basename(sourceFiles[0], newExtension=".obj"))
```

I am fine with any of those.

frerich:
I think what bothers me about `basenameWithoutExtension` is that we have no code which wants the base name without the file extension. What we do want however is a base name with a different extension. Mangling that functionality (selecting a new extension) into a function called `basename` seems inappropriate do, exactly because the reason you gave: `basename` has a pretty conventional meaning already. My feeling is that adding optional support for replacing the file extension makes it move towards a generic `rewritePath` function (which did the same, except that it also permitted changing the dirname component). It makes it do more than one thing.

It may be true that `replaceExtension` composed with `os.path.basename` is not the shortest variant, but I think it nicely captures what we want (replacing the extension of some path's base name). However, if in doubt I tend to favor 'He who codes decides.', so I can only hope that I managed to give some convincing case for my preferred version, but I'd leave the final call to you.

webmaster128:
I share the concerns about basename with extended functionality.

While `replaceExtension` together with `os.path.basename` is an elegant an solid way that beautifully defines responsibilities, it leaves all the `srcFileName` variables existing due to line length:

```python
                srcFileName = os.path.basename(sourceFiles[0])
                outputFile = os.path.join(outputFile, replaceExtension(srcFileName, ".obj"))


                    srcFileName = os.path.basename(sourceFiles[0])
                    outputFile = os.path.join(os.getcwd(), replaceExtension(srcFileName, ".i")


            srcFileName = os.path.basename(sourceFiles[0])
            outputFile = replaceExtension(srcFileName, ".obj")
```

I definitely see your point about basenameWithoutExtension but I think being able to kick out this local `srcFileName` variables is worth it.

So I am back at my original commit.

frerich:
Alright then, you mad a call - and I don't feel strongly enough to outweigh the fact that you actually wrote code. So let's merge this! Thanks!

PR #130, opened 2016-06-27
Move all test files into either tests/unittests/ or tests/integrationtests/

webmaster128:
this gives a better overview about what is used where. In addition
ASSETS_DIR can be used as per-module constant setting the parent folder,
which removes a lot of magic paths in the test implementation.

This does not change any functionality.


frerich:
Thanks, looks good. I suspect a simple

```python
def assetPath(path):
  return os.path.join(ASSET_DIR, path)
```

might simplify things even further, seeing how often that `os.path.join(ASSET_DIR, ...` construct is repeated. Just a minor thing though.

webmaster128:
Good idea. This is the kind of thing you start to see after basic cleanup.

PR #131, opened 2016-06-27
Use relative path to preprocess output file (.i file) when generated from source file

webmaster128:
relative files in outputFile are supported and well tested.

This simplifies code and testing but does not fix a bug.

Actually it is one step before fixing a real bug (`/Fo` must be ignored when preprocessing is used). Follow-up tests are about to come.


frerich:
Looks good, merged. Thanks!

Issue #30, opened 2015-04-07
We always use /Zi, can clcache be adapted to copy with this?


broesdecat:
Hi all,

In our company, /Zi is enabled by default for our development builds.
Is there a way clcache (and/or our workflow) can be adapted so we can use it, or has any of the existing forks already done this?

Thanks a lot!
Broes

frerich:
I think that technically it's perfectly possible. As far as I know, the only requirement is that clcache copies the generated {{.pdb}} files into/out of the cache. In fact, there's a source code comment which says the same: https://github.com/frerich/clcache/blob/master/clcache.py#L642

inorton:
As far as I know it isn't possible, partly because pdb files are like databases. As I understand it, if you compile ten objects with the same pdb target then the pdb will only be good if none of the ten objects are rebuilt. If you rebuilt only one object, the pdb would be updated again and would be invalid for the other unaffected objects.

The best solution I had was to force Zy until the target required was a dll or exe and use Zi then. The effect is that you get the pdb you want and get caching at the object level.

frerich:
Not sure I understand @inorton : you can have one .pdb file per object by using the '/Fd' switch (https://msdn.microsoft.com/en-us/library/9wst99a9.aspx) no? I use that in some projects, having every single 'cl' invocation generate one .obj and one .pdb file. I was thinking that this could be used for the cache.

webmaster128:
@frerich Hot stuff. By adding `/Zi /Fd$@.pdb` globally to compile flags, one can easily set this up. You just have to replace nmake with [jom](https://wiki.qt.io/Jom), because nmake does not like the `$@` shortcut in batch mode.

So what you describe looks very doable at least for qmake projects.

inorton:
Hiya, sorry, I didn't mean to say it was impossible. Rather that if the pdb file exists already on disk the compiler will merge the new pdb data into the file so you can only cache if the pdb file doesn't exist yet and only score hit if it is absent

tfar:
I a lot merged PRs related to this issue. What's the current state? We also sue /Zi on our windows builds and it the latest release no cache seems to be created.

frerich:
@tfar A lot of work was done on making cache accesses work better (and analysing the command line such that we notice that external `.pdb` files are generated in the first place...). `/Zi` itself is still unsupported, but I think we're in a better position to implement support for it than ever before. Hence this issue is still open.

TiloW:
[...]

PR #132, opened 2016-06-27
Fix bug in analyzing preprocessor arguments

webmaster128:


unittests.py
```diff
@@ -183,6 +186,32 @@ def testOutputFileFromSourcefile(self):
...
+        # to stdout (/EP)
+        self._testPreprocessingOutfile(['/c', '/EP', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/EP', '/FoSome.obj', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/EP', '/FaListing.asm', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/EP', '/FmMapfile.map', 'main.cpp'], None)
+        # to stdout (/P /EP)
+        self._testPreprocessingOutfile(['/c', '/P', '/EP', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/P', '/EP', '/FoSome.obj', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/P', '/EP', '/FaListing.asm', 'main.cpp'], None)
+        self._testPreprocessingOutfile(['/c', '/P', '/EP', '/FmMapfile.map', 'main.cpp'], None)
```
frerich: I'm not sure these last four tests are correct, for me (using Visual Studio 2013), passing both `/P` and `/EP` makes the compiler write the preprocessed output to an `.i` file, too (i.e. `/P` overrules `/EP`). Note that this contradicts the documentation [which says](https://msdn.microsoft.com/en-us/library/becb7sys(v=vs.120).aspx)

> To send the preprocessed output to stdout, with #line directives, use /P and /EP together.

clcache.py
```diff
@@ -817,29 +817,27 @@ def analyze(cmdline):
...
+            else:
+                # Preprocess to file
                 if 'Fi' in options:
                     outputFile = options['Fi'][0]
                 else:
                     outputFile = basenameWithoutExtension(sourceFiles[0]) + ".i"
-            else:
-                # Preprocess to stdout. Use empty string rather then None to ease
-                # output to log.
-                outputFile = ''
```
frerich: Hm, I see commit a3771033 intentionally did not use `None` here, but I don't quite understand the comment. I suspect there was some string concatenation or so going on which `None` prohibits? In any case, this doesn't seem to hurt... I'm just wary of reverting code which has a comment attached to it explaining that things are done intentionally.

frerich:
Thanks, looks good except for the part where the tests verify the documented instead of the actual (at least for me) behaviour of the compiler.

I guess eventually we'll want to support a list of output files in order to implement support for #30.

clcache.py
webmaster128: Ah thanks for looking at the source. [When vchigrin commited this](https://github.com/frerich/clcache/blob/a37710335e62e98a3140757743fa161f2d818057/clcache.py), we introduced the first non-path outputFile value and the rest of the code was not ready to deal with it. [Existing](https://github.com/frerich/clcache/blob/a37710335e62e98a3140757743fa161f2d818057/clcache.py#L777) [tracing](https://github.com/frerich/clcache/blob/a37710335e62e98a3140757743fa161f2d818057/clcache.py#L791) and [os path](https://github.com/frerich/clcache/blob/a37710335e62e98a3140757743fa161f2d818057/clcache.py#L779) [methods](https://github.com/frerich/clcache/blob/a37710335e62e98a3140757743fa161f2d818057/clcache.py#L790) could not deal with `None`. So he did a plausible thing in order to get a minimal commit.

unittests.py
webmaster128: Nice one, thanks! I have the same behavior using MSVS2015. /P is with #line annotations and /P /EP is without – both go to file.

webmaster128:
Thanks for review. I think I'll add a small helper function `_preprocessToStdout` in which I document the documentation bug.

Yes, #30 is one big goal. Beeing unable to break existing functionality by excessive testing is another.

webmaster128:
Check

* cleaned up everything
* changed test cases, implementation and changelog for the /E /EP case
* pulled out preprocessToStdout that explicitly takes care of the documentation bug

Next up: test more different /Fi values, detect bug when a directory is used, fix implementation, handle /Fo and /Fi using a common helper.

clcache.py
```diff
@@ -737,6 +737,21 @@ def expandCommandLine(cmdline):
...
+        # Note: For MSVS 2013 and 2015 there is a documentation bug in
+        # https://msdn.microsoft.com/en-us/library/becb7sys.aspx stating that
+        # "To send the preprocessed output to stdout, with #line directives,
+        # use /P and /EP together."
+        # The actual compiler behavior is that both /P and /P /EP go to file.
+        # /P is with #line annotations and /P /EP is without.
+        if 'P' in options:
+            return False
+        else:
+            return True
```
frerich: Is the `assert` maybe a bit harsh? I think Python permits expressing whether some options cause preprocessor output to be printed to `stdout` in almost plain english:

```python
def _preprocessToStdout(options):
    return ('E' in options or 'EP' in options) and 'P' not in options
```

I think that reads very nicely.

frerich:
Looks fine to me except for the somewhat elaborate definition of `_preprocessToStdout`.

clcache.py
webmaster128: My rule of thumb is to place assertions whenever a failing expression is a programming error.

This method must never be called when not in preprocessing because the entire logic only woks there. So instead of pushing the error down the road where it gets harder to debug, we should fail fast.

I am fine with rewriting the implementation in one line but I think the assertion adds value.

frerich: I agree on the value of assertions in general, but for a method call `_preprocessToStdout` I think it's perfectly fine if it yields `True` if the given options indicate that preprocessed output will be printed to `stdout` and `False` otherwise. I.e. I think it's plausible that this method returns `False` for e.g. empty lists.

webmaster128: ok

webmaster128:
Alright, alternative implementation of  preprocessToStdout() is in

frerich:
Thanks, looks excellent - merged!

PR #133, opened 2016-06-28
Fix /Fi analysis

webmaster128:


clcache.py
```diff
@@ -748,13 +748,17 @@ def _preprocessToStdout(options):
...
+        optionsWithParameter = [
+            # length 3
+            'doc',
+            # length 2
+            'AI', 'Fa', 'FA', 'Fd', 'Fe', 'Fi', 'FI', 'Fm', 'Fo',
+            'Fp', 'Fr', 'FR', 'FU', 'Gs', 'MP', 'Ob', 'Tc', 'Tp',
+            'vm', 'wd', 'we', 'wo', 'Yc', 'Yl', 'Yu', 'Zm', 'Zp',
+            # length 1
+            'D', 'F', 'I', 'U', 'V', 'W',
+        ]
```
frerich: Good catch! However, I think it would be preferable to rather just add the line

```python
optionsWithParameter.sort(key=len, reverse=True)
```

The diff is smaller, and it's harder to break this in case we need to extend the list of options.

clcache.py
```diff
@@ -747,14 +747,32 @@ def _preprocessToStdout(options):
...
+            # Handle user input
+            outputFile = options[argumentName][0]
+            outputFile = os.path.normpath(outputFile)
+
+            if os.path.isdir(outputFile):
+                outputFile = os.path.join(outputFile, basenameWithoutExtension(sourceFile) + extension)
+        else:
+            # Generate from .c/.cpp filename
+            outputFile = basenameWithoutExtension(sourceFile) + extension
+        return outputFile
```
frerich: I think if you rewrite this function slightly such that at the top, you define (and then use)

```python
inferredOutputFile = basenameWithoutExtension(sourceFile) + extension
```

you end up with only this one place where `basenameWithoutExtension` is used in this file, in which case it may be interesting to just inline the function here, i.e. define

```python
inferredOutputFile = os.path.splitext(os.path.basename(sourceFile))[0] + extension
```

...and then move the `basenameWithoutExtension` function out into `unittests.py`.

clcache.py
```diff
@@ -747,14 +747,32 @@ def _preprocessToStdout(options):
         return ('E' in options or 'EP' in options) and 'P' not in options
 
     @staticmethod
+    def _parsePathArgument(options, argumentName, sourceFile, extension):
```
frerich: I think this method name is unfortunate since
* `parsePathArgument` is quite generic (I'd try to add the word `Output` somewhere) and
* the method doesn't actually only parse a path argument, it also does something (inferring an output file name from the given source file name) if the argument is not specified at all.

How about renaming this to `_determineOutputFile`?

clcache.py
webmaster128: Yes, I see basically 2 good ways:

1. Prepare a ready-to use list literal and make sure it is correct (using comments)
2. Define optionsWithParameter as a set because it is the better type here (unsorted, elements contained only once), take it as an implementation detail that the algorithm needs a sorted list and use the sorting you posted.

Both are fine for m, feel free to choose one

clcache.py
webmaster128: I don't like introducing this extra variable, which I only need in two of three branches. I don't see a lot of benefit in deduplicating a function call and a string concatenation.

clcache.py
webmaster128: :+1: I was not happy with this as well

clcache.py
frerich: I'd very much prefer 2. (indeed, a `set` is a more appropriate type here) since -- at least without benchmarks -- I value code which does not need comments to express the intention higher than the performance benefit from not having to sort the list at runtime.

clcache.py
frerich: Okay, I don't really feel strongly about this, so let's go with what you propose.

webmaster128:
Thanks for critical review!

Btw.: let's find a consistent wording that should be used in all future code: (command line) "option" or "argument"? I have the feeling that "option" is the common one in MSVS, but I do not really care which one is used.

frerich:
Glad to hear the code reviews are not frustrating you (yet? :-]). I want you to know that I'm extremely grateful for your contributions. You're not only (by far) the most productive contributor to this project, you're also picking jobs which are somewhat ungrateful. Setting up CI infrastructure, adding tests, refactoring etc.. Usually, people like doing the crazy moonshot things (say, caching linker invocations, server processes which clcache talks to etc.). :-)

As for the wording: my understanding is that an *option* is something which is *optional*, i.e. which can be omitted in which case some default behaviour is selected. In that sense, it's an "option" as in a "choice".


*argument* eems more generic (and more appropriate as far as `clcache` is concerned): it's simply something which is passed to a program without any associated semantic such as whether it's a path or a switch (I think a "switch" would be a binary setting).

Based on this armchair theorising, I think I favour the term "argument".

webmaster128:
Okay, I think I got everything in place

frerich:
Looks good, merged. Thanks!

PR #134, opened 2016-06-28
Make local variable `sourceFiles` plural; let analyze() return list

webmaster128:
analyze() returned a list of source files or None in all cases but
AnalysisResult.Ok. This commit ensures that the second return value is
always a list. This is important because the second return value is
stored in a local variable at a point where we don't yet know the
analysis result. This local variable now has a proper plural name
`sourceFiles`.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -852,7 +852,7 @@ class CommandLineAnalyzer(object):
             outputFile = CommandLineAnalyzer._outputFileFromArgument(options, 'Fo', sourceFiles[0], '.obj')
 
         printTraceStatement("Compiler output file: {}".format(outputFile))
-        return AnalysisResult.Ok, sourceFiles[0], outputFile
+        return AnalysisResult.Ok, sourceFiles, outputFile
 
 
 def invokeRealCompiler(compilerBinary, cmdLine, captureOutput=False):
@@ -1249,10 +1249,10 @@ def processCompileRequest(cache, compiler, args):
 
     cmdLine = expandCommandLine(sys.argv[1:])
     printTraceStatement("Expanded commandline '%s'" % cmdLine)
-    analysisResult, sourceFile, outputFile = CommandLineAnalyzer.analyze(cmdLine)
+    analysisResult, sourceFiles, outputFile = CommandLineAnalyzer.analyze(cmdLine)
 
     if analysisResult == AnalysisResult.MultipleSourceFilesSimple:
-        return reinvokePerSourceFile(cmdLine, sourceFile), '', ''
+        return reinvokePerSourceFile(cmdLine, sourceFiles), '', ''
 
     if analysisResult != AnalysisResult.Ok:
         with cache.lock:
@@ -1280,7 +1280,7 @@ def processCompileRequest(cache, compiler, args):
     if 'CLCACHE_NODIRECT' in os.environ:
         return processNoDirect(cache, outputFile, compiler, cmdLine)
     else:
-        return processDirect(cache, outputFile, compiler, cmdLine, sourceFile)
+        return processDirect(cache, outputFile, compiler, cmdLine, sourceFiles[0])
 
 
 def processDirect(cache, outputFile, compiler, cmdLine, sourceFile):
--- a/clcache.py
+++ b/clcache.py
@@ -516,9 +516,9 @@ class CacheStatistics(object):
 
 
 class AnalysisResult(object):
-    Ok, NoSourceFile, MultipleSourceFilesSimple, \
+    Ok, NoSourceFile, \
         MultipleSourceFilesComplex, CalledForLink, \
-        CalledWithPch, ExternalDebugInfo = list(range(7))
+        CalledWithPch, ExternalDebugInfo = list(range(6))
 
 
 def getCompilerHash(compilerBinary):
@@ -841,7 +841,7 @@ class CommandLineAnalyzer(object):
         if len(sourceFiles) > 1:
             if compl:
                 return AnalysisResult.MultipleSourceFilesComplex, None, None
-            return AnalysisResult.MultipleSourceFilesSimple, sourceFiles, None
+            return AnalysisResult.Ok, sourceFiles, None
 
         if preprocessing:
             if CommandLineAnalyzer._preprocessToStdout(options):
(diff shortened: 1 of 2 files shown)


frerich:
I'm not sure this is plausible. At this point, `AnalysisResult.Ok` implies that there is only one source file (this stems from the fact that originally, only invokations with exactly one source file were supported). If there are multiple source files, `AnalysisResult.MultipleSourceFilesSimple` is returned.

I think there would be more conceptual integrity if `MultipleSourceFilesSimple` is dropped and `AnalysisResult.Ok` is used (but it now returns a list of source files).

webmaster128:
I don't like the idea that one function returns different types in the seconds value depending on the first value. This is something you might be allowed to do in Python but I would not call it a feature.

In this case it causes real trouble because we assign a local variable and do not know if it should be singular (string) or plural (list).

The conceptually proper solution would be to have return types, that take sourceFiles as an Argument

```python
class AnalysisResult(object):
    pass

class AnalysisResultOk(AnalysisResult):
    def __init(self, sourceFile, outputFile)__:
        self.sourceFile = sourceFile
        self.outputFile = outputFile

class AnalysisResultMultipleSourceFilesSimple(AnalysisResult):
    def __init(self, sourceFiles)__:
        self.sourceFiles = sourceFiles

class MultipleSourceFilesComplex(AnalysisResult)
    # No arguments
    pass
```

frerich:
I don't care strongly whether you express the type of the analysis result as a Python class or as an enumeration flag which is the first element of a tuple. It amounts to the same information really.

The purpose of `AnalysisResult.Ok` is to indicate that the code detected a compiler invocation which can be handled by `clcache`, i.e. it's not a failure condition. My main point is that I believe that there's no point in distinguishing `AnalysisResult.Ok` and `AnalysisResult.MultipleSourceFilesSimple` if the former also transports a list of source files (like the latter), no matter whether you express this via classes or by other means.

I think that if you make `AnalysisResult.Ok` transport a list of source files, then you may just as well drop the `MultipleSourceFilesSimple` type and use `Ok` there, too. The two values are apparently exactly the same, except that at the top of `processCompileRequest` you'd do:

```python
if analysisResult == AnalysisResult.Ok:
  if len(sourceFiles) > 1:
    reinvokePerSourceFile(..)
  else:
    procesDirect/indirect call
else:
  logFailureCause(cache, analysisResult)
  invokeRealCompiler()

That being said, *if* it is deemed appropriate to replace the current 3-tuples with classes at some point, I think I'd rather vote for just two classes to model `Success` (in which case a list of source files and output file(s) is returned) or `Failure` (in which case a plain constant is returned to indicate the error type).

**Comment by webmaster128** [comment]
Merging Ok and MultipleSourceFilesSimple is fine here

**Comment by webmaster128** [comment]
After thinking about this for some time I always need the differentiation Ok and MultipleSourceFilesSimple in my head. Maybe the namings are not optimal. What I have in my mind is differentiating "I can do that directly" and "I need to split this task into smaller ones and then I can make it". In my option this is a huge difference when trying to understand the caller and the callee code, because I do not ask analyze() the yes/no question

> Can I handle the command line?

but

> Please help me understand what kind of command line we got here

So I think there are tree big classes of results:

* CannotHandle
* ReinvocationNecessary
* Ok

Given that, I favour the following result class hirarchy

```
- AnalysisResult
  \- CannotHandle
     \- MultipleSourceFilesComplex
     \- NoSourceFile
     \- ExternalDebugInfo
     \- CalledWithPch
     \- CalledForLink
  \- ReinvocationNecessary
     \- MultipleSourceFilesSimple
  \- Ok
```

The additional data passed from analyze() heavily depends on the case. For CannotHandle types no data should be passed, MultipleSourceFilesSimple gets `sourceFiles` list and no `outputFile` and Ok gets a single `sourceFile` string and an `outputFile`.

**Comment by frerich** [comment]
I'm not sure I understand, are you suggesting to replace the AnalysisResult enumeration which is currently defined over four lines with a class hierarchy made of 10 classes, three levels deep and which will probably amount to at least 30 lines of code (assuming you need no more than three lines per class on average)? I'm not sure this is moving in the right direction. :-]

I think what `analyze` provides is *either* a pair of input/output files to consider for the cache access *or* a reason why the invocation cannot be handled. This is a typical case of a 'sum type'. There's a package to represent this called [`either`](https://pypi.python.org/pypi/either/0.2), but I a much simpler solution for a *dynamically typed language* like Python is to use two types:

```python
class SupportedInvocation(object):
  def __init__(self, sourceFiles, objectFile):
    self.sourceFiles = sourceFiles
    self.objectFile = objectFile

class UnsupportedInvocation(object):
  NoSourceFile, MultipleSourceFilesComplex, CalledForLink, \
  CalledWithPch, ExternalDebugInfo = list(range(5))

  def __init__(self, reason):
    self.reason = reason
```

The callee, i.e. `_analyze` thus either does `return SupportedInvocation(sourceFile, objectFile)` or it does something like `return UnsupportedInvocation(UnsupportedInvocation.CalledForLink)`.

The caller would do

```python
result = _analyze(...)
if isinstance(result, SupportedInvocation):
  # process result.sourceFile and result.objectFile
else:
  # process result.reason
```

frerich:
Now, after checking back with some other Python people, I'm reassured that exceptions are not only ubiquitous in Python, they are also used for non-error (non-exceptional) cases and even control flow (see e.g. the `StopIteration` exception).

With this in mind, we could slightly adjust the concept of `_analyze` such that it's actually called `_parseInputOutputFiles`. It returns a 2-tuple with the input and output file(s). In case the given invocation cannot be handled, it raises an `UnsupportedInvocation` exception which has ` reason` field giving the reason for why the invocation is inappropriate.

frerich:
I guess `analyze` is a somewhat poor name for a function in the first place. :-]

webmaster128:
Whatever type system we use, you're saying that `ReinvocationNecessary` is the same thing as `Ok`. I still disagree because I think we loose information – but okay.

How would you handle preprocessing multiple source files (e.g. `cl /c /ES file1.c file2.c`)? The current implementation would handle them in parallel, which is most likely to be a bug.

frerich:
I don't think we lose information actually, because the list of source files returned has more than one element. I'm currently working on a first PR to illustrate my proposal. :-)

As for preprocessing multiple source files: good question! I suspect processing them in parallel would be buggy indeed. However, before simply making it sequential: I wonder whether there is actual benefit in caching such invocations at all. I see that we currently actually do not (contrary to what the `README.md` file says!) require that `/c` is specified, we also cache preprocessor invocations. I suspect this isn't actually worth it (the preprocessor tends to be a lot faster than the compiler, and I suspect that our work needed for determining whether an invocation can be cached at all adds an unreasonable overhead to the preprocessor call), so maybe we should just pass-through?

webmaster128:
> maybe we should just pass-through

:+1: You can add this case in your PR

frerich:
I submitted #138 now which implements the exception-based idea I mentioned earlier. It's mostly a proof-of-concept, the code works and the tests pass, but it's just one big commit. In hindsight, I guess

* Renaming the method
* Making it return a list of source files
* Porting away from AnalysisResult towards exceptions

could be done in three steps. It's just to get some code to toy around with.

webmaster128:
check

frerich:
Looks great, thanks!

PR #136, opened 2016-06-29
Test parsing of arguments starting with backslash

frerich:
Looking at the coverage of the TestSplitCommandsFile test case showed
that the only statements in the CommandLineTokenizer which were not
covered yet were the ones handling the case that some argument *starts*
with a backslash.

A real-life scenario of this situation is when source files are in the
root directory (not uncommon when using the 'subst' command), e.g.

  C:\>subst X: C:\Users\Frerich\src\fooproject

  C:\>X:

  X:\>cl \foo.cpp

Let's add a test for this, so all statements have been covered according
to Coverage.py.

### DIFF
--- a/unittests.py
+++ b/unittests.py
@@ -148,6 +148,12 @@ class TestSplitCommandsFile(BaseTest):
         self._genericTest('-A\r\n-B', ['-A', '-B'])
         self._genericTest('-A -B\r\n-C -D -E', ['-A', '-B', '-C', '-D', '-E'])
 
+    def testInitialBackslash(self):
+        self._genericTest(r'/Fo"C:\out dir\"', [r'/FoC:\out dir"'])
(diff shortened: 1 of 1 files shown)


frerich:
I used [Coverage.py](https://coverage.readthedocs.io/en/coverage-4.1/). It was a simple as running the three commands

    pip install coverage
    coverage run unittests.py
    coverage html

After that, a pretty HTML report is stored in `htmlcov/index.html`.

Alas, Coverage.py only generates statement coverage (I work for a company who sells a code coverage tool, so I have somewhat strong opinions on this topic), but it's very easy to get up and running and way better than nothing.

I'd love to have AppVeyor generate coverage information automatically, and found [a blog article](http://www.endzone.co.uk/opencover-nunit-test-coverage-on-appveyor/) on this topic. Ideally, a separate test case might be nice which verifies that no merge reduces the code coverage. :-)

PR #138, opened 2016-06-29
Alternative (more Pythonic?) approach to parsing command lines

frerich:
After expanding command files, a somewhat generic
CommandLineAnalyzer.analyze method was used to - well - analyse a
command line. What it did was to try to determine the source file(s)
mentioned in the command line and the output file. In case anything went
wrong and the command line was deemed unsupported, an 'AnalysisResult'
code was returned.

This patch adjusts this such that the method no longer returns a
3-tuple. Instead, it returns a list of source files and an output file.
In case the command line is deemed inappropriate, an exception is
raised.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -523,11 +523,28 @@ class CacheStatistics(object):
         self._stats.save()
 
 
-class AnalysisResult(object):
-    Ok, NoSourceFile, \
-        MultipleSourceFilesComplex, CalledForLink, \
-        CalledWithPch, ExternalDebugInfo, \
-        CalledForPreprocessing = list(range(7))
+class NoSourceFileError(Exception):
+    pass
+
+
+class MultipleSourceFilesComplexError(Exception):
+    pass
+
+
+class CalledForLinkError(Exception):
+    pass
+
+
+class CalledWithPchError(Exception):
+    pass
+
+
+class ExternalDebugInfoError(Exception):
+    pass
+
+
+class CalledForPreprocessingError(Exception):
+    pass
 
 
 def getCompilerHash(compilerBinary):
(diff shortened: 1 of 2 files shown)


clcache.py
```diff
@@ -825,23 +839,23 @@ def analyze(cmdline):
...
 
         if 'link' in options or ('c' not in options and not preprocessing):
-            return AnalysisResult.CalledForLink, None, None
+            raise CalledForLinkError()
 
         if len(sourceFiles) > 1:
             if compl:
-                return AnalysisResult.MultipleSourceFilesComplex, None, None
-            return AnalysisResult.MultipleSourceFilesSimple, sourceFiles, None
+                raise MultpleSourceFilesComplexError()
```
webmaster128: There is an "i" missing. Let's see if pylint find's this ...

webmaster128:
The main idea looks good.

The control flow around the `finally` part looks very confusing because

> A finally clause is always executed before leaving the try statement, whether an exception has occurred or not.

https://docs.python.org/2/tutorial/errors.html#defining-clean-up-actions

So it might only work here because there are return statements.

But it can handled much more elegantly when we have a common parent class `AnalysisError`:

```python
try:
   # ...
except AnalysisError as e:
    if isinstance(e, NoSourceFileError):
        printTraceStatement("Cannot cache invocation as {}: no source file found".format(cmdLine))
        updateCacheStatistics(cache, CacheStatistics.registerCallWithoutSourceFile)
    elif isinstance(e, MultipleSourceFilesComplexError):
        printTraceStatement("Cannot cache invocation as {}: multiple source files found".format(cmdLine))
        updateCacheStatistics(cache, CacheStatistics.registerCallWithMultipleSourceFiles)
    # all the other cases ...
    
    invokeRealCompiler(compiler, args[1:])
```

frerich:
The `finally` is a leftover actually. It's wrong, because it's even invoked if `return` is executed. That line can be removed completely, so you get:

```python
try:
  ...
  return ...
except NoSourceFileError:
  ..
except CalledForLinkingError:
  ..

invokeRealCompiler(..)
```

So the selected exception handler falls through.

webmaster128:
Do you develop on Unix or Windows? I was already thinking of implementing an easy lint shortcut for developers (e.g. `make lint`), which saves some CI roundtrips.

frerich:
I work on OS X and Windows. On OS X, I use Syntastic with pylint enabled, i.e. I see such errors right away in my editor. I didn't configure that in this Windows 10 VM though, so sometimes I just let the CI sort it out.

frerich:
I guess this doesn't look too bad, but the utter lack of tests is not exactly impressive. I suppose a good route to get this code in would be to

1. Add tests (with good coverage!) for `processCompileRequest`.
2. Adjust code to always return a list of source files.
3. Merge `Ok` and `MultipleSourceFilesSimple` analysis results.
4. Replace `AnalysisResult` status code with exceptions such that `analyze` always only returns a 2-tuple.
5. Rename `CommandLineAnalyzer.analyze` to something like `CommandLineAnalyzer.parseInputOutputFiles`.

webmaster128:
1. There are already a bunch of tests testing this via clcache command line call. Maybe not perfect, but I would not call this a show stopper
2. This is what #134 does in a minimal way. Let me know if I should work on that.
3. Can easily be implemented in #134 or later
4. When 3., is it really worth the change? The idea of having classes came from different workload types, which now (list, string) anyway
5. Renaming yes, but is it really "parse"? Isn't "parse" the string to list part? `determineInputOutputFiles`? `extractInputOutputFiles`?

I would like to do a 2.b: Add an error case for CalledForPreprocessing. There is no need for storing preprocessor output, it makes code much easier (e.g. no /Fi handling, concrete names like `outputFile` -> `objectFile`) and we don't need to handle the `/E file1.c file2.c` bug.

webmaster128:
:+1:

frerich:
PR #140 implements the removal of the support for caching preprocessor invocations.

frerich:
This seems okay to me now, except maybe for the fact that many of the preprocessor-specific tests are now redundant (they all just verify that some invocation is correctly flagged as unsupported).

PR #139, opened 2016-06-29
Add Makefile for local lint

webmaster128:
This allows you to run lint locally in a very convenient way.

Linux/OS X:
* `make lint`
* `make lint -j 3` (parallel linting in 3 jobs)

Windows:
* `nmake lint`
* `jom lint -j 3` (parallel linting in 3 jobs)

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -192,7 +192,7 @@ class ObjectCache(object):
         for o in objects:
             try:
                 objectInfos.append((os.stat(o), o))
-            except WindowsError:
+            except OSError:
                 pass
 
         objectInfos.sort(key=lambda t: t[0].st_atime)
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+.PHONY: lint-clcache lint-unittests lint-integrationtests lint
+
+lint-clcache:
+	pylint --rcfile .pylintrc clcache.py
+
+lint-unittests:
+	pylint --rcfile .pylintrc unittests.py
+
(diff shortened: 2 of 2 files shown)


clcache.py
```diff
@@ -192,7 +192,7 @@ def clean(self, stats, maximumSize):
         for o in objects:
             try:
                 objectInfos.append((os.stat(o), o))
-            except WindowsError:
+            except OSError:
```
frerich: Why this part?

webmaster128: Because on Ubuntu I always get `E:195,19: Undefined variable 'WindowsError' (undefined-variable)` and a non-zero return value.

Since this is the only place where pylint reacts differently on both platforms and `WindowsError` is a `OSError`, it is just convenient to replace it.

webmaster128: `OSError` is the only documented exception type for os.stat, so whatever the concrete type is, is implementation detail.

> Note: All functions in this module raise OSError in the case of invalid or inaccessible file names and paths, or other arguments that have the correct type, but are not accepted by the operating system.

https://docs.python.org/2/library/os.html
https://docs.python.org/2/library/os.html#os.stat

frerich: Interesting, thanks for digging that up! I'm surprised Ubuntu didn't complain about the

```python
```

at the top of `clcache.py`!

frerich:
Looks good, merged - thanks! Guess this is something one could invoke from the `appveyor.yml` file for code reuse.

PR #141, opened 2016-07-01
Publish coverage reports for tests at codecov.io

frerich:
An initial attempt at generating coverage reports during CI runs.
Ideally, I'd like to have a single HTML coverage report telling me which
parts of clcache.py were hit (and not hit) when executing both unit- as
well as integration tests.

For now, we limit ourselves to:

1. Only collecting coverage data when running unit tests
2. Generating HTML, but offering the artifact as a .zip file

Let's see how it goes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,6 +36,12 @@ install:
   - pip install pylint
   - pylint --version
 
+  # Install coverage plugin for pytest
+  - pip install pytest-cov
+
+  # Install tool for uploading coverage data to codecov.io
+  - pip install codecov
+
 build_script:
   - python clcache.py --help
   - python clcache.py -s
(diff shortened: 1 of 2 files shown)


webmaster128:
This is how you can integrate Coverage results in a fancy web dashboard:

https://github.com/codecov/example-python

I never did this with Python but this is how codecov.io looks for the C++ project Botan: https://codecov.io/github/randombit/botan

frerich:
Ah, interesting. I've been playing with coveralls.io in the meanwhile.

webmaster128:
cool

frerich:
The coverage results are visible at https://coveralls.io/github/frerich/clcache -- I think what's missing is that the integration tests contribute to the coverage, and I suspect there's not much point in calculating the coverage four times (for each Python version). Every AppVeyor build counts as four coveralls builds.

webmaster128:
Can this badge somehow go into the same line as the build status? https://github.com/frerich/clcache/tree/coverage_reports

frerich:
I guess it could. My AsciiDoc-fu is not strong enough yet, but I'm sure there is a way.

What's more important to me right now is that we only submit the coverage data to coveralls.io just once, e.g. by only taking the coverage data generated for the Python 2.7 build. For that, I'd somehow need to tell the individual matrix rows apart. And the other thing is that I'd like the integration tests to get included as well, but I think I saw a --cov-append switch for that somewhere.

webmaster128:
What I would do is

```
    matrix:
     # AppVeyor installed Python versions
     # http://www.appveyor.com/docs/installed-software#python
 
     - PYTHON_VERSION: "27"
     - PYTHON_VERSION: "33"
     - PYTHON_VERSION: "34"
     - PYTHON_VERSION: "35"
```

and replace `%PYTHON_INSTALL%` with `C:\Python%PYTHON_VERSION%`.

Then you just check for the PYTHON_VERSION to be whatever value you want to upload.

frerich:
Hm, not really happy with the presentation of coveralls.io -- I'll give https://codecov.io/ a try next.

appveyor.yml
```diff
@@ -73,9 +76,7 @@ after_test:
   - python clcache.py -s
 
   # Upload coverage data to codecov.io for analysis
-  - bash --version
-  - curl --version
-  - curl -s https://codecov.io/bash | bash -
+  - codecov
```
frerich: Turns out that bash and curl are indeed available - but using the Python uploader which was suggested by the codecov.io founder (who actually listens to the web chat facility!) seems kinda nice, so let's use that.

PR #142, opened 2016-07-01
Refactor analysis

webmaster128:
This is just one minor step towards /Zi.

On the way I found a bug when handling optional filepath parameters, e.g. in [/Yc](https://msdn.microsoft.com/en-us/library/7zc28563.aspx).

What I have in here is everything I could finish before dealing with the bug.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -844,12 +844,12 @@ class CommandLineAnalyzer(object):
             sourceFiles += options['Tc']
             compl = True
 
+        if len(sourceFiles) == 0:
+            raise NoSourceFileError()
+
         for opt in ['E', 'EP', 'P']:
             if opt in options:
                 raise CalledForPreprocessingError()
-
-        if len(sourceFiles) == 0:
-            raise NoSourceFileError()
 
         # Technically, it would be possible to support /Zi: we'd just need to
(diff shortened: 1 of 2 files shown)


frerich:
I like the simplification of `_parseOptionsAndFiles` - but I really miss the _Why_ in the "Move /E, /EP, /P checks below source files check" commit log. Even after staring at the code and the adjusted test a bit, I still don't get it.

webmaster128:
I documented it in the tests that lead to the change

> No source file is the worst thing that can happen. In this case there is no chance we can help, so it has priority over other errors.

This is mainly a consistency thing

frerich:
Yes, I saw the comment but I didn't understand the 'no chance we can help' part. I didn't se how we could help with 'CalledForPreprocessing'.

Your 'This is mainly a consistency thing' part is interesting though. I think I tend to agree that any command line without a source file should be logged as such, logging 'called for preprocessing' is a bit strange if we don't even have a source file to preprocess

webmaster128:
I found another reason why I think NoSourceFile has the highest priority: This case is likely to be a invalid command line in contrast to command lines that are just passed to the real compiler because we do not want to deal with it. So counting those as NoSourceFile in the stats might help pointing to the configuration issue.

PR #144, opened 2016-07-01
Optional argument parameter

webmaster128:
In MSVS, arguments use the formats `/NAME[ ]value` (optional space), `/NAME value` (required space), `/NAME[value]` (no space, optional value), `/NAMEvalue` (no space, required value).

Right now we always try `/NAMEvalue` and if there if no value, try  `/NAME value`. This strategy is too simple and fails for `/doc`, which must not consume the following argument when no value is set.

I guess we need to group the arguments in those four categories and handle them accordingly.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -402,6 +402,7 @@ class Configuration(object):
 
 class CacheStatistics(object):
     RESETTABLE_KEYS = {
+        "CallsWithInvalidArgument",
         "CallsWithoutSourceFile",
         "CallsWithMultipleSourceFiles",
         "CallsWithPch",
@@ -426,6 +427,12 @@ class CacheStatistics(object):
         for k in CacheStatistics.RESETTABLE_KEYS | CacheStatistics.NON_RESETTABLE_KEYS:
             if k not in self._stats:
                 self._stats[k] = 0
+
+    def numCallsWithInvalidArgument(self):
+        return self._stats["CallsWithInvalidArgument"]
+
+    def registerCallWithInvalidArgument(self):
+        self._stats["CallsWithInvalidArgument"] += 1
 
     def numCallsWithoutSourceFile(self):
         return self._stats["CallsWithoutSourceFile"]
@@ -1067,6 +1074,7 @@ clcache statistics:
     header changed             : {}
     source changed             : {}
   passed to real compiler
+    called w/ invalid argument : {}
     called for preprocessing   : {}
     called for linking         : {}
     called for external debug  : {}
@@ -1082,6 +1090,7 @@ clcache statistics:
         stats.numEvictedMisses(),
         stats.numHeaderChangedMisses(),
         stats.numSourceChangedMisses(),
+        stats.numCallsWithInvalidArgument(),
         stats.numCallsForPreprocessing(),
         stats.numCallsForLinking(),
         stats.numCallsForExternalDebugInfo(),
@@ -1344,6 +1353,9 @@ def processCompileRequest(cache, compiler, args):
                 return processNoDirect(cache, outputFile, compiler, cmdLine)
             else:
                 return processDirect(cache, outputFile, compiler, cmdLine, sourceFiles[0])
+    except InvalidArgumentError:
+        printTraceStatement("Cannot cache invocation as {}: invalid argument".format(cmdLine))
+        updateCacheStatistics(cache, CacheStatistics.registerCallWithInvalidArgument)
     except NoSourceFileError:
         printTraceStatement("Cannot cache invocation as {}: no source file found".format(cmdLine))
         updateCacheStatistics(cache, CacheStatistics.registerCallWithoutSourceFile)
--- a/clcache.py
+++ b/clcache.py
@@ -552,6 +552,10 @@ class CalledForPreprocessingError(AnalysisError):
     pass
 
 
+class InvalidArgumentError(AnalysisError):
+    pass
+
+
 def getCompilerHash(compilerBinary):
     stat = os.stat(compilerBinary)
     data = '|'.join([
@@ -772,6 +776,9 @@ class Argument(object):
     def __len__(self):
         return len(self.name)
 
+    def __str__(self):
+        return "/" + self.name
+
     def __eq__(self, other):
         return type(self) == type(other) and self.name == other.name
 
@@ -835,11 +842,23 @@ class CommandLineAnalyzer(object):
                 for arg in argumentsWithParameterSorted:
                     if cmdLineArgument.startswith(arg.name, 1):
                         isParametrized = True
-                        if len(cmdLineArgument) > len(arg) + 1:
+                        if isinstance(arg, ArgumentT1):
                             value = cmdLineArgument[len(arg) + 1:]
-                        else:
+                            if not value:
+                                raise InvalidArgumentError("Parameter for {} must not be empty".format(arg))
+                        elif isinstance(arg, ArgumentT2):
+                            value = cmdLineArgument[len(arg) + 1:]
+                        elif isinstance(arg, ArgumentT3):
+                            value = cmdLineArgument[len(arg) + 1:]
+                            if not value:
+                                value = cmdline[i + 1]
+                                i += 1
+                        elif isinstance(arg, ArgumentT4):
                             value = cmdline[i + 1]
                             i += 1
+                        else:
+                            raise AssertionError("Unsupported argument type.")
+
                         arguments[arg.name].append(value)
                         break
 
--- a/clcache.py
+++ b/clcache.py
@@ -810,7 +810,7 @@ class ArgumentT4(Argument):
 class CommandLineAnalyzer(object):
 
     @staticmethod
-    def parseArgumentsAndInputFiles(cmdline):
+    def _getParameterizedArgumentType(cmdLineArgument):
         argumentsWithParameter = {
             # /NAMEparameter
             ArgumentT1('Ob'), ArgumentT1('Yl'), ArgumentT1('Zm'),
@@ -830,6 +830,13 @@ class CommandLineAnalyzer(object):
         }
         # Sort by length to handle prefixes
         argumentsWithParameterSorted = sorted(argumentsWithParameter, key=len, reverse=True)
+        for arg in argumentsWithParameterSorted:
+            if cmdLineArgument.startswith(arg.name, 1):
+                return arg
+        return None
+
+    @staticmethod
+    def parseArgumentsAndInputFiles(cmdline):
         arguments = defaultdict(list)
         inputFiles = []
         i = 0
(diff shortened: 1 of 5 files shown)


clcache.py
```diff
@@ -793,13 +831,17 @@ def _outputFileFromArgument(options, argumentName, sourceFile, extension):
...
+            ArgumentT2('F'), ArgumentT2('FU'),
+            ArgumentT3('doc'), ArgumentT3('FA'), ArgumentT3('FR'), ArgumentT3('Fr'),
+            ArgumentT3('Gs'), ArgumentT3('MP'), ArgumentT3('Wv'), ArgumentT3('Yc'),
+            ArgumentT3('Yu'), ArgumentT3('Zp'),
+            ArgumentT4('AI'), ArgumentT4('D'), ArgumentT4('Fa'), ArgumentT4('Fd'),
+            ArgumentT4('Fe'), ArgumentT4('Fi'), ArgumentT4('Fm'), ArgumentT4('Fo'),
+            ArgumentT4('Fp'), ArgumentT4('Ob'), ArgumentT4('Tc'), ArgumentT4('Tp'),
+            ArgumentT4('V'), ArgumentT4('w1'), ArgumentT4('w2'), ArgumentT4('w3'),
+            ArgumentT4('w4'), ArgumentT4('wd'), ArgumentT4('we'), ArgumentT4('wo'),
+            ArgumentT4('Yl'), ArgumentT4('Zm'),
```
frerich: Are you sure this classification is correct? E.g. `/w3 1234` is not equivalent to `/w31234`?

webmaster128:
The current code is getting stable. However, I think we have a crash-bug in this since an exception is not handled. So I'd like to test this crash before fixing it.

Regarding classification: I checked all arguments in the MSVS 2015 documentation, where there is a nice monospaced notation for every type, e.g.

* Type 1 `/NAME[ ]parameter`: [/I in MSVS 2010](https://msdn.microsoft.com/en-us/library/73f9s62w(v=vs.100).aspx)
* Type 2 `/NAME parameter`: [/F in MSVS 2010](https://msdn.microsoft.com/en-us/library/tdkhxaks(v=vs.100).aspx)
* Type 3 `/NAME[parameter]`: [/doc in MSVS 2010](https://msdn.microsoft.com/en-us/library/ms173501(v=vs.100).aspx)
* Type 4 `/NAMEparameter`: [/Yl in MSVS 2010](https://msdn.microsoft.com/en-us/library/w6y1zk9f(v=vs.100).aspx)

First I added the tests as documented and then I changed the implementation to match the test.

There might however be cases that work in an undocumented way and ware used in an undocumented way. Those cases can be changed in the code on a case-by-case bugreport, and the diff to the documentation should be commented.

frerich:
Sounds good. For what it's worth, I'm currently experimenting with various GitHub services (such as 'QuantifiedCode', some kind of linter). We'll see whether we can get any benefit out of those, I suspect some tuning of warnings will be needed.

webmaster128:
In contradiction to the documentation, the MSVS 2015 implementation of cl.exe takes an optinal space after `/D`, `AI` and `/w1` `/w2` `/w3` `/w4`.

```
cl.exe /nologo /c /D minimal.cpp
cl.exe /nologo /c /AI minimal.cpp
cl.exe /nologo /c /w1 minimal.cpp
cl.exe /nologo /c /w2 minimal.cpp
cl.exe /nologo /c /w3 minimal.cpp
cl.exe /nologo /c /w4 minimal.cpp
```

all complain about missing source file and

```
cl.exe /nologo /c /D OUTminimal.cpp
cl.exe /nologo /c /AI dir minimal.cpp
cl.exe /nologo /c /w1 5999 minimal.cpp
cl.exe /nologo /c /w2 5999 minimal.cpp
cl.exe /nologo /c /w3 5999 minimal.cpp
cl.exe /nologo /c /w4 5999 minimal.cpp
```

work. Who wins now? Documentation or implementation?

frerich:
Maybe neither the documentation nor the implementation is at fault, but the premise that the `[ ]` notation in the MSDN documentation means what you think it does is wrong. It might indicate a preference or something?

In any case, the implementation always wins - the goal is to be a drop-in replacement, which means we should be bug-compatible as well.

I wonder whether there aren't similar parsers (maybe in Clang?) which give some precedence. Maybe we don't need to be as strict as possible but just as strict as necessary to resolve ambiguities.

webmaster128:
As far as I can see, the documentation matches the implementation in most cases.

I think the best thing we can do here is classify all `/NAMEparameter` arguments as `/NAME[ ]parameter` when the implementaion supports optional spaces. This would cover all examples from above.

Not all arguments accept an additional space. `/Fo` for example strictly requires the parameter to follow without space.

frerich:
I think it would be great if there was some way to automatically test whether the interpretation of `clcache` matches the parsing of whatever compiler is being used. Something which automatically tries all kinds of permutations for command line arguments (with and without space etc.) and checks whether the real `cl` sees it the same way.

webmaster128:
I am testing the arguments one by one and regroup everything based on my findings. I am very happy with the current state of testArgumentParameters() and the implementation. testParseArgumentsAndInputFiles() is going to be adjusted soon.

One thing I found was that `/Wv` is documented but not implemented up to including MSVS 2015 update 2. I am installing update 3 now and see if something changes. Maybe I did not find the expected input values, but everything I tried with `/Wv` resulted in an argument parsing error from cl.exe.

webmaster128:
Okay, `/Wv` is not a real thing (yet). It is included in all English documentation, but not in any other language.

```
https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/en-ca/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/en-gb/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/en-nz/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/en-za/library/thxezb7y(v=vs.140).aspx

https://msdn.microsoft.com/de-de/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/fr-fr/library/thxezb7y(v=vs.140).aspx
https://msdn.microsoft.com/es-es/library/thxezb7y(v=vs.140).aspx
```

Including MSVS 2015 update 3, it is not implemented.

So I am testing it as if it worked as documented.

webmaster128:
Okay made it – I would consider this to be done now.

clcache.py
```diff
@@ -402,6 +402,7 @@ def save(self):
 
 class CacheStatistics(object):
     RESETTABLE_KEYS = {
+        "CallsWidthInvalidArgument",
```
frerich: Typo here. Width should be With.

clcache.py
```diff
@@ -765,43 +776,110 @@ def expandCommandLine(cmdline):
...
+        return self.name == other.name
+
+    def __hash__(self):
+        return hash(self.name)
+
+
+# /NAMEparameter (no space, required parameter).
+class ArgumentT1(Argument):
+    def __init__(self, name):
+        super(ArgumentT1, self).__init__(name)
```
frerich: I think you don't need an explicit constructor here (and in the other subclasses), you can just write `pass` which will get you the `Argument` constructor.

clcache.py
webmaster128: Thanks, fixed

clcache.py
webmaster128: Good idea, thanks.

I wanted to use those later on to implement `__eq__`  and `__hash__` on `Argument`, such that `ArgumentT1('W') != ArgumentT3('W')` but then did not do it.

I now use type() in `Argument` to get the behavior, such that I can use the `pass` as suggested.

webmaster128:
Thanks for feedback. Implemented it.

clcache.py
```diff
@@ -765,43 +776,107 @@ def expandCommandLine(cmdline):
...
+            ArgumentT2('Zp'), ArgumentT2('Fa'), ArgumentT2('Fd'), ArgumentT2('Fe'),
+            ArgumentT2('Fi'), ArgumentT2('Fm'), ArgumentT2('Fo'), ArgumentT2('Fp'),
+            ArgumentT2('Wv'),
+            # /NAME[ ]parameter
+            ArgumentT3('AI'), ArgumentT3('D'), ArgumentT3('Tc'), ArgumentT3('Tp'),
+            ArgumentT3('FI'), ArgumentT3('U'), ArgumentT3('I'), ArgumentT3('F'),
+            ArgumentT3('FU'), ArgumentT3('w1'), ArgumentT3('w2'), ArgumentT3('w3'),
+            ArgumentT3('w4'), ArgumentT3('wd'), ArgumentT3('we'), ArgumentT3('wo'),
+            ArgumentT3('V'),
+            # /NAME parameter
```
frerich: Interesting - `ArgumentT4` is only defined for symmetry, but it's not currently used. That's intentional?

webmaster128: That is true. It was created because `/F` and `/FU` are documented as `/NAME param`. But then I found out those are implemented as `/NAME[ ]param`. That was the point where I moved it to number 4 from somewhere in the middle.

I kind of would like to keep it as reference for now, but would also be okay with removing it.

clcache.py
```diff
@@ -864,7 +864,11 @@ def parseArgumentsAndInputFiles(cmdline):
                     else:
                         raise AssertionError("Unsupported argument type.")
 
-                    arguments[arg.name].append(value)
+                    # consistently use empty list for flags like /c or /doc
+                    if not arg.name in arguments:
+                        arguments[arg.name] = []
+                    if value:
+                        arguments[arg.name].append(value)
```
frerich: I don't understand the purpose of this commit - this `if` just manually replicates the logic provided by `defaultdict`, no? What does this patch improve?

clcache.py
```diff
@@ -850,7 +925,7 @@ def analyze(cmdline):
             raise MultipleSourceFilesComplexError()
 
         if len(inputFiles) == 1:
-            if 'Fo' in options:
+            if 'Fo' in options and len(options['Fo']) > 0:
```
frerich: [PEP 8](https://www.python.org/dev/peps/pep-0008/) explains:

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
> 
> Yes: if not seq:
>     if seq:
>
> No: if len(seq):
>    if not len(seq):

Hence, I'd say `if 'Fo' in options and options['Fo']:` would be more idiomatic.

clcache.py
webmaster128: It changes the output of empty-value arguments like `/doc` from `{'doc': ['']}` to `{'doc': []}`

clcache.py

clcache.py
webmaster128: With this change, the appending is only done when `value` is non-empty. But in both cases (empty/non-empty `value`), we need to set the dictionary key.

webmaster128: But I just see that this is the wrong fix, since it converts `/X /Xab` to `{X: ['ab']}` instead of `{X: ['', 'ab']}`, so we're losing information.

The correct fix is to go for

```python
                    argumentName = cmdLineArgument[1:] # name not followed by parameter in this case
                    arguments[argumentName].append('')
```

below. Then switches and arguments without parameter value always get the '' value.

webmaster128:
Okay, I reworked the consistency fox for `/c` and `/doc`. Both result in `{'NAME': ''}` now. This makes the defaultdict-change obsolete. The `Fo` part was adapted as well.

frerich:
Quite a big change after all, but the tests look sensible so I guess it should be alright. Merged, thanks! 👍

Issue #143, opened 2016-07-01
What is the /vm argument?


webmaster128:
In [optionsWithParameter](https://github.com/frerich/clcache/blob/f4748f6e9659308aece0365737468913dbfd8bca/clcache.py#L791), there is a /vm argument listed, which I cannot find in the MSVS documentation.

Instead there are a bunch of switches with that prefix: /vmb /vmg /vmm /vms /vmv, which are not parameters of /vm but independent arguments.

Can /vm go?

webmaster128:
Same goes for /W

frerich:
`/W` takes an argument, see e.g. the [Visual Studio 2008 documentation](https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.90).aspx). Both `/W4` as well as `/W 4` are legal, it seems.

frerich:
As for `/vm`: I think it's wrong to consider it an option with an argument. `/vmm` is an option in itself, `/vm m` is not the same thing.

webmaster128:
Thanks for pointing out the older MSVS version. /W takes an argument until including MSVS 2013 and was [changed with MSVS 2015](https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.140).aspx).

By the way, which one is the oldest MSVS version we are willing to support? I think there should be some official note on that in the README. That might be a question for the community (who ever that is). We're happy with MSVS 2015.

webmaster128:
When we combine the signature from MSVS 2008-2013 (`/Wn` with `n = 0, ..., 4`) and the signature from MSVS 2015 (`/W0`, `/W1`, `/W2`, `/W3`, `/W4`) we can treat it just as `/NAMEparameter` in both cases and thus keep it in the list.

PR #145, opened 2016-07-02
Fix optionsWithParameter list

webmaster128:
This is a small set of patches to make #144 smaller. It fixes issue #143.

This is ready to merge as it is.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -795,6 +795,7 @@ class CommandLineAnalyzer(object):
             'D', 'U', 'I', 'Zp',
             'MP', 'Tc', 'V', 'wd', 'wo',
             'W', 'Yc', 'Yl', 'Tp', 'we',
+            'w1', 'w2', 'w3', 'w4', 'Wv',
             'Yu', 'Zm', 'F', 'Fi',
         }
         # Sort by length to handle prefixes
--- a/clcache.py
+++ b/clcache.py
@@ -788,13 +788,15 @@ class CommandLineAnalyzer(object):
 
     @staticmethod
     def _parseOptionsAndFiles(cmdline):
-        optionsWithParameter = {'Ob', 'Gs', 'Fa', 'Fd', 'Fm',
-                                'Fp', 'FR', 'doc', 'FA', 'Fe',
-                                'Fo', 'Fr', 'AI', 'FI', 'FU',
-                                'D', 'U', 'I', 'Zp', 'vm',
-                                'MP', 'Tc', 'V', 'wd', 'wo',
-                                'W', 'Yc', 'Yl', 'Tp', 'we',
-                                'Yu', 'Zm', 'F', 'Fi'}
+        optionsWithParameter = {
+            'Ob', 'Gs', 'Fa', 'Fd', 'Fm',
+            'Fp', 'FR', 'doc', 'FA', 'Fe',
+            'Fo', 'Fr', 'AI', 'FI', 'FU',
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -792,9 +792,9 @@ def _parseOptionsAndFiles(cmdline):
             'Ob', 'Gs', 'Fa', 'Fd', 'Fm',
             'Fp', 'FR', 'doc', 'FA', 'Fe',
             'Fo', 'Fr', 'AI', 'FI', 'FU',
-            'D', 'U', 'I', 'Zp', 'vm',
+            'D', 'U', 'I', 'Zp',
             'MP', 'Tc', 'V', 'wd', 'wo',
-            'W', 'Yc', 'Yl', 'Tp', 'we',
+            'Yc', 'Yl', 'Tp', 'we',
```
frerich: I don't think this is correct for Visual Studio 2013. The [documentation](https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.120).aspx) explains that `/W` takes a numeric argument, e.g. both `/W4` as well as `/W 4` are valid.

clcache.py
```diff
@@ -793,9 +793,9 @@ def _parseOptionsAndFiles(cmdline):
             'Fp', 'FR', 'doc', 'FA', 'Fe',
             'Fo', 'Fr', 'AI', 'FI', 'FU',
             'D', 'U', 'I', 'Zp',
-            'MP', 'Tc', 'V', 'wd', 'wo',
-            'Yc', 'Yl', 'Tp', 'we',
-            'Yu', 'Zm', 'F', 'Fi',
+            'F', 'Fi', 'MP', 'Tc', 'Tp', 'V',
+            'w1', 'w2', 'w3', 'w4', 'wd', 'we', 'wo', 'Wv',
```
frerich: A test for this would be nice which verifies that e.g. `['/w3', '1234']` is parsed correctly.

frerich:
Looks promising - I think the fix for `/wn` which is apparently new in Visual Studio 2015 deserves a change log entry.

clcache.py
webmaster128: Is `/W 4` really valid? I don't see any source for that. All documentation is talking about `/W4`. Is the space version used in the wild?

webmaster128:
I push the test into a separate issue because I would like to do an interface change in CommandLineAnalyzer, that allows to do those tests. Right now, there is no getter to access those intermediate data and it is thrown away because unused in analyze().

But the implementation will handle it correctly, since when `/w3` prefix is found it looks for more parameter data in the argument (`/w3xxxxxxx`) and when there is none, it takes the next argument. So `['/w3', '1234']` as well as `['/w31234']` leads to `{'/w3': ['1234']}`.

`/wd`, `/we` and `/wo` are unchanged and just sorted alphabetically. So the commit log includes all added elements.

Since the parsed options are thrown away in analyze(), I don't think there is a real point for the changelog. The only thing that happens is that `cl /c /w1 123 main.cpp` will not read 123 and main.cpp as two different source files anymore. But I think `/w1 123` is wrong anyway and must be `/w1123` in both [MSVS 2008-2013](https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.120).aspx) (see /wln) and [MSVS 2015](https://msdn.microsoft.com/en-us/library/thxezb7y(v=vs.140).aspx).

webmaster128:
I updated the PR. /W stays in and the `/w1 /w2 /w3 /w4 /Wv` commit is now smaller.

PR #147, opened 2016-07-03
Remove unused function CommandLineAnalyzer._preprocessToStdout

webmaster128:
This is obsolete now since preprocessing calls are forwarded to the
compiler.


frerich:
Merged, thanks!

PR #150, opened 2016-07-05
Soothe quantifiedcode

frerich:


clcache.py
```diff
@@ -1334,7 +1334,8 @@ def processDirect(cache, outputFile, compiler, cmdLine, sourceFile):
             origCmdLine = cmdLine
             stripIncludes = False
             if '/showIncludes' not in cmdLine:
-                cmdLine = ['/showIncludes'] + cmdLine
+                cmdLine = ['/showIncludes']
+                cmdLine.extend(origCmdLine)
                 stripIncludes = True
```
webmaster128: Would `cmdLine = ['/showIncludes'] + origCmdLine` have the same effect?

webmaster128:
> Quantifiedcode (maybe incorrectly so) complained that there were methods
in a TestCase which were neither setUp() nor tearDown(), nor did they
start with "test_"

I think this is because we do camelCase. When `test*` is the rule for tests names, `test_` is the expected prefix when you do sname_case.

See also the Anti Pattern here: https://www.quantifiedcode.com/app/issue_class/5vXxuyoh

clcache.py
frerich: *Probably* yes, I didn't think of that. In this case, the suggested solution (either calling `cmdLine.extend` or `cmdLine += `) is actually harmful because it modifies the list argument (which is passed by reference). I.e. the program

```python
def f(xs):
  xs.extend([3,4])

l = [1,2]
f(l)
print len(l)
```

prints '4'. So it was actually intentional to create a new list here, but I tend to agree with QC that it seemed suspicious.

frerich:
Interestingly, simplifying the code by inheriting from `unittest.TestCase` directly made QC complain *more*. It now complains about each and every method in the test cases. Even internal helper functions (`_testFi` and such) cause this warning. It's not clear to me yet whether QC is just being pedantic and should be silenced, or whether we indeed miss some conventions for the `unittest` module (I think I read something about `test discovery` being a thing in this context, but did not yet research more).

clcache.py
webmaster128: It complains that

> The list `cmdLine` is extended using reassignment. 

which is not optimal, because we copy everything from `cmdLine` into a new list. While the general advise might make sense, it should not be an issue in our case.

The suggested workarounds are both for extending at the end of the list. But we need to add somthing at the beginning, which we can do by `cmdLine.insert(0, '/showIncludes')`.

Since I don't think it is very readable, I came up with `cmdLine = ['/showIncludes'] + origCmdLine`, which is combining A and B to C, not extending something.

clcache.py
```diff
@@ -1334,7 +1334,7 @@ def processDirect(cache, outputFile, compiler, cmdLine, sourceFile):
             origCmdLine = cmdLine
```
webmaster128: Btw. is that a copy of the list or a copy of the reference?

frerich:
I disabled the QC warnings about the 'test' prefix for unit test methods for now. A lot of noise.

clcache.py
frerich: It's a reference, i.e. in

```python
a = [1,2]
b = a
b.append(3)
```

both `len(a)` and `len(b)` will be 3

webmaster128: But I guess we want a deep copy here, right?

frerich: I don't think so: we don't modify either list. We do however make `cmdLine` reference a *new* list, but doing so will not change the list referenced by `origCmdLine`. Or am I missing something? It's getting late...

webmaster128: But isn't the entire purpose of  `origCmdLine` to be able to modify `cmdLine`? It's like a Backup. So it should relyably work no matter *how* we change `cmdLine` later on.

If we went for `cmdLine.insert(0, '/showIncludes')`, the `origCmdLine` would be altered just because we did not do a proper copy.

frerich: At the time the `origCmdLine = cmdLine` alias is done, we don't even know yet whether we will need to create a new command line, i.e. whether we need a deep copy at all. And indeed, if we had used `insert` then that would modify both lists. You could also have done

```python
cmdLine = list(origCmdLine)
cmdLine.insert(0, '/showIncludes')
```

and I hope QC would accept that just as well.

webmaster128: Right, so the question remains: is `origCmdLine` a backup or an alias. I think the first one is more obviours and we should do `origCmdLine = cmdLine.copy()` no matter how we deal with the place below.

btw. copy() is not a deep copy (as I thought before), but a shallow copy, so it only copies a bunch of pointers to the (immutable) string elements.

frerich: In Python2, lists have no `copy` method. The `list` constructor can be used to copy lists.

webmaster128: hmm, right, good point. Maybe using the [copy module](https://docs.python.org/2/library/copy.html) leads to nicer code (`origCmdLine = copy.copy(cmdLine)` because it does not put focus on the type but on the copying.

frerich: I think if you insist on the term 'copy', you could also just define `def copyList(xs): return list(xs)` but at this point I think it's safe to assume that anybody who's somewhat familiar with Python knows that if you do `list(xs)` you create a new list object. ;-)

frerich:
Merged this, it does move things into the right direction. The issue with how we construct a new command line can be considered separately.

PR #151, opened 2016-07-05
Minor refactorings

webmaster128:
This is a set of commits extracted from #144 to make the PR smaller.


frerich:
Looks good, merged. Thanks!

PR #153, opened 2016-07-07
Fix/piped output

gqmelo:
After [MBCS decoding/encoding](https://github.com/frerich/clcache/commit/5514b7b30f14a11f4ed60d6777d812fad55a7406) clcache started to output duplicated CR chars.
I think this is due to Windows C API which automatically converts LF to CRLF
More info on https://github.com/ninja-build/ninja/issues/773 and https://connect.microsoft.com/VisualStudio/feedback/details/523174/c-library-fprintf-and-fwrite-convert-line-endings-incorrectly

### DIFF
--- a/integrationtests.py
+++ b/integrationtests.py
@@ -151,6 +151,29 @@ class TestCompileRuns(BaseTest):
         subprocess.check_call(cmd) # Compile once
         subprocess.check_call(cmd) # Compile again
 
+    def testPipedOutput(self):
+        commands = [
+            # passed to real compiler
+            [PYTHON_BINARY, CLCACHE_SCRIPT, '/?'],
+            [PYTHON_BINARY, CLCACHE_SCRIPT, '/E', 'fibonacci.c'],
+            # Unique parameters ensure this was not cached yet (at least in CI)
+            [PYTHON_BINARY, CLCACHE_SCRIPT, '/wd4267', '/wo4018', '/c', 'fibonacci.c'],
+            # Cache hit
+            [PYTHON_BINARY, CLCACHE_SCRIPT, '/wd4267', '/wo4018', '/c', 'fibonacci.c'],
+        ]
+
+        with cd(ASSETS_DIR):
+            for cmd in commands:
+                try:
+                    output = subprocess.check_output(cmd).decode(clcache.CL_DEFAULT_CODEC)
+                except subprocess.CalledProcessError as e:
+                    self.fail('{}\nOutput:\n{}'.format(e, e.output))
+                self.assertTrue('\r\r\n' not in output,
+                                'Command:{}\nOutput has duplicated CR: {}'.format(cmd, output))
+                # Just to be sure we have newlines
+                self.assertTrue('\r\n' in output,
+                                'Command:{}\nOutput has no CRLF: {}'.format(cmd, output))
+
 
 class TestCompilerEncoding(BaseTest):
     def testNonAsciiMessage(self):
--- a/clcache.py
+++ b/clcache.py
@@ -86,6 +86,14 @@ Manifest = namedtuple('Manifest', ['includeFiles', 'hashes'])
 
 
 def printBinary(rawData, stream):
+    # See http://stackoverflow.com/questions/2374427/python-2-x-write-binary-output-to-stdout
+    if sys.version_info[0] < 3:
+        # msvcrt and O_BINARY are only available on Windows, but we like to
+        # use pylint on Unix developer machines to speed-up development, thus
+        # we disable some lint error on a per-line base here.
+        import msvcrt # pylint: disable=import-error
+        msvcrt.setmode(stream.fileno(), os.O_BINARY) # pylint: disable=no-member
+
     with os.fdopen(stream.fileno(), 'wb') as fp:
         fp.write(rawData)
 


webmaster128:
Interesting, I'll have a look at this. I'm sure this can be done and the test case looks good to me.

However, using `universal_newlines` is not the way to go, since it decodes the process output automatically into unicode strings, which we need to do manually in order to have control over the encoding.

gqmelo:
@webmaster128, as far as I know ```universal_newlines``` doesn't decode to unicode.
It merely do this:
```python
    def _translate_newlines(self, data):
        data = data.replace("\r\n", "\n")
        data = data.replace("\r", "\n")
        return data
```
Also if you add this to my test it passes:
```python
self.assertTrue(isinstance(output, str), output.__class__)
```

webmaster128:
It does in Python 3.

gqmelo:
Hmm, OK, I did not know that.
So maybe just doing the replaces manually would do?

webmaster128:
I did not yet understand the problem, but I guess, yes. We can apply useful filters on the `stdout` and `stderr` variables. I think both variables are native str() types, i.e. binary string in Python 2 and unicode string in Python 3. So you can do all the usual string manipulation there.

gqmelo:
The problem is that if you write "\r\n" to a file or to a PIPE on Windows, it will be converted to "\r\r\n".
That is a windows peculiarity: https://connect.microsoft.com/VisualStudio/feedback/details/523174/c-library-fprintf-and-fwrite-convert-line-endings-incorrectly
I imagine that is because on C/C++ the standard is LF only, regardless of OS. So they had to convert LF to CRLF automatically on their C calls (which python uses)

Having an additional CR mess the output up, as a CR alone is normally printed as a newline. To be more specific we use [ninja](https://ninja-build.org/) to build and it captures the compiler output to be processed/printed later.
After updating clcache, ninja started to print lots of newlines.

webmaster128:
I wonder if it is really that code that causes the problem, because if stdout contains a "\r\n" from the compiler, it is printing it via `printBinary(compilerStdout.encode(CL_DEFAULT_CODEC), stream=sys.stdout)` in binary mode. So the "\r\n" should be preserved.

Might `returnCode = subprocess.call(realCmdline)` be the issue?

gqmelo:
> Might returnCode = subprocess.call(realCmdline) be the issue?
I don't think so, because the problem was happening even when I got a cache hit

gqmelo:
So, that is weird, it is exactly ```printBinary``` that is doing that. If I just change:

```python
def printBinary(rawData, stream):
    with os.fdopen(stream.fileno(), 'wb') as fp:
        fp.write(rawData.replace(b"\r\n", b"\n"))
```
the test passes

webmaster128:
We indeed changed the line endings in 5514b7b30f14a11f in such a way that we used to store normalized `\n` line endings and now store the original unchanged content from the compiler `\r\n`. I think this is plausible to do.

But the output part gets interesting. When I run this little script in the command line, line endings look good:

```python
string = "1\r\n2\r\n3\r\n"

print("String printing:")
print(string)

print("Binary printing:")
raw = string.encode("mbcs")
print(raw)
```

But when I pipe it into a file via `python binary_printer.py > out.txt`, I get the `\r\r\n` line endings.

So the extra `\r\n` -> `\r\n` is from piping?

-------------------------

Next observation is that `python binary_printer.py > out.txt` with

```python
from clcache import printBinary
import sys

string = "1\n2\n3\n"

raw = string.encode("mbcs")
printBinary(raw, stream=sys.stdout)
```

results in single `\n` line endings

but 

```python
from clcache import printBinary
import sys

string = "1\r\n2\r\n3\r\n"

raw = string.encode("mbcs")
printBinary(raw, stream=sys.stdout)
```

results in `\r\r\n`.

So this cannot be an `\n` -> `\r\n` conversion but a `\r\n` -> `\r\r\n` conversion... I have no clue how that can happen.

gqmelo:
> We indeed changed the line endings in 5514b7b in such a way that we used to store normalized \n line endings and now store the original unchanged content from the compiler \r\n. I think this is plausible to do.

I agree, keeping the output unchanged should be reasonable.

I also made some tests and created a bat with this:

```batch
:: As expected (CRLF)
python -c "f = open('crlf_binary', 'wb'); f.write('Print\r\nbinary\r\n')"
:: As expected (CRCRLF)
python -c "f = open('crlf_nonbinary', 'w'); f.write('Print\r\nnonbinary\r\n')"
:: As expected (no extra newlines on output)
python -c "import os, sys; f = os.fdopen(sys.stdout.fileno(), 'wb'); f.write('Print\r\nbinary\r\n')"
:: As expected? (no extra newlines on output)
python -c "import os, sys; f = os.fdopen(sys.stdout.fileno(), 'w'); f.write('Print\r\nnonbinary\r\n')"
:: Weird (CRCRLF)
python -c "import os, sys; f = os.fdopen(sys.stdout.fileno(), 'wb'); f.write('Print\r\nbinary\r\n')" > crlf_stdout_binary
:: As expected (CRCRLF)
python -c "import os, sys; f = os.fdopen(sys.stdout.fileno(), 'w'); f.write('Print\r\nnonbinary\r\n')" > crlf_stdout_nonbinary
:: As expected (CRLF)
python -c "import os, sys, msvcrt; msvcrt.setmode(sys.stdout.fileno(), os.O_BINARY); f = os.fdopen(sys.stdout.fileno(), 'wb'); f.write('Print\r\nnonbinary\r\n')" > crlf_stdout_binary_msvcrt
```

So it seems that both stdout and pipes have an odd behaviour. When printing to stdout without piping it always write CRLF regardless of the mode. When piping it writes CRCRLF regardless of the mode.

So what do you think about using ```msvcrt.setmode(sys.stdout.fileno(), os.O_BINARY)``` on ```printBinary```?

gqmelo:
I think this is really the way to go:
http://stackoverflow.com/questions/2374427/python-2-x-write-binary-output-to-stdout
http://code.activestate.com/recipes/65443-sending-binary-data-to-stdout-under-windows/

Probably doing ```os.fdopen(sys.stdout.fileno(), 'wb')``` does not work because python already opened stdout and reuses the open file.

integrationtests.py
```diff
@@ -151,6 +151,16 @@ def testRecompileObjectSetOtherDir(self):
...
 
+    def testPipedOutput(self):
+        with cd(ASSETS_DIR):
+            cmd = [PYTHON_BINARY, CLCACHE_SCRIPT, "/nologo", "/c", "fibonacci.c", "/Fofibonacci_c.obj"]
+            try:
+                output = subprocess.check_output(cmd)
+            except subprocess.CalledProcessError as e:
+                self.fail('{}\nOutput:\n{}'.format(e, e.output))
+            self.assertNotEqual(output, b'')
+            self.assertTrue(b'\r\r\n' not in output, 'Output has duplicated CR: {}'.format(repr(output)))
```
webmaster128: Looks like there is no `in` operation on binary data in Python 3.5. We can get around that by converting to string. This does not change the `\r`/`\n` situation:

```
output = subprocess.check_output(cmd).decode(clcache.CL_DEFAULT_CODEC)
```

integrationtests.py
```diff
@@ -151,6 +151,16 @@ def testRecompileObjectSetOtherDir(self):
         subprocess.check_call(cmd) # Compile once
         subprocess.check_call(cmd) # Compile again
 
+    def testPipedOutput(self):
+        with cd(ASSETS_DIR):
+            cmd = [PYTHON_BINARY, CLCACHE_SCRIPT, "/nologo", "/c", "fibonacci.c", "/Fofibonacci_c.obj"]
```
webmaster128: Do we really know that there are some proper line endings from this call? As far as I know, there is only the file name.

What about adding dropping /nologo?

gqmelo:
> Interesting. Maybe this could be done using os.open as well?
os.open takes a filename

os.fdopen should be OK, but I think the problem is that stdout/stderr were already open by python. So it seems to be ignoring the 'b' flag

integrationtests.py
gqmelo: Actually there is an output, The test fails when I leave the old code.
But it is better to have more output. I will remove /nologo

integrationtests.py
gqmelo: Done

integrationtests.py
```diff
@@ -151,6 +151,16 @@ def testRecompileObjectSetOtherDir(self):
...
+            try:
+                output = subprocess.check_output(cmd).decode(clcache.CL_DEFAULT_CODEC)
+            except subprocess.CalledProcessError as e:
+                self.fail('{}\nOutput:\n{}'.format(e, e.output))
+            self.assertNotEqual(output, '')
+            self.assertTrue('\r\r\n' not in output, 'Output has duplicated CR: {}'.format(repr(output)))
```
webmaster128: Nice – the `repr` is obsolete now

integrationtests.py
gqmelo: I tried to remove /nologo, but then clcache fails because it tries to parse the output for includes

webmaster128: Could you the deleting the cache including the manifests directory? Who knows what is in there ...

webmaster128:
> os.open takes a filename

right, fine.

One other thing: Could you please add this not-so-fancy but handy decoration around the implementation?

```python
    # msvcrt and O_BINARY are only available on Windows, but we like to use pylint on Unix developer
    # machines to speed-up development, thus we disable some lint error on a per-line base here.
    import msvcrt # pylint: disable=import-error
    msvcrt.setmode(stream.fileno(), os.O_BINARY) # pylint: disable=no-member
```

webmaster128:
Regarding the test: I think we can simplify the call to `[PYTHON_BINARY, CLCACHE_SCRIPT, '/?']`. This gives us a bunch of native line breaks passed though clcache and ensures that no cache content is involved.

webmaster128:
Okay that was not true. We don't capture and re-print the output in case of `/?`.

So we really need different tests, e.g.

```python
[
    # passed to real compiler
    [PYTHON_BINARY, CLCACHE_SCRIPT, '/?'],
    [PYTHON_BINARY, CLCACHE_SCRIPT, '/E', 'fibonacci.c'],
    # Unique parameters ensure this was not cached yet (at least in CI)
    [PYTHON_BINARY, CLCACHE_SCRIPT, '/wd4267', '/wo4018', '/c', 'fibonacci.c'],
    # Cache hit
    [PYTHON_BINARY, CLCACHE_SCRIPT, '/wd4267', '/wo4018', '/c', 'fibonacci.c'],
]

gqmelo:
> One other thing: Could you please add this not-so-fancy but handy decoration around the implementation?

Sure. Done!

> So we really need different tests, e.g.

Not sure if I understood, but I added the four commands to the test

Also as this problem does not affect python3, I changed to use msvcrt conditionally

webmaster128:
The commands I mentioned take entirely different code path through clcache, which affect how STDOUT and STDERR output is handled.

The first one for example goes in `main()`, `processCompileRequest()`, handled by `NoSourceFileError`, calls `invokeRealCompiler` with `captureOutput=False`, calling `subprocess.call(realCmdline)`, which does the printing, returning `stdout = ''` and `sterr = ''`, so `printBinary` is called twice with empty input.

webmaster128:
From my side: great! You could squash both commits, I think.

gqmelo:
Thanks, I did not see there was two commits

frerich:
I tried to follow the debugging you guys performed (thanks a lot!) and would like to make sure that I got this right: due to a somewhat unintuitive behaviour of the CRT on Windows, running clcache using Python 2.x would cause a cache hit to reproduce the cached compiler output wrongly if the cached output ends in CRLF - in that case, CRCRLF was printed, which appears like an extra newline. Does that sound about right?

Sounds like a good catch! Two things which catch my eye:

1. Why does the code coverage fall (even if only by 0.13%)?
2. I think this bugfix deserves mentioning in the ChangeLog (in fact, maybe it justifies doing a new bugfix release!)

gqmelo:
> in that case, CRCRLF was printed, which appears like an extra newline. Does that sound about right?

Exactly, this is probably because of how python2 handles stdout and stderr (it ignores the 'b' in the 'wb' flags if you try to open them again)

> Why does the code coverage fall (even if only by 0.13%)?

I guess the coverage is done with python3. The fix is done only with python2 (I don't think we should risk messing up python3)

frerich:
Merged, thanks! I'll add a ChangeLog entry for this fix shortly.

webmaster128:
> Why does the code coverage fall (even if only by 0.13%)?

This is because this commit increases the codebase by three lines and `printBinary` is no covered by unittests. It is just called from integrationtests via a separate clcache process, which is not counted.

PR #154, opened 2016-07-07
Fix variable declaration non-conformant to C89

gqmelo:
Declaring variables on C89 need to be done at the beginning of the
function scope. The old code failed on VS2010.

### DIFF
--- a/tests/integrationtests/fibonacci.c
+++ b/tests/integrationtests/fibonacci.c
@@ -6,9 +6,8 @@ void print_fib(int n)
     int b = 1;
     while (a < n)
     {
-        printf("%d ", a);
-
         int old_a = a;
+        printf("%d ", a);
         a = b;
         b = old_a + b;
     }


webmaster128:
Confirmed using gcc:

```
$ gcc -pedantic -std=c99 fibonacci.c
// ok
$ gcc -pedantic -std=c99 fibonacci.c
fibonacci.c: In function ‘print_fib’:
fibonacci.c:11:9: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
         int old_a = a;
         ^
```

Looks like it just has to be at the beginning of any scope, so

```
    while (a < n)
    {
        int old_a = a;
        printf("%d ", a);
        a = b;
        b = old_a + b;
    }
```

should be okay too. This version makes the lifespan of `old_a` clearer.

gqmelo:
You are right, it is any scope. Confirmed with VS2010 too
I moved it inside the while

webmaster128:
Great :+1:

Looks like we have some MSVS 2010 guys here – interesting

gqmelo:
Unfortunately my company is stuck with 2010 until we migrate to python3 because we use a lot of compiled extensions.

frerich:
Looks good, merged. Thanks!

PR #156, opened 2016-07-09
Fix print stderr bug on preprocessor fail

webmaster128:
This is a follow-up on #153 because I think I found a place in the code, that does not do proper binary output to stdout/stderr:

```python
    @staticmethod
    def computeKey(compilerBinary, commandLine):
        ppcmd = [compilerBinary, "/EP"]
        ppcmd += [arg for arg in commandLine if arg not in ("-c", "/c")]
        preprocessor = Popen(ppcmd, stdout=PIPE, stderr=PIPE)
        (preprocessedSourceCode, pperr) = preprocessor.communicate()

        if preprocessor.returncode != 0:
            sys.stderr.write(pperr)
            sys.stderr.write("clcache: preprocessor failed\n")
            sys.exit(preprocessor.returncode)
```

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -250,11 +250,11 @@ class ObjectCache(object):
         ppcmd = [compilerBinary, "/EP"]
         ppcmd += [arg for arg in commandLine if arg not in ("-c", "/c")]
         preprocessor = Popen(ppcmd, stdout=PIPE, stderr=PIPE)
-        (preprocessedSourceCode, pperr) = preprocessor.communicate()
+        (preprocessedSourceCode, ppStderrBinary) = preprocessor.communicate()
 
         if preprocessor.returncode != 0:
-            sys.stderr.write(pperr)
-            sys.stderr.write("clcache: preprocessor failed\n")
+            printBinary(ppStderrBinary, stream=sys.stderr)
+            print("clcache: preprocessor failed", file=sys.stderr)
             sys.exit(preprocessor.returncode)
 
         compilerHash = getCompilerHash(compilerBinary)
--- a/clcache.py
+++ b/clcache.py
@@ -85,17 +85,26 @@ BASEDIR_REPLACEMENT = '?'
 Manifest = namedtuple('Manifest', ['includeFiles', 'hashes'])
 
 
-def printBinary(rawData, stream):
-    # See http://stackoverflow.com/questions/2374427/python-2-x-write-binary-output-to-stdout
-    if sys.version_info[0] < 3:
-        # msvcrt and O_BINARY are only available on Windows, but we like to
-        # use pylint on Unix developer machines to speed-up development, thus
-        # we disable some lint error on a per-line base here.
-        import msvcrt # pylint: disable=import-error
-        msvcrt.setmode(stream.fileno(), os.O_BINARY) # pylint: disable=no-member
+class BinaryPrinter(object):
+    didSetMsvcrtMode = defaultdict(lambda: False)
 
-    with os.fdopen(stream.fileno(), 'wb') as fp:
-        fp.write(rawData)
+    def __init__(self, stream):
+        self._stream = stream
+
+    def write(self, rawData):
+        # See http://stackoverflow.com/questions/2374427/python-2-x-write-binary-output-to-stdout
+        # Note: msvcrt.setmode must not be called twice on al given stream
+        if sys.version_info[0] < 3:
+            # msvcrt and O_BINARY are only available on Windows, but we like to
+            # use pylint on Unix developer machines to speed-up development, thus
+            # we disable some lint error on a per-line base here.
+            import msvcrt # pylint: disable=import-error
+            if not BinaryPrinter.didSetMsvcrtMode[self._stream.fileno()]:
+                msvcrt.setmode(self._stream.fileno(), os.O_BINARY) # pylint: disable=no-member
+                BinaryPrinter.didSetMsvcrtMode[self._stream.fileno()] = True
+
+        with os.fdopen(self._stream.fileno(), 'wb') as fp:
+            fp.write(rawData)
 
 
 def basenameWithoutExtension(path):
@@ -253,7 +262,7 @@ class ObjectCache(object):
         (preprocessedSourceCode, ppStderrBinary) = preprocessor.communicate()
 
         if preprocessor.returncode != 0:
-            printBinary(ppStderrBinary, stream=sys.stderr)
+            BinaryPrinter(sys.stderr).write(ppStderrBinary)
             print("clcache: preprocessor failed", file=sys.stderr)
             sys.exit(preprocessor.returncode)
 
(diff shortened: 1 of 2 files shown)


webmaster128:
Check, test triggered the suspected bug and I patched the implementation.

webmaster128:
@frerich Could you turn of these coverage checks as long as the integration tests are not counted? It looks somehow frustrating when a well written and properly tested patch (or even test driven developed like this one) is marked with a red error cross.

frerich:
I agree it's annoying. I now filed https://github.com/codecov/support/issues/254 to get some help with this, I just can't figure out why the integration tests don't count. I didn't quite grok how codecov.io merges builds.

webmaster128:
I don't thing that merging the results is the problem but how methods are called. In integrationtests, test process A is coverage checked, i.e. some kind of mechanism is injected into A to determine what is executed. Everything that happens in A is counted. Now A calls a separate clcache process B via command line. But the coverage mechanism is not hooked into B and A cannot see what happens internally in B.

If there is no super magic way to work around that, I think that is a limitation of every coverage mechanism in every language. The only thing that comes to my mind to solve this would be to run a clcache server that handles the command line calls all in one process. But that might end up in a huge amount of overhead.

frerich:
Maybe this is indeed caused by the integration tests running `clcache.py` in separate Python interpreter processes. Maybe it would be viable to either run them in the same interpreter as the one executing the processes, or the integration tests do not invoke `clcache.py` directly but rather via `codecov` (assuming that one can call `codecov myprogram.py` or the like).

webmaster128:
I like the later way. When we call clcache in the same process, we loose a lot of error potential and loose contact to the real world scenario. E.g. the \r\n issues only occur because we call clcache from command line.

frerich:
Submitted #157 to try getting more meaningful coverage data (which is also a better place to continue this discussion, I guess).

frerich:
#157 was just merged which includes the integration test coverage data into the overall coverage value. I assume that if this PR gets rebased on top of master, you should see an improvement in coverage, too.

webmaster128:
Good job – let's see

frerich:
Pity, looks like there's still something to be understood.

frerich:
Seeing lines [like this](https://ci.appveyor.com/project/frerich/clcache/build/1.0.515/job/pnp1c852avux3vhc#L279) in the output makes me wonder whether the issue is caused by the argument vector made of Unicode as well as non-Unicode strings (which might also explain why this issue only exists with Python 2.x).

webmaster128:
Right now, I don't get why there is not clcache tracing in [this CI run](https://ci.appveyor.com/project/frerich/clcache/build/1.0.519). The environment contains `'CLCACHE_LOG': '1'`.

frerich:
I think there is no tracing because `clcache.py` is not executed at all. Instead, the `coverage` program returns with exit code `-1073740777` which in (32bit) hexadecimal is c0000417. This looks like a typical NTSTATUS code in the Windows API. Indeed, checking the [reference documentation](https://msdn.microsoft.com/en-us/library/cc704588.aspx) explains that this code correspnds to

> STATUS_INVALID_CRUNTIME_PARAMETER - An invalid parameter was passed to a C runtime function.

Now it would be good to figure out why `coverage` returns this. And why this only happens for this single integration test.

frerich:
I just checked the [CI output](https://ci.appveyor.com/project/frerich/clcache/build/1.0.519/job/7da008mtlqpv8sjg) again and noticed that the first failing case is the first one to have the `directoMode` flag set to `False`. In that case, the environment is created slightly differently. The code does:

    testEnvironment = dict(os.environ, CLCACHE_NODIRECT=str("1"), CLCACHE_LOG=str("1"))

Given that this is a notable difference and that the issue apparently only occurs with Python 2, maybe it would be worth changing this `str` to `unicode` just for the sake of checking whether it makes a difference. A Unicode issue in passing the environment to the Windows API would fit the `STATUS_INVALID_CRUNTIME_PARAMETER` error code.

webmaster128:
Interesting. unicode leads to the issue we always get whenpassing unicode strings to env:

> E                                    TypeError: environment can only contain strings

> c:\python27\lib\subprocess.py:958: TypeError

But I think that clcache is run [in the test you linked](https://ci.appveyor.com/project/frerich/clcache/build/1.0.519/job/7da008mtlqpv8sjg) because we get a proper STDOUT and STDERR from compiler.

clcache.py
```diff
@@ -85,17 +85,26 @@
...
 
-def printBinary(rawData, stream):
-    # See http://stackoverflow.com/questions/2374427/python-2-x-write-binary-output-to-stdout
-    if sys.version_info[0] < 3:
-        # msvcrt and O_BINARY are only available on Windows, but we like to
-        # use pylint on Unix developer machines to speed-up development, thus
-        # we disable some lint error on a per-line base here.
-        import msvcrt # pylint: disable=import-error
-        msvcrt.setmode(stream.fileno(), os.O_BINARY) # pylint: disable=no-member
+class BinaryPrinter(object):
```
frerich: Interesting - do you think that this has something to do with the Python 2.7-specific (possibly Unicode-related) issues with launching that peculiar integration test?

webmaster128: This is definitely the fix. Oh man, I traced this down to death.

The particular point is

> Note: msvcrt.setmode must not be called twice on al given stream

frerich:
Wow, good job!

frerich:
Seeing how painful this was, I do consider dropping Python 2.x support in the near future. We'll see.

frerich:
Merged, at last! Thanks!

webmaster128:
Arg, damn. That was not the core issue.

Whatever happens to `std.stderr` after the first write causes a crash.

Calling this via `C:\Python27\python.exe dummyscript.py` crashes after *ok 3*. When I comment out `msvcrt.setmode(stream.fileno(), os.O_BINARY)`, it crashes after *ok 4*.

```python
from __future__ import print_function
import os, sys, msvcrt

stream = sys.stderr

print("ok 1")

msvcrt.setmode(stream.fileno(), os.O_BINARY)

print("ok 2")

with os.fdopen(stream.fileno(), 'wb') as fp:
    fp.write(b"1")

print("ok 3")
    
msvcrt.setmode(stream.fileno(), os.O_BINARY)

print("ok 4")

with os.fdopen(stream.fileno(), 'wb') as fp:
    fp.write(b"2")

print("ok 5")
```

webmaster128:
I am out of ideas and don't know how to continue with this. Dropping Python 2 support seems like a good idea, since solid IO is a core requirement for clcache.

The list of supported Windows systems should be roughly same as long as we support Python 3.4:

> Up to 2.5, Python was still compatible with Windows 95, 98 and ME (but already raised a deprecation warning on installation). For Python 2.6 (and all following releases), this support was dropped and new releases are just expected to work on the Windows NT family.

https://docs.python.org/2/using/windows.html

> As specified in PEP 11, a Python release only supports a Windows platform while Microsoft considers the platform under extended support. This means that Python 3.5 supports Windows Vista and newer. If you require Windows XP support then please install Python 3.4.

https://docs.python.org/3/using/windows.html#supported-versions

frerich:
Hm. I guess at this point, it's not an actual problem but rather a risk (AKA a problem waiting to manifest...), right? Or is it possible to write a test case which breaks things as they are?

I just considered writing to the `CON` pseudo-file but AFAICS that exhibits the same issue. Interestingly, it appears you cannot redirect such writes via `>`, they always go to the console.

frerich:
Here's another idea: since we only need to support Windows, we can also access the Windows API directly via the `ctypes` module and use the `WriteFile` function to write straight to stdout. This works for me:

```python
import ctypes
import ctypes.wintypes

STD_OUTPUT_HANDLE = ctypes.wintypes.DWORD(-11)

def write(text):
    hStdout = ctypes.windll.kernel32.GetStdHandle(STD_OUTPUT_HANDLE)
    n = ctypes.wintypes.DWORD(0)
    retval = ctypes.windll.kernel32.WriteFile(
        hStdout,
        ctypes.create_string_buffer(text),
        len(text),
        ctypes.byref(n),
        None
    )
```

Using this, `write('foo\n')` redirected to a file prints 4 bytes (as expected, unlike sys.stdout.write which writes 5 bytes). It should get tested with Unicode characters still, or very large inputs (it may be needed to call the function in multiple chunks), but maybe it works out. It's certainly less code than the new `BinaryWriter` class. :-}

webmaster128:
If that works, great. I think you can revert https://github.com/frerich/clcache/commit/903a4f152810ca9880b59365da8738e9a84ed770 in a PR and update the printBinary implementation from there. The test fix came (I guess) from disabling the CLCACHE_LOG in the tests.

frerich:
I'm more inclined to push #158 :-)

PR #157, opened 2016-07-11
Integrationtest coverage

frerich:


webmaster128:
Looks good to me. I think `clcache.py` should somehow be absolute in the Powershell, such that `CLCACHE_CMD` works even when we change directory first in tests.

frerich:
Yeah, looks like it. Let's see whether my Powershell-fu is strong enough to make this work.

PR #158, opened 2016-07-13
Drop python2 support

frerich:


frerich:
It this works out, it may be worth considering whether we want to support the scenario that the encoding used for cached compiler output differs from the encoding used when restoring cached compiler output. Right now, we do support this which implies some encoding/decoding is required for cache hits.

I tend to favour speed here, so maybe we should just store compiler output as binary data and also restore (and print) it as such, assuming that the encoding did not change? We also get binary data from the subprocess module, which we could easily forward.

clcache.py
```diff
@@ -33,10 +33,6 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
-from __future__ import print_function
-from __future__ import unicode_literals # All string literals are unicode strings. Requires Python 3.3+
-from __future__ import division
```
webmaster128: Could cou do those changes in the 2 test files as well?

webmaster128:
I would vote for the [printBinary implementation](https://github.com/frerich/clcache/blob/072f7bcef5d22329dbf81fdc567873c752095ace/clcache.py#L88) before #153. It works fine in Python 3, is simple to call and has a simple, adaptable implementation.

webmaster128:
> I tend to favour speed here, so maybe we should just store compiler output as binary data and also restore (and print) it as such, assuming that the encoding did not change?

Given that mbcs is system dependent, I would keep storing the STDOUT/STDERR as UTF-8. This allows us to share compiler caches, when shared locking is possible. This has a real speed impact. Additionaly, since we do not store preprocessout output, STDOUT/STDERR is very small amounts of data.

frerich:
> I would vote for the printBinary implementation before #153. It works fine in Python 3, is simple to call and has a simple, adaptable implementation.

Not sure I understand. To me, it seems easier to just use `sys.stdout.buffer.write(data)` instead of `printBinary(sys.stdout, data)`. The latter isn't really shorter (unless I'm miscounting, it's just as long) but it probably does make Python people wonder what it does in addition to just writing to `sys.stdout.buffer`. Or am I getting this all wrong?

frerich:
Hm, turns out that maybe there is a compelling reason to factor out the `sys.*.buffer.write` calls: it appears that pylint has some parse quirk which makes it complain that in e.g.

```python
sys.stdout.buffer.write(foo)
```

the `buffer` class (which is a builtin type) has no `write` member. It appears pylint doesn't notice that the token is not used as a type name here. Using

```python
getattr(sys.stdout, 'buffer').write(foo)
```

works around the issue, but is too convoluted. So factoring this out into `printBinary` and then disabling the `no-member` warning for that light might be a good idea.

webmaster128:
Hmm.. looks like `sys.stdout.buffer` is not the proper way to do it: https://ci.appveyor.com/project/frerich/clcache/build/1.0.526/job/8b8ek37cqq12e8dk But this is still documented all over the web.

I don't see a `buffer` property in stderr documented: https://docs.python.org/3/library/sys.html#sys.stdout and can't find a class documentation for `buffer`. So I don't know if it works by hacking private APIs or if this is clean.

frerich:
Other people seem to be surprised that pylint complains here, too, so I created https://github.com/PyCQA/pylint/issues/1004 to report this issue. Let's see.

frerich:
For what it's worth, the buffer of a stream does show up in the output of e.g. `help(sys.stdout)` (though it has no explanation assigned to it), and `help(sys.stdout.buffer)` gives the documentation of a `BufferedWriter` class which doesn't look like it's an internal thing.

webmaster128:
Cool, thanks. I did not know this help() thing. I always got stuck with the online documentation.

frerich:
Technically, the test isn't directly related to this particular PR (it's not related to the Python 2 vs. Python 3 functionality). However, it addresses some code which was rewritten, which made the coverage drop.

integrationtests.py
```diff
@@ -477,6 +473,21 @@ def testHitsSimple(self):
...
 
 
+class TestNoDirectCalls(unittest.TestCase):
+    def testPreprocessorFailure(self):
+        cache = clcache.ObjectCache()
+
+        oldStats = clcache.CacheStatistics(cache)
+
+        cmd = CLCACHE_CMD + ["/nologo", "/c", "doesnotexist.cpp"]
+        env = dict(os.environ, CLCACHE_NODIRECT=str("1"))
```
webmaster128: ther str() is obsolete now, since str() and "1" are both the native Python 3 string type

frerich: Ah, is that so? Excellent! I mostly based this on the existing code, ahem - this is now my first 'real' Python 3 project, I still need to become more familiar with all the goodies. :-)

PR #161, opened 2016-07-22
Remove cPickle import

webmaster128:
This is obsolete since we don't support Python 2 anymore.

"In Python 2, you can speed up your pickle access with cPickle. (In
Python3, importing pickle will automatically use the accelerated version
if it is available.)" https://wiki.python.org/moin/UsingPickle


frerich:
A no-brainer. Merged, thanks!

PR #164, opened 2016-07-27
Enable skipped TestCommandLineArguments.testValidMaxSize() test

frerich:
This test was disabled since it modifies the clcache configuration, but
executing the test should not modify the user's global configuration.

We can overcome this by specifying a custom CLCACHE_DIR path, and we
pass a temporary directory.

### DIFF
--- a/integrationtests.py
+++ b/integrationtests.py
@@ -14,6 +14,7 @@ import glob
 import os
 import subprocess
 import sys
+import tempfile
 import unittest
 
 import clcache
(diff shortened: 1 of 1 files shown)


integrationtests.py
```diff
@@ -40,12 +41,13 @@ def cd(targetDirectory):
...
     def testValidMaxSize(self):
-        validValues = ["1", "  10", "42  ", "22222222"]
-        for value in validValues:
-            cmd = CLCACHE_CMD + ["-M", value]
-            self.assertEqual(subprocess.call(cmd), 0, "Command must not fail for max size: '" + value + "'")
+        with tempfile.TemporaryDirectory() as tempDir:
+            validValues = ["1", "  10", "42  ", "22222222"]
+            for value in validValues:
+                cmd = CLCACHE_CMD + ["-M", value]
+                p = subprocess.Popen(cmd, env=dict(os.environ, CLCACHE_DIR=tempDir))
```
webmaster128: Why plain Popen? All subprocess functions can be used with `env` as well

From https://docs.python.org/3/library/subprocess.html#subprocess.call

> The arguments shown above are merely the most common ones. The full function signature is largely the same as that of the Popen constructor - this function passes all supplied arguments other than timeout directly through to that interface.

frerich: Good point - totally forgot about that! I based the code on existing tests in this file, which also used `Popen` for no good reason it seems. PR #165 addresses this.

PR #167, opened 2016-07-28
Rework Manifests

webmaster128:
This brings some basic clarity in Manifests handling.

* ManifestsManager now deals with manifests
* Renamings make it easier to understand what all those hashes are for
* The use of JSON makes manifests human readable, easier to debug and better compatible between Python versions (pickle used different formats in Python 2 and Python 3)
* Manifests are now cleaned and cleared

This does invalidate manifest keys but not objects keys. So when nodirect-mode is used, nothing changes. The old .dat manifests are cleaned away automatically over time.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -78,19 +78,19 @@ class ManifestsManager(object):
     def __init__(self, manifestsRootDir):
         self._manifestsRootDir = manifestsRootDir
 
-    def _manifestDir(self, manifestHash):
+    def manifestDir(self, manifestHash):
         return os.path.join(self._manifestsRootDir, manifestHash[:2])
 
-    def _manifestName(self, manifestHash):
-        return os.path.join(self._manifestDir(manifestHash), manifestHash + ".dat")
+    def manifestPath(self, manifestHash):
+        return os.path.join(self.manifestDir(manifestHash), manifestHash + ".dat")
 
     def setManifest(self, manifestHash, manifest):
-        ensureDirectoryExists(self._manifestDir(manifestHash))
-        with open(self._manifestName(manifestHash), 'wb') as outFile:
+        ensureDirectoryExists(self.manifestDir(manifestHash))
+        with open(self.manifestPath(manifestHash), 'wb') as outFile:
             pickle.dump(manifest, outFile)
 
     def getManifest(self, manifestHash):
-        fileName = self._manifestName(manifestHash)
+        fileName = self.manifestPath(manifestHash)
         if not os.path.exists(fileName):
             return None
         try:
--- a/clcache.py
+++ b/clcache.py
@@ -14,7 +14,6 @@ import errno
 import hashlib
 import json
 import os
-import pickle
 from shutil import copyfile, rmtree
 import subprocess
 from subprocess import Popen, PIPE
@@ -82,6 +81,8 @@ class LogicException(Exception):
 
 
 class ManifestsManager(object):
+    HASH_INVALIDATION_COUNTER = 2
+
     def __init__(self, manifestsRootDir):
         self._manifestsRootDir = manifestsRootDir
 
@@ -89,23 +90,24 @@ class ManifestsManager(object):
         return os.path.join(self._manifestsRootDir, manifestHash[:2])
 
     def manifestPath(self, manifestHash):
-        return os.path.join(self.manifestDir(manifestHash), manifestHash + ".dat")
+        return os.path.join(self.manifestDir(manifestHash), manifestHash + ".json")
 
     def setManifest(self, manifestHash, manifest):
         ensureDirectoryExists(self.manifestDir(manifestHash))
-        with open(self.manifestPath(manifestHash), 'wb') as outFile:
-            pickle.dump(manifest, outFile)
+        with open(self.manifestPath(manifestHash), 'w') as outFile:
+            # Convert namedtupel to dict to get key names
+            doc = {'includeFiles': manifest.includeFiles, 'hashes': manifest.hashes}
+            json.dump(doc, outFile, sort_keys=True, indent=2)
 
     def getManifest(self, manifestHash):
         fileName = self.manifestPath(manifestHash)
         if not os.path.exists(fileName):
             return None
         try:
-            with open(fileName, 'rb') as inFile:
-                return pickle.load(inFile)
-        except (IOError, pickle.UnpicklingError):
-            # - file does not exist or cannot be opened (IOError)?
-            # - file is corrupted (pickle.UnpicklingError)
+            with open(fileName, 'r') as inFile:
+                doc = json.load(inFile)
+                return Manifest(doc['includeFiles'], doc['hashes'])
+        except IOError:
             return None
 
     def clean(self, maxManifestsSize):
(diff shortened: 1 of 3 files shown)


clcache.py
```diff
@@ -74,6 +80,80 @@ def __str__(self):
...
+
+    def setManifest(self, manifestHash, manifest):
+        ensureDirectoryExists(self.manifestDir(manifestHash))
+        with open(self.manifestPath(manifestHash), 'w') as outFile:
+            # Convert namedtupel to dict to get key names
+            doc = {
+                'includeFiles': manifest.includeFiles,
+                'includesContentToObjectMap': manifest.includesContentToObjectMap
+            }
+            json.dump(doc, outFile, sort_keys=True, indent=2)
```
frerich: There is a [`collections.namedtuple._asdict`](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict) method to convert a `namedtuple` to a dictionary.

A [StackOverflow answer](http://stackoverflow.com/a/26180604/91757) explains:

> This is a documented method of namedtuples, i.e. unlike the usual convention in python the leading underscore on the method name isn't there to discourage use. Along with the other methods added to namedtuples, _make, _replace, _source, _fields, it has the underscore only to try and prevent conflicts with possible field names.

clcache.py
```diff
@@ -74,6 +80,80 @@ def __str__(self):
...
+                doc = json.load(inFile)
+                return Manifest(doc['includeFiles'], doc['includesContentToObjectMap'])
+        except IOError:
+            return None
+
+    def clean(self, maxManifestsSize):
+        manifestFiles = []
+        for path, _, filenames in WALK(self._manifestsRootDir):
+            for filename in filenames:
+                manifestFiles.append(os.path.join(path, filename))
```
frerich: This (potentially long) list of manifest files is not needed, you can define a generator via e.g.

```python
def _manifestFiles(self):
    for path, _, filenames in WALK(self._manifestsRootDir):
        for filename in filenames:
            yield os.path.join(path, filename)
```

clcache.py
webmaster128: Great, thanks. I did not find anything like that. I'll have a look at it.

clcache.py
webmaster128: Yeah, right. I used the code from objects cleaning, where we should use the generator too.

webmaster128: Ah, but below we need the entire list at one, right?

`manifestFileInfos.sort(key=lambda t: t[0].st_atime, reverse=True)`

clcache.py
```diff
@@ -74,6 +80,80 @@ def __str__(self):
...
+
+        manifestFileInfos = []
+        for filepath in manifestFiles:
+            try:
+                manifestFileInfos.append((os.stat(filepath), filepath))
+            except OSError:
+                pass
```
frerich: I wonder whether there's an opportunity for [`contextlib.suppress`](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) here...

clcache.py
```diff
@@ -74,6 +80,80 @@ def __str__(self):
...
+
+    @staticmethod
+    def getManifestHash(compilerBinary, commandLine, sourceFile):
+        compilerHash = getCompilerHash(compilerBinary)
+
+        # NOTE: We intentionally do not normalize command line to include
+        # preprocessor options. In direct mode we do not perform
+        # preprocessing before cache lookup, so all parameters are important
+        additionalData = "{}|{}|{}".format(
+            compilerHash, commandLine, ManifestsManager.HASH_INVALIDATION_COUNTER)
```
frerich: Even after seeing how `HASH_INVALIDATION_COUNTER` is used, it's still not clear to me what it's good for. Is it supposed to be some sort of versioning mechanism?

clcache.py
```diff
@@ -74,6 +80,80 @@ def __str__(self):
...
+        return getFileHash(sourceFile, additionalData)
+
+    @staticmethod
+    def getIncludesContentHash(listOfHeaderHashes):
+        hasher = HashAlgorithm()
+        hasher.update(','.join(listOfHeaderHashes).encode("UTF-8"))
+        return hasher.hexdigest()
```
frerich: This can be shortened to

```python
return HashAlgorithm(','.join(listOfHeaderHashes).encode()).hexdigest()
```

frerich:
Looks good to me overall. I think this makes a few further code deduplication opportunities apparent.

clcache.py
frerich: Note that that line sorts `manifestFileInfos`, not `manifestFiles`.

frerich: If the same code is used for cleaning, too, then maybe this would make a nice general purpose generator like:

```python
def filesBeneath(path):
    for path, _, filenames in WALK(path):
        for filename in filenames:
            yield os.path.join(path, filename)
```

...because then you can write

```python
manifestFiles = filesBeneath(self._manifestsRootDir)
```

which looks kinda nice.

clcache.py
webmaster128: Does this help? The line becomes a bit long (87 columns)

```
        return HashAlgorithm(','.join(listOfHeaderHashes).encode('UTF-8')).hexdigest()
```

I do not have a strong opinion for any version here

frerich: Part of the shortening was based on the insight that the [`str.encode`](https://docs.python.org/3/library/stdtypes.html#str.encode) function provides 'utf-8' as the default value for the encoding.

clcache.py
webmaster128: contextlib.suppress is Python 3.4+

clcache.py
```diff
@@ -74,6 +86,78 @@ def __str__(self):
         return repr(self.message)
 
 
+class ManifestsManager(object):
+    # Bump this counter whenever the current manifest file format changes.
+    # E.g. changing the file format from {'oldkey': ...} to {'newkey': ...} requires
+    # invalidation, such that a manifest that was stored using the old format is not
+    # interpreted using the new format. Instead the old file will not be touched
+    # again due to a new manifest hash and is cleaned away after some time.
+    HASH_INVALIDATION_COUNTER = 3
```
frerich: A counter which is changed whenever some format changes is often just called a 'version number'. So how about this gets renamed to `MANIFEST_FORMAT_VERSION`?

webmaster128:
Reopened-to trigger AppVeyor. Github's status API is a bit unreliable these days. Happens for Travis builds too in other projects.

clcache.py
webmaster128: Very nice. Namedtuple -> _asdict() -> OrderedDict -> JSON preserves the key names as well as the order.

clcache.py
webmaster128: Beautiful, implemented.

clcache.py

webmaster128:
Okay, I am happy and think I got all comments now.

frerich:
Looks good to me, thanks! I think this certainly moves things into the right direction. I think the fact that the manifests have been changed in a backwards-incompatible way, but that they are now cleaned up automatically is certainly worth an entry to the `CHANGELOG.md` file.

webmaster128:
okay, fine. I am writing a changelog

PR #168, opened 2016-07-28
Pyinstaller

frerich:


.gitignore
```diff
@@ -1,5 +1,6 @@
 *.pyc
 *.pyo
+clcache.spec
```
webmaster128: Could you add a commented block where the file comes from? E.g.

```
...
/build
/dist

# pyinstaller
clcache.spec

PR #169, opened 2016-07-28
Mention manifests

webmaster128:


frerich:
Oops, sorry. I only saw that you were going to write the changelog entry after I did it... luckily you did it anyway, because your version is actually nicer than mine. So I took that now, thanks! :-)

PR #170, opened 2016-07-28
Document alternative installation via clcache.bat

webmaster128:
This is the way we use clcache all the time. It becomes especially interesting, when a system's default Python is Python 2.


frerich:
Thanks, I actually used clcache exactly like that (via a .bat file) in the very beginning.

PR #171, opened 2016-07-28
Remove dead code

webmaster128:
when preprocessing is in place, processDirect() is not called =>
postprocessNoManifestMiss() is not called => postprocessNoManifestMiss()
is only called without '/E', '/EP', '/P' => grabStderr is always false.

Credits go to the code coverage visualization tool.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1229,19 +1229,9 @@ def postprocessHeaderChangedMiss(cache, outputFile, manifest, manifestHash, incl
     return compilerResult
 
 
-def postprocessNoManifestMiss(
-        cache, outputFile, manifestHash, baseDir, cmdLine, sourceFile, compilerResult, stripIncludes):
+def postprocessNoManifestMiss(cache, outputFile, manifestHash, baseDir, sourceFile, compilerResult, stripIncludes):
     returnCode, compilerOutput, compilerStderr = compilerResult
-    grabStderr = False
-    # If these options present, cl.exe will list includes on stderr, not stdout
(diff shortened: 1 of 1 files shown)


webmaster128:
Hmm, I don't understand why I don't get coverage feedback for this one.

frerich:
It seems that [codecov.io saw your commit](https://codecov.io/github/frerich/clcache/commit/a8252b02827776b4960f071a7781d8a557a77162) but is waiting or something, or processing something? Maybe the service is a bit slow...

frerich:
Good catch, by the way! I just read the code myself and indeed, `CommandLineAnalyzer.analyze` will throw if any of `/E`, `/EP` or `/P` is in the given command line, in which case `processDirect` is not called. So `processDirect` and hence `postprocessNoManifestMiss` is always only called with command lines which contain none of those three options, so `grabStderr` is always `False`.

I just checked whether there's a good unreachable code detection tool for Python which can do this, but my first find [vulture](https://github.com/myint/vulture)  doesn't find this. :-/

frerich:
Re-triggered a build on AppVeyor now (you can do so with a single click on the AppVeyor site, no need to close/open a PR).

webmaster128:
> Re-triggered a build on AppVeyor now (you can do so with a single click on the AppVeyor site, no need to close/open a PR).

Only you can do that as the repo owner.

frerich:
Merged this, thanks! More coverage by removing code, and the tests still pass. Good stuff.

PR #172, opened 2016-07-29
Feature prepare environment script

Un1oR:
Setup for Visual Studio usage looks several complicated and not convenient.
I wrote script which modify installed Visual Studio toolsets by prepending ExecutablePath with directory contains clacache executable. Also script sets system environment variables CLCACHE_DIR and CL.
I'm not sure my approach too perfect, but ideas can by useful.

### DIFF
--- a/env.py
+++ b/env.py
@@ -36,19 +36,25 @@ def generate_props_content(clcache_dir):
         </Project>''').format(clcache_dir=clcache_dir)[1:]
 
 
-def set_system_variable(var, value):
+def check_call_quiet(*args, **kwargs):
     with open(os.devnull, 'w') as devnull:
-        if value:
-            subprocess.check_call(['setx', '-m', var, value], stderr=devnull, stdout=devnull)
+        kwargs['stderr'] = devnull
+        kwargs['stdout'] = devnull
+        subprocess.check_call(*args, **kwargs)
+
+
+def set_system_variable(var, value):
+    if value:
+        check_call_quiet(['setx', '-m', var, value])
+    else:
+        reg_path = r'HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
+        try:
+            check_call_quiet(['reg', 'query', reg_path, '/V', var])
+        except subprocess.CalledProcessError:
+            # Variable doesn't exists
+            pass
         else:
-            reg_path = r'HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
-            try:
-                subprocess.check_call(['reg', 'query', reg_path, '/V', var], stderr=devnull, stdout=devnull)
-            except subprocess.CalledProcessError:
-                # Variable doesn't exists
-                pass
-            else:
-                subprocess.check_call(['reg', 'delete', reg_path, '/F', '/V', var], stderr=devnull, stdout=devnull)
+            check_call_quiet(['reg', 'delete', reg_path, '/F', '/V', var])
     
         
 def install(exe, cache_dir):
--- a/env.py
+++ b/env.py
@@ -1,0 +1,98 @@
+from __future__ import print_function
+
+import argparse
+import glob
+import os
+import sys
+import textwrap
+import subprocess
+
+
+def ensure_dir(path):
+    if os.path.exists(path):
+        return
+    
+    os.mkdir(path)
+    return path
+
+
+def traverse_toolsets(f):
+    msbuild_dir = r'C:\Program Files (x86)\MSBuild'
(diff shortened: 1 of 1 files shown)


env.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import print_function
```
frerich: The rest of the clcache script code now requires Python 3.3 or newer. I think imposing the same requirement here would be consistent, and it would simplify things abit (e.g. this import is not needed).

env.py
```diff
@@ -0,0 +1,111 @@
...
+def install(exe, cache_dir):
+    def f(toolset):
+        clcache_props = clcache_props_path(toolset)
+        ensure_dir(os.path.dirname(clcache_props))
+        with open(clcache_props, 'w') as f:
+            f.write(generate_props_content(os.path.dirname(exe)))
+    
+    traverse_toolsets(f)
+    set_system_variable('CLCACHE_DIR', cache_dir)
+    set_system_variable('CL', '/MP4')
```
frerich: I don't think this script should change the `CL` variable behind the scenes.

env.py
```diff
@@ -0,0 +1,111 @@
...
+    traverse_toolsets(f)
+    set_system_variable('CLCACHE_DIR', None)
+    set_system_variable('CL', None)
+
+
+def main(args=sys.argv[1:]):
+    clcache_default_exe = os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        'dist',
+        'cl.exe')
```
frerich: Does the executable have to be called `cl.exe` such that Visual Studio accepts it? Otherwise, I'd use `clcache.exe` here since that's the file name you get when you follow the documented way to build an executable, namely by running

    pyinstaller --onefile clcache.exe

env.py
```diff
@@ -0,0 +1,111 @@
...
+        kwargs['stderr'] = devnull
+        kwargs['stdout'] = devnull
+        subprocess.check_call(*args, **kwargs)
+
+
+def set_system_variable(var, value):
+    if value:
+        check_call_quiet(['setx', '-m', var, value])
+    else:
+        reg_path = r'HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment'
```
frerich: Would it also be sufficient to affect just the current user, i.e. use `HKCU` instead of `HKLM`?

env.py
```diff
@@ -0,0 +1,111 @@
...
+    return os.path.join(toolset, 'ImportAfter', 'Clcache.props')
+
+
+def generate_props_content(clcache_dir):
+    return textwrap.dedent('''
+        <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+          <PropertyGroup>
+            <ExecutablePath>{clcache_dir};$(ExecutablePath)</ExecutablePath>
+          </PropertyGroup>
+        </Project>''').format(clcache_dir=clcache_dir)[1:]
```
frerich: Is there some documentation on how these `.props` files work? I never saw this before.

Un1oR: I just found in MSBuild\Microsoft.Cpp\v4.0\V140\Platforms\x64\PlatformToolsets\v140\Toolset.props this lines:
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="$(MSBuildThisFileDirectory)ImportBefore\*.props" Condition="Exists('$(MSBuildThisFileDirectory)ImportBefore')" />
  ...
  <Import Project="$(MSBuildThisFileDirectory)ImportAfter\*.props" Condition="Exists('$(MSBuildThisFileDirectory)ImportAfter')" />
</Project>
```

Un1oR: First my idea was to create own toolset like clang-cl. I read existing toolset files and found this extension point.

env.py
Un1oR: Yes, this is doubtful place.
The thing is that in out projects /Fo always point to directory, not to file. And without setting CL variable we always has CallsWithMultipleSourceFiles.
Since I wrote the script in the first place for my team and was not sure of its usefulness for the clcache community, I left this line. But it can easily be removed if need be to get into the upstream.

env.py
Un1oR: Likely. Thanks, I will try it =)

env.py
Un1oR: I'll try to use $(ClToolExe) and $(ClToolPath) instead of $(ExecutablePath) in Clcache.props. Would be great if it would works.

env.py
Un1oR: Hm... looks like my experience about CallsWithMultipleSourceFiles is outdated now.

Issue #160, opened 2016-07-21
Global cache lock hurts cache hits during heavily concurrent builds


Un1oR:
Hi!
I tried to use clcache for my project and found speed up only on 35% (23 min vs 35 min). This result greatly embarrassed me because ccache at Linux and MacOS boost build time more then 10 times. I try change msbuild to ninja, but it gave similar results.
Then I suggested problem in long waits at lock (at other machine I even got crashes by timeouts). I try cut all lock logic and stats file overwriting. See diff below. Then build time became less then 7 min! It's more then 3 times faster. Looks like absolutely unacceptable overhead for writing statistics.
I'm not sure is access to manifests needs lock too, but looks if it is, we can do it more optimal. And about statistics, I think we can write simple server for writing it with locks, and clcache processes will just sent stat to server and exit without waiting responce. Or we can add option to disable statistics (and lock respectively) as quick fix.

```
--- a/clcache.py
+++ b/clcache.py
@@ -123,6 +123,7 @@ class ObjectCacheLock(object):
     WAIT_ABANDONED_CODE = 0x00000080

     def __init__(self, mutexName, timeoutMs):
+        return
         mutexName = 'Local\\' + mutexName
         self._mutex = windll.kernel32.CreateMutexW(
             wintypes.INT(0),
@@ -133,17 +134,21 @@ class ObjectCacheLock(object):
         assert self._mutex

     def __enter__(self):
+        return
         if not self._acquired:
             self.acquire()

     def __exit__(self, typ, value, traceback):
+        return
         if self._acquired:
             self.release()

     def __del__(self):
+        return
         windll.kernel32.CloseHandle(self._mutex)

     def acquire(self):
+        return
         result = windll.kernel32.WaitForSingleObject(
             self._mutex, wintypes.INT(self._timeoutMs))
         if result not in [0, self.WAIT_ABANDONED_CODE]:
@@ -154,6 +159,7 @@ class ObjectCacheLock(object):
         self._acquired = True

     def release(self):
+        return
         windll.kernel32.ReleaseMutex(self._mutex)
         self._acquired = False

@@ -373,8 +394,9 @@ class PersistentJSONDict(object):

     def save(self):
         if self._dirty:
-            with open(self._fileName, 'w') as f:
-                json.dump(self._dict, f, sort_keys=True, indent=4)
+            pass
+            #with open(self._fileName, 'w') as f:
+            #    json.dump(self._dict, f, sort_keys=True, indent=4)

[...]

PR #173, opened 2016-07-29
Manifest sections

frerich:
This PR introduces the notion of 'manifest sections'. The `ManifestsManager` class provides access to one of up to 256 `ManifestSection` objects, each of which representing one of the up to 256 subdirectories of the `manifests` directory. A manifest section is then the object which provides access to individual manifests.

The PR makes the code base use `ManifestSection` objects throughout. The purpose of this is to make it easy to introduce section-specific locks in a future commit, by adding a single `with manifestSection.lock:` in the `processDirect` function and then having the rest of the code operate on that single locked section, instead of locking the entire cache.

The PR should not introduce any functional change, it's just preparation for adding more fine-grained locking in order to address #160 .

@webmaster128 can you have a look? The `ManifestsManager` is kinda your baby, I'd rather not botch it right away.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -86,25 +86,15 @@ class LogicException(Exception):
         return repr(self.message)
 
 
-class ManifestsManager(object):
-    # Bump this counter whenever the current manifest file format changes.
-    # E.g. changing the file format from {'oldkey': ...} to {'newkey': ...} requires
-    # invalidation, such that a manifest that was stored using the old format is not
-    # interpreted using the new format. Instead the old file will not be touched
-    # again due to a new manifest hash and is cleaned away after some time.
-    MANIFEST_FILE_FORMAT_VERSION = 3
-
-    def __init__(self, manifestsRootDir):
-        self._manifestsRootDir = manifestsRootDir
-
-    def manifestDir(self, manifestHash):
-        return os.path.join(self._manifestsRootDir, manifestHash[:2])
+class ManifestSection(object):
+    def __init__(self, manifestSectionDir):
+        self.manifestSectionDir = manifestSectionDir
 
     def manifestPath(self, manifestHash):
-        return os.path.join(self.manifestDir(manifestHash), manifestHash + ".json")
+        return os.path.join(self.manifestSectionDir, manifestHash + ".json")
 
     def setManifest(self, manifestHash, manifest):
-        ensureDirectoryExists(self.manifestDir(manifestHash))
+        ensureDirectoryExists(self.manifestSectionDir)
         with open(self.manifestPath(manifestHash), 'w') as outFile:
             # Converting namedtuple to JSON via OrderedDict preserves key names and keys order
             json.dump(manifest._asdict(), outFile, indent=2)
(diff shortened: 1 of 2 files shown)


webmaster128:
Sure, I'll look at it in detail. Please give me a couple of days for that.

webmaster128:
While the change looks good overall, I am asking myself, if splitting manifests really reduces locking times.

Right now, the manifests need to be locked in 

1. processDirect (restore / compile&store)
2. postprocessNoManifestMiss (compile&store)
3. postprocessHeaderChangedMiss (compile&store)

Now there question is: is a manifest operation slow? A manifest operation contains of the following things:

* Read a json file from disc and parse 100-300 include entries (300 is an extreme value from a huge amalgamation) (medium fast: 1 IO and some simple memory stuff)
* Read the entire content of those files in binary mode and hash it (slow: multiple harddisk IO)
* Hash the hashes together (fast: all in memory)

The time of this "manifest operation" is the time one call of `getIncludesContentHashForFiles` from #174, let's call that value manifest operation time.

Compared to a compiler call (usage 2. and 3.), a manifest operation will be fast enough such that we don't need to case. It might get interesting when in parallel restore (usage 1.).

webmaster128:
Turns out the manifest operation is slow compared to restoring an object.

I took 3 manifests with approx. 1000, 500 and 100 includes with objects of size 650, 570 and 80 kB. In all cases, the manifest operation results in multiples on copying time.

This is the result with hardlinks enabled.

```
Manifest operation

L 942 includes
M 504 includes
S 127 includes

L 314.64345000000003 ms
M 166.14678000000004 ms
S 48.97450000000001 ms


Copy and hit count operation
L 645 kB
M 565 kB
S 82 kB
L 4.750289999999957 ms
M 4.19571000000003 ms
S 4.198900000000005 ms
```

Without hardlinks, the copying takes twice the time:

```
Manifest operation

L 942 includes
M 504 includes
S 127 includes

L 323.68877000000003 ms
M 162.54767999999996 ms
S 48.13206000000001 ms


Copy and hit count operation
L 645 kB
M 565 kB
S 82 kB
L 8.776709999999976 ms
M 8.943410000000007 ms
S 6.7199200000000125 ms
```

So the resulting factors are 12x-66x for hardlink mode and 7x-37x.

Given those numbers, the question is if we should touch objects and stats locks at all.

-----------------------------

Regarding this PR: Could you rebase it after merging #174, please?

frerich:
Thanks for the review - however, I'm not sure I understand what you're getting at. I think it's no surprise that restoring a plain object is a lot faster than deciding whether it can be restored at all, i.e computing the cache key.

Also, we cannot reduce the number of locks really unless we disallow concurrent invocations.

The purpose of this PR is to lay the foundation for a good fix for #160 in which clcache is much slower than it could be when concurrently restoring objects, because each object restore (and each manifest calculation, and each statistics update) locks the entire object cache. So you cannot restore two or more objects at a time.

To address this, the notion of 'sections' (here: `ManifestSection`, but I imagine the same approach is applicable to the cached objects where we may have `CacheSection` objects) is introduced. The idea is that when interacting with the cache, we only lock the relevant parts of it such that two restores which work on separate sections can safely be performed concurrently. Since there are up to 256 sections, and assuming the hash sum has a normal distribution, there's only a chance of 1/256 that two concurrent restores block each other.

webmaster128:
> think it's no surprise that restoring a plain object is a lot faster than deciding whether it can be restored at all, i.e computing the cache key.

I was indeed expecting the opposite. I thought that

* the number of includes per compilation unit is smaller
* the filesize of the headers is smaller
* the average object size is bigger
* writing to disk (in the copy step) is way slower than reading

However, we're on the same page here and it is shown by initial benchmarks that `getIncludesContentHashForFiles()` is the slow operation. I would like to go through the implementation of manifest sections again, when #174 is done. But the concept is perfectly fine.

PR #174, opened 2016-07-30
Includes content sanity update

webmaster128:
Minor refactoring deduplicating the *includes filepaths* -> *includes content hashes* conversation

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -150,7 +150,8 @@ class ManifestsManager(object):
         return getFileHash(sourceFile, additionalData)
 
     @staticmethod
-    def getIncludesContentHash(listOfHeaderHashes):
+    def getIncludesContentHash(listOfIncludes):
+        listOfHeaderHashes = [getFileHash(expandBasedirPlaceholder(fileName, baseDir)) for fileName in listOfIncludes]
         return HashAlgorithm(','.join(listOfHeaderHashes).encode()).hexdigest()
 
 
@@ -1239,8 +1240,7 @@ def postprocessNoManifestMiss(cache, outputFile, manifestHash, baseDir, sourceFi
     if returnCode == 0 and (outputFile is None or os.path.exists(outputFile)):
         # Store compile output and manifest
         manifest = Manifest(listOfIncludes, {})
-        listOfHeaderHashes = [getFileHash(expandBasedirPlaceholder(fileName, baseDir)) for fileName in listOfIncludes]
-        includesContentHash = ManifestsManager.getIncludesContentHash(listOfHeaderHashes)
+        includesContentHash = ManifestsManager.getIncludesContentHash(listOfIncludes)
         cachekey = ObjectCache.getDirectCacheKey(manifestHash, includesContentHash)
         manifest.includesContentToObjectMap[includesContentHash] = cachekey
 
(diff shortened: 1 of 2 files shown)


clcache.py
```diff
@@ -1384,10 +1384,7 @@ def processDirect(cache, outputFile, compiler, cmdLine, sourceFile):
             listOfHeaderHashes = []
             for fileName in manifest.includeFiles:
                 fileHash = getFileHash(expandBasedirPlaceholder(fileName, baseDir))
-                if fileHash is not None:
-                    # May be if source does not use this header anymore (e.g. if that
-                    # header was included through some other header, which now changed).
-                    listOfHeaderHashes.append(fileHash)
+                listOfHeaderHashes.append(fileHash)
```
frerich: I think this change is correct in itself, but it's based on a false premise: I thought it was suspicious that there's even a comment explaining why `None` may occur, so I checked the commit history and I think I did not port away from `getRelFileHash` correctly in 15eb86fe.

The function not only expanded the `CLCACHE_BASEDIR` placeholder, it also returned `None` if the given file does not exist. Thinking about it, I guess it's plausible: header files which existed in the past may no longer exist, so after expanding the placeholder, we should probably handle the case that the file does not exist. Something like

```python
try:
    fileHash = getFileHash(expandBasedirPlaceholder(fileName, baseDir))
    listOfHeaderHashes.append(fileHash)
except OSError:
    pass
```

...or so.

clcache.py
```diff
@@ -150,7 +150,8 @@ def getManifestHash(compilerBinary, commandLine, sourceFile):
         return getFileHash(sourceFile, additionalData)
 
     @staticmethod
-    def getIncludesContentHash(listOfHeaderHashes):
+    def getIncludesContentHash(listOfIncludes):
+        listOfHeaderHashes = [getFileHash(expandBasedirPlaceholder(fileName, baseDir)) for fileName in listOfIncludes]
         return HashAlgorithm(','.join(listOfHeaderHashes).encode()).hexdigest()
```
frerich: At this point, I guess one could ask why

1. The code needs a list at all, only to join it again in the next line and
2. Why we're hashing a list of hashes here, instead of hashing the file contents directly (changing this probably means that the hash sums change).

I.e. I suspect something like

```python
hasher = HashAlgorithm()
for fileName in listOfIncludes:
    with open(expandBasedirPlaceholder(fileName, baseDir), 'rb') as f:
        hasher.update(f.read())
return hasher.hexdigest()
```

...might be a good idea. However, all this is out of scope for this particular commit, of course (I like how you're taking small steps!).

webmaster128: I was that point 1. before be294fd3dcd. But then the tests did not work anymore and thus I decided to split it in two methods again. Testing the interface at the files list level can be done as well, but it is not trivial.

Regarding 2: you want to be able to differentiate

A.h

```
# content 1
# content 2
```

B.h

```
# content 3
```

and

A.h

```
# content 1
```

B.h

```
# content 2
# content 3
```

so you need to have whatever kind of file separator.

There is room for improvements, sure. But I wanted to have this as small as possible to get to the point where you see this kind of opportunities.

clcache.py
webmaster128: I leave this for tomorrow ;)

webmaster128: I just realized the issue is much more global. A missing include can have multiple meanings:

1. The file was deleted for no reason -> we must not restore the object because it is a compiler error
2. One include changed, making another obsolete. In this case the entire manifest is invalid as it is defined by the list of includes.

I think we need a way to get into the `postprocessHeaderChangedMiss` branch for both cases. Somethink like:

```python
            try:
                includesContentHash = ManifestsManager.getIncludesContentHashForFiles(
                    [expandBasedirPlaceholder(include, baseDir) for include in manifest.includeFiles])
                cachekey = manifest.includesContentToObjectMap.get(includesContentHash)
                if cachekey is None:
                    raise HeaderChanged()

                if cache.hasEntry(cachekey):
                    return processCacheHit(cache, objectFile, cachekey)
                else:
                    postProcessing = lambda compilerResult: postprocessObjectEvicted(
                        cache, objectFile, cachekey, compilerResult)
            except HeaderChanged e:
                postProcessing = lambda compilerResult: postprocessHeaderChangedMiss(
                    cache, objectFile, manifest, manifestHash, includesContentHash, compilerResult)
```

frerich:
Sounds like there are some tests waiting to be written to trigger these issues.

PR #175, opened 2016-07-30
Avoid redundant objectFile = None checks

webmaster128:
The objectFile is only None, when called for multiple source files


frerich:
Good observation, merged. Thanks!

PR #176, opened 2016-07-30
Calculate cache stats from objects+manifests

webmaster128:


frerich:
Looks good, merged. Thanks!

I suppose this means that the 'size used' field of the cache statistics will now be slightly higher than they used to be, which more accurately reflects the disk usage? If so, this would be a nice bugfix to mention in the ChangeLog.

PR #177, opened 2016-07-30
Deduplicate stats key strings

webmaster128:
This defines stats key strings at a single place, allowing the linter
and compiler to check for any further typos in usage.


frerich:
Makes sense, merged. Thanks!

PR #178, opened 2016-07-31
Optimized cache misses in direct mode if CLCACHE_BASEDIR is set

frerich:
In case CLCACHE_BASEDIR was set, the parseIncludesList() function would
automatically make all include paths relocatable by replacing the base
directory with a magic placeholder. In postprocessNoManifestMiss()
however, we would immediately reverse this operation by expanding the
placeholder again.

The only reason why we may want a placeholder is to store relocatable
paths in the Manifest object, so let's do that there and otherwise deal
with full paths.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -217,11 +217,13 @@ class ObjectCacheLock(object):
 
 
 class ObjectCache(object):
-    def __init__(self):
-        try:
-            self.dir = os.environ["CLCACHE_DIR"]
-        except KeyError:
-            self.dir = os.path.join(os.path.expanduser("~"), "clcache")
+    def __init__(self, cacheDirectory=None):
+        self.dir = cacheDirectory
+        if not self.dir:
+            try:
+                self.dir = os.environ["CLCACHE_DIR"]
+            except KeyError:
+                self.dir = os.path.join(os.path.expanduser("~"), "clcache")
 
         manifestsRootDir = os.path.join(self.dir, "manifests")
         ensureDirectoryExists(manifestsRootDir)
--- a/clcache.py
+++ b/clcache.py
@@ -652,6 +652,15 @@ def expandBasedirPlaceholder(path, baseDir):
         return path
 
 
+def collapseBasedirToPlaceholder(path, baseDir):
+    assert path == os.path.normcase(path)
+    assert baseDir == os.path.normcase(baseDir)
+    if path.startswith(baseDir):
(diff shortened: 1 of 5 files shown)


clcache.py
```diff
@@ -650,6 +650,13 @@ def expandBasedirPlaceholder(path, baseDir):
         return path
 
 
+def collapseBasedirToPlaceholder(path, baseDir):
```
webmaster128: Can we add a comment here like

> Both parameters `path` and `baseDir` must have normed case

Otherwise the method does not work

clcache.py
```diff
@@ -1265,16 +1268,20 @@ def postprocessHeaderChangedMiss(
...
     manifest = None
     cachekey = None
 
     if returnCode == 0 and os.path.exists(objectFile):
         # Store compile output and manifest
-        manifest = Manifest(listOfIncludes, {})
-        includesContentHash = ManifestsManager.getIncludesContentHashForFiles(
-            [expandBasedirPlaceholder(include, baseDir) for include in listOfIncludes])
+        if baseDir:
+            baseDir = os.path.normcase(baseDir)
```
webmaster128: Can we norm the basedir at the point where we read it from user input? There we already do some normalization

```python
baseDir = os.environ.get('CLCACHE_BASEDIR')
        if baseDir and not baseDir.endswith(os.path.sep):
            baseDir += os.path.sep
```

webmaster128:
Beautiful separation of concerns

clcache.py
```diff
@@ -1265,16 +1268,20 @@ def postprocessHeaderChangedMiss(
...
+            relocatableIncludePaths = [collapseBasedirToPlaceholder(fn, baseDir) for fn in listOfIncludes]
```
frerich: Sounds good, will change.

clcache.py
frerich: How about an

```python
assert path == os.path.normcase(path)
```

I tend to prefer documenting requirements either in types (but Python doesn't have a terribly strong type system, in particular it has no `Path` class or the like AFAIK) or at least `assert`ions which are enforced at runtime.

clcache.py
frerich: I'll have a look, I tend to agree that normalisation should happen as early as possible.

clcache.py
webmaster128: Even better and self maintaining.

frerich:
The surprising red cross next to the coverage-related checks made me realise that there is no test for the basedir functionality. I'll add that.

PR #182, opened 2016-08-01
Cache sections

frerich:
This PR introduces the notion of `CacheObjects`, objects which represent a single of the up to 256 different subdirectories of the `objects` directory. By making the code work with individual code sections, we get closer to locking just those sections which are being interacted with, improving the concurrency behaviour to address #160 .

After merging this PR, the next step would be to store the statistics (at least the cache hits) per section.

Once that is done, we should be able to implement finer-grained locking by just locking the manifest and cache sections we need.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -384,21 +384,6 @@ class ObjectCache(object):
         # set of includes.
         return ObjectCache.getHash(manifestHash + includesContentHash)
 
-    def hasEntry(self, key):
-        return self.cacheSection(key).hasEntry(key)
-
-    def setEntry(self, key, objectFileName, compilerOutput, compilerStderr):
-        return self.cacheSection(key).setEntry(key, objectFileName, compilerOutput, compilerStderr)
-
-    def cachedObjectName(self, key):
-        return self.cacheSection(key).cachedObjectName(key)
-
-    def cachedCompilerOutput(self, key):
-        return self.cacheSection(key).cachedCompilerOutput(key)
-
-    def cachedCompilerStderr(self, key):
-        return self.cacheSection(key).cachedCompilerStderr(key)
-
     def _cacheEntryDir(self, key):
         return self.cacheSection(key).cacheEntryDir(key)
 
@@ -1259,7 +1244,7 @@ def parseIncludesList(compilerOutput, sourceFile, strip):
 
 def addObjectToCache(stats, cache, objectFile, compilerStdout, compilerStderr, cachekey):
     printTraceStatement("Adding file {} to cache using key {}".format(objectFile, cachekey))
-    cache.setEntry(cachekey, objectFile, compilerStdout, compilerStderr)
+    cache.cacheSection(cachekey).setEntry(cachekey, objectFile, compilerStdout, compilerStderr)
     stats.registerCacheEntry(os.path.getsize(objectFile))
     cfg = Configuration(cache)
     cache.clean(stats, cfg.maximumCacheSize())
@@ -1271,9 +1256,10 @@ def processCacheHit(cache, objectFile, cachekey):
     printTraceStatement("Reusing cached object for key {} for object file {}".format(cachekey, objectFile))
     if os.path.exists(objectFile):
         os.remove(objectFile)
-    copyOrLink(cache.cachedObjectName(cachekey), objectFile)
-    compilerOutput = cache.cachedCompilerOutput(cachekey)
-    compilerStderr = cache.cachedCompilerStderr(cachekey)
+    section = cache.cacheSection(cachekey)
+    copyOrLink(section.cachedObjectName(cachekey), objectFile)
+    compilerOutput = section.cachedCompilerOutput(cachekey)
+    compilerStderr = section.cachedCompilerStderr(cachekey)
     printTraceStatement("Finished. Exit code 0")
     return 0, compilerOutput, compilerStderr
 
@@ -1472,7 +1458,7 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
                 [expandBasedirPlaceholder(include, baseDir) for include in manifest.includeFiles])
             cachekey = manifest.includesContentToObjectMap.get(includesContentHash)
             if cachekey is not None:
-                if cache.hasEntry(cachekey):
+                if cache.cacheSection(cachekey).hasEntry(cachekey):
                     return processCacheHit(cache, objectFile, cachekey)
                 else:
                     postProcessing = lambda compilerResult: postprocessObjectEvicted(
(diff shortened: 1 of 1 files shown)


webmaster128:
What about renaming this to ObjectsSection?
```
Cache
\- Manifests
   \- ManifestSection
\- Objects
   \- ObjectsSection
\- Stats
```

> After merging this PR, the next step would be to store the statistics (at least the cache hits) per section.

Given that manifest evaluation takes 7x–66x the time of objectcopy+stats count, objectcopy+stats only makes up 1.5%–12.5% of the time. Most of that small portion will be the objectcopy operation. Thus I would avoid splitting the stats.

frerich:
Unifying the naming sounds good, I'll have a look. As for the performance - I'm reminded that I neglected creating a performance baseline. ;-/

frerich:
I think I'll leave renaming to a separate step because I'm not sure I want to stick with 'Object' at all. After all, each section not only contains objects but also the compiler output (and maybe eventually the debug information generated via `/Zi`).

frerich:
For what it's worth, it's less important to move fast stuff (such as copying object files) out of the global-lock phase than than moving slow stuff out of it if the goal is to hold the global lock as little as possible.

The issue though is that even if e.g. updating the stats is very fast, it's still a single point of synchronization. So with heavily concurrent cache hits (as is the case in #160), you still only precede one after the other. So the slowdown not only depends on the time which we hold the global lock but also the number of concurrent processes trying to access the cache. I'd like to fix at least the latter dimension by locking per section instead of for the overall lock.

webmaster128:
I think "compilation unit" would be the best name to group everything in a directory. It's key is defined by source and command line and the output is object, stdout and stderr. Given that the object is the main output and all the other stuff is attached to it, the name "object" is still okay and intuitive.

webmaster128:
There is also "entry" in the mix: `cacheEntryDir()`

frerich:
How about this: we just have a `Cache` which corresponds to what `CLCACHE_DIR` is. A cache contains:

* A configuration
* A statistics file
* A repository of manifests (`ManifestsManager`)
* A repository of compiler artefacts (`ArtifactManager`?)

Each of the two managers (maybe we should rename them to `*Repository) has 'entries' identified by 'keys'. Each 'entry' then has different fields (e.g. 'cache.manifestRepository().entry(someKey).manifestFile` or `cache.artifactRepository().entry(anotherKey).objectFile`).

webmaster128:
artifacts sounds good. I would drop the Manager part of the name entirely

Let all capital names be classes:

```
Cache
\- Manifests
   \- ManifestsSection
\- Artifacts
   \- ArtifactsSection
\- Statistics (stats.txt)
\- Configuration (config.txt)
```

The *'entries' identified by 'keys'* part then happens in ManifestsSection and  ArtifactsSection. The `Manifest` entry type already exists. The same could be adopted for Artifact as well.

One could rename the new class CacheSection -> ArtifactsSection or really do it in another step.

ps.: singular or plural before "section"? ManifestSection/ManifestsSection, ArtifactSection/ArtifactsSection.

webmaster128:
Thinking about it, I would keep "Manager" or "Repository" to be able to better differentiate the parent class and the plural of Manifest or Artifact. Both are equally good.

frerich:
I think I'll do a sweep over the discusssed types to rename them to  a more consistent scheme in a future PR, unless you beat me to it.

frerich:
I'm really wondering how to call the `add` method for `Cache` (or rather: `CacheSection`): does it take a `key` and a `cacheEntry`? Somehow, to me a 'cache entry' is the tuple of the key and the value, i.e. the key is included. But just passing `key` and `cacheData` sounds a bit too generic.

Or maybe you shouldn't add cache entries like that but rather do

```python
cacheEntry = CacheEntry
cacheEntry.outputFilePath = 'foo.obj'
cacheEntry.compilerStdout = 'foo.cpp'
cacheEntry.compilerStderr = None
```

...and these property setters actually write to the file, i.e. `Cache.addEntry` just bumps statistics but `CacheEntry` writes the files? Sounds odd to spread that related code in two methods...

webmaster128:
What you describe above shoult be a single Artifact object, I think. Analogue to Manifest.

```
artifact = Artifact
artifact.objectFilePath = 'full-path-to/db/abababababaabab/object'
artifact.stdout = 'foo.cpp'
artifact.stderr = None
```

`stdout` and `stderr` might be the full content instead of the path. I think this better fits the use.

Now you can add the Artefact to the cache

```
cache.addEntry(key, artifact)
```

This can be moved to the ArtifactsSection as well.

frerich:
To me, an artifact is a generic term for something produced by the compiler. I.e. the object file as well as the stdout and stderr output (and potentially the .pdb data) are artifacts. So it's more like an 'ArtifactSet' or so. Or maybe just 'CompilerArtifacts'?

webmaster128:
What about just Artifacts in plural? ArtifactCollection would be an alternative.

While I think ArtifactSet is good as well, it might be confused with a [set](https://en.wikipedia.org/wiki/Set_(abstract_data_type)).

I would avoid the word "compiler" as it adds no value.

frerich:
The more I think about it, the more I like the name `CompilerArtifacts`. The birds-eye view is that a `Cache` is made of an `ArtifactRepository` in which some sort of 'key' is mapped to an `CompilerArtifacts` value. And that value then provides access to whatever we cache. For the sake of finer-grained locking, we may have `ArtifactSections` which roughly correspond to `CacheSections`. So the API may be something like (read `::` as `has type` and `a -> b` as a function mapping a value of type `a` to a value of type `b`):

```
class Cache:
  artifactRepository :: ArtifactRepository
  manifestRepository :: ManifestRepository
  stats :: Statistics
  config :: Configuration

  clean :: void -> void

// Hmm, this class doesn't pull its own weight...
class ArtifactRepository:
    sections :: Generator<ArtifactSection>
    section :: Key -> ArtifactSection

// This class closely resembles a dictionary, maybe we could make good use of some Python-specific tricks such that we can write '<key> in <section>' or 'section[key]'.
class ArtifactSection:
    lock :: CacheLock
    hasArtifact :: Key -> Bool
    getArtifact :: Key -> CompilerArtifacts
    setArtifact :: (Key, CompilerArtifacts) -> void

class CompilerArtifacts:
    objectFilePath :: Path
    compilerStdout :: String
    compilerStderr :: String

class ManifestRepository:
  sections :: Generator<ManifestSection>
  section :: Key -> ManifestSection

class ManifestSection:
    lock :: CacheLock
    hasManifest :: Key -> Bool
    getManifest :: Key -> Manifest
    setManifest :: (Key, Manifest) -> void

class Configuration:
  // same as CacheConfiguration

class Statistics:
  // same as CacheStatistics
```

Not sure if it would work out, but I think there are two interesting observations here:

1. The `*Repository` classes don't really add much value. They just provide access to a sequence of sections (useful for operations which affect all sections, e.g. when cleaning the cache), and a shortcut for accessing a specific section. Right now, you can also clean both repositories independently - but is there a good reason for that other than just having a separate function for each? This could be achieved without having two classes, too...

2. The `*Section` classes are *very* similiar, and very dictionary-like. I wonder whether we really only need a generic `CacheSection` class which is somehow parametrised on the type of value it stores, and then that single class is used for both manifest as well as artifact sections.

P.S.: We are kind of hijacking this issue, but what better place does GitHub offer for this kind of discussion? Wiki pages? Gists? Something else?

webmaster128:
> P.S.: We are kind of hijacking this issue, but what better place does GitHub offer for this kind of discussion? Wiki pages? Gists? Something else?

I think your last post is a good start for a discussion issue ticket

frerich:
I guess an issue is as good as it gets, so #187 it is.

PR #183, opened 2016-08-01
Renamings part 1

webmaster128:
Initial renaming action


frerich:
Looks good, merged. Thanks!

PR #184, opened 2016-08-01
Centralized code for getting subdirectories in given path

frerich:
This generated was replicated in a couple of places, let's centralize it
in a generic 'subdirectories' generator which optionally returns just
the directory names instead of their absolute paths.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -73,6 +73,13 @@ def filesBeneath(path):
             yield os.path.join(path, filename)
 
 
+def childDirectories(path, absolute=True):
+    for entry in os.listdir(path):
+        absPath = os.path.join(path, entry)
+        if os.path.isdir(absPath):
+            yield absPath if absolute else entry
+
+
 def normalizeBaseDir(baseDir):
     if baseDir:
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -73,6 +73,13 @@ def filesBeneath(path):
             yield os.path.join(path, filename)
 
 
+def subdirectories(path, absolute=True):
```
webmaster128: From the name it is not clear to me, if this is recursive or not. What about `directChildDirectories`?

frerich: I think "direct" and "child" are redundant (a child is a direct descendant per definition, otherwise it would be a "grand child", "grand grand child" etc.). However, "childDirectories" seemed like a very uncommon name to me.

My personal style is to use "files" and "subdirectories" as the non-recursive ones, and add a 'Beneath' for the functions which get all descendants, not just children.

webmaster128: I like childDirectories. However, it should not matter much.

frerich:
Well I don't care too much either way, and you bothered to review the code so there's no harm in changing the name. ;-)

PR #185, opened 2016-08-01
Renamings 2

webmaster128:
This removes the dependency from Statistics on Cache, making it possible to test components individually.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -294,8 +294,10 @@ class Cache(object):
         timeoutMs = int(os.environ.get('CLCACHE_OBJECT_CACHE_TIMEOUT_MS', 10 * 1000))
         self.lock = CacheLock(lockName, timeoutMs)
 
+        self.statistics = Statistics(os.path.join(self.dir, "stats.txt"))
+
     def openStatistics(self):
-        return Statistics(os.path.join(self.dir, "stats.txt"))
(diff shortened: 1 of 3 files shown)


clcache.py
```diff
@@ -294,6 +294,9 @@ def __init__(self, cacheDirectory=None):
         timeoutMs = int(os.environ.get('CLCACHE_OBJECT_CACHE_TIMEOUT_MS', 10 * 1000))
         self.lock = CacheLock(lockName, timeoutMs)
 
+    def openStatistics(self):
+        return Statistics(os.path.join(self.dir, "stats.txt"))
```
frerich: About being a bit more funky (may I say 'pythonic'?) here and define a property which returns a context manager? Like:

```python
@property
def statistics(self):
    return Statistics(os.path.join(self.dir, "stats.txt"))
```

This allows writing

```python
with cache.statistics as stats:
    stats.doThis()
```

webmaster128: I just implemented that in the last commit. But I did not know this property thing – will change.

webmaster128:
Check. The property was not necessary because I changed Statistics such that I __enter__ and __exit__ the same object (just like cache.lock).

frerich:
Merged, thanks!

PR #186, opened 2016-08-02
Prevent usage of Statistics instance outside of a with block

webmaster128:
We need to call `__enter__` to ensure the current stats are reloaded from
disk.


webmaster128:
This just makes things complicated since it prevents comparing two stats objects. For now, we just need to take care that `cache.statistics` only reloads data from disk when used in a `with`.

PR #188, opened 2016-08-02
Remove dependency from Configuration on Cache

webmaster128:
analogue to Statistics

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -9,7 +9,6 @@
 from ctypes import windll, wintypes
 import codecs
 from collections import defaultdict, namedtuple
-from contextlib import closing
 import errno
 import hashlib
 import json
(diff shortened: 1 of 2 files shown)


frerich:
Looks good in principle, but I wonder whether you couldn't save code by dropping the explicit `__enter__`/`__exit__` definitions and rather defining a property on `Cache` like:

```python
@property
@contextlib.contextmanager
def configuration(self):
    config = Configuration(os.path.join(self.dir, "config.txt"))
    try:
        yield config
    finally:
        config.save()
```

...which is really just the same as

```python
@property
def configuration(self):
    return contextlib.closing(Configuration(os.path.join(self.dir, "config.txt"))
```

webmaster128:
This adds a bunch of indirection and dependencies where you need to constantly check documentation. I don't see a benefit, as the content of `__enter__` needs to move to the constructor and the content of `__exit__` needs to move to a `save()` method.

Additionally, I want to be able to not store stuff to disk in case of an error as it may corrupt data. This extension can easily be made in out `__exit__`s.

frerich:
Ah well, I'd have thought using some standard library functionality for the sake of code reuse might simplify things, but I don't care strongly enough to really try it - so let's just merge this. It certainly moves things in the right direction. Thanks! 👍

Issue #187, opened 2016-08-02
Harmonize naming in source code, introduce dedicated types


frerich:
We're being somewhat inconsistent with the naming in the clcache source code, which is quite confusing. We recently started introducing manifest sections as well as cache sections to permit implementing a finer-grained locking strategy, but that didn't exactly make things any easier.

This ticket is meant to serve as a discussion forum for where we may want to go. Here's a moonshot idea:

The birds-eye view is that a `Cache` is made of an `ArtifactRepository` in which some sort of 'key' is mapped to an `CompilerArtifacts` value. And that value then provides access to whatever we cache. For the sake of finer-grained locking, we may have `ArtifactSections` which roughly correspond to `CacheSections`. So the API may be something like (read `::` as `has type` and `a -> b` as a function mapping a value of type `a` to a value of type `b`):

```
class Cache:
  artifactRepository :: ArtifactRepository
  manifestRepository :: ManifestRepository
  stats :: Statistics
  config :: Configuration

  clean :: void -> void

// Hmm, this class doesn't pull its own weight...
class ArtifactRepository:
    sections :: Generator<ArtifactSection>
    section :: Key -> ArtifactSection

// This class closely resembles a dictionary, maybe we could make good use of some Python-specific tricks such that we can write '<key> in <section>' or 'section[key]'.
class ArtifactSection:
    lock :: CacheLock
    hasArtifact :: Key -> Bool
    getArtifact :: Key -> CompilerArtifacts
    setArtifact :: (Key, CompilerArtifacts) -> void

class CompilerArtifacts:
    objectFilePath :: Path
    compilerStdout :: String
    compilerStderr :: String

class ManifestRepository:
  sections :: Generator<ManifestSection>
  section :: Key -> ManifestSection

class ManifestSection:
    lock :: CacheLock
    hasManifest :: Key -> Bool
    getManifest :: Key -> Manifest
    setManifest :: (Key, Manifest) -> void

class Configuration:
  // same as CacheConfiguration

class Statistics:
  // same as CacheStatistics
```

Not sure if it would work out, but I think there are two interesting observations here:

1. The `*Repository` classes don't really add much value. They just provide access to a sequence of sections (useful for operations which affect all sections, e.g. when cleaning the cache), and a shortcut for accessing a specific section. Right now, you can also clean both repositories independently - but is there a good reason for that other than just having a separate function for each? This could be achieved without having two classes, too...

[...]

PR #189, opened 2016-08-02
Extract CompilerArtifactsRepository from Cache

webmaster128:
this doe not change any functionality

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -274,56 +274,32 @@ class CacheSection(object):
                 f.write(compilerStderr.encode(CACHE_COMPILER_OUTPUT_STORAGE_CODEC))
 
 
-class Cache(object):
-    def __init__(self, cacheDirectory=None):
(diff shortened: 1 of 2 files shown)


webmaster128:
Check, rebased. This is a big change bit it is just moving around existing code.

frerich:
A lot of diff, but it seems straightforward. Thanks! Do you plan to do more refactorings according to the ideas discussed in #187 ? Just so that we don't step on each others toes, this stuff is bound to cause merge conflicts.

webmaster128:
I can do all of the renaming if there is no time pressure. I somehow like this cleanup stuff when the brain is already fucked. It's a bit meditative.

frerich:
Sure, go ahead. And indeed, I think a straightforward (but not totally trivial) refactoring if you have a clear plan definitely has a meditative effect.

PR #190, opened 2016-08-03
Rename class to CompilerArtifactsSection

webmaster128:


frerich:
Looks good, merged. Thanks!

PR #192, opened 2016-08-04
Added wrapper for Popen for run under Visual Studio.

Un1oR:
I found clcache not working properly under Visual Studio.
When I change header file and try to build I get same hash for source so get same object from cache. It happens because information about includes doesn't saved in manifests. Moreover I see messages from cl.exe calls with /showIncludes in Visual Studio console.
The reason is Visual Studio defines VS_UNICODE_OUTPUT environment variables which affects cl.exe (more at https://blogs.msdn.microsoft.com/freik/2006/04/05/how-to-get-cl-exe-proxy-filters-like-stlfilt-working-under-vs2005/).
I wrote simple wrapper for Popen which unset VS_UNICODE_OUTPUT if it was set.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -90,6 +90,14 @@ def normalizeBaseDir(baseDir):
         return None
 
 
+def popen(*args, **kwargs):
+    if 'env' not in 'kwargs':
+        kwargs['env'] = os.environ.copy()
+    if 'VS_UNICODE_OUTPUT' in kwargs['env']:
+        del kwargs['env']['VS_UNICODE_OUTPUT']
+    return Popen(*args, **kwargs)
+
+
 class CacheLockException(Exception):
     pass
 
@@ -326,7 +334,7 @@ class CompilerArtifactsRepository(object):
     def computeCompilerArtifactsKey(compilerBinary, commandLine):
         ppcmd = [compilerBinary, "/EP"]
         ppcmd += [arg for arg in commandLine if arg not in ("-c", "/c")]
-        preprocessor = Popen(ppcmd, stdout=PIPE, stderr=PIPE)
+        preprocessor = popen(ppcmd, stdout=PIPE, stderr=PIPE)
         (preprocessedSourceCode, ppStderrBinary) = preprocessor.communicate()
 
         if preprocessor.returncode != 0:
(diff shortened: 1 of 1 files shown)


webmaster128:
Looks like a solid solution to me.

Two additional suggestions:
* Could you add a short comment about why we have the custom popen including the link from above? Just want to make sure it is not killed later on
* Could you remove the explicit `Popen` import from `from subprocess import Popen, PIPE` such that we don't have Popen and popen in the global namespace

I guess we don't need to remove VS_UNICODE_OUTPUT in runJobs(), right?

frerich:
I don't really like how using a lower-case `p` instead of an `P` actually causes some environment variable to be deleted, that's somewhat unintuitive.

How about making `computeCompilerArtifactsKey` call `invokeRealCompiler` for invoking the real compiler, and then you can simply remove the environment variable (if present) in `invokeRealCompiler`, so that you don't need a separate function anymore?

webmaster128:
> How about making computeCompilerArtifactsKey call invokeRealCompiler for invoking the real compiler

The difference is String vs. binary STDOUT/STDERR, such that invokeRealCompiler would need to blow up.

What about a COMPILER_ENV constant, that can be passed into every Popen call?

```python
COMPILER_ENV = {k:v for k, v in os.environ.items() if k != 'VS_UNICODE_OUTPUT'}
```

frerich:
I think it's  abit of a shame to have `Popen` calls around when we already have a nicely named `invokeRealCompiler` method.

I suppose making `invokeRealCompiler` always return bytes and then decoding on the caller side would be obnoxious, because there are many callers. Or maybe the `compilerResult` tuple returned should be turned into a real object which has `stdoutBinary` and `stdoutString` members, the latter of which returning the decoded binary data.

I suppose a short-term solution might be to just add another optional argument to `invokeCompiler`, like `captureAsString` which is set to `True`. That way, `computeCompilerArtifactsKey` could just pass `captureAsString=False` to get binary data.

webmaster128:
I like the "short-term solution".

> Or maybe the compilerResult tuple returned should be turned into a real object which has stdoutBinary and stdoutString members, the latter of which returning the decoded binary data.

In case of preprocessing, this is a lot of megabytes of data being decoded and flying around in memory for no real use.

frerich:
The point of making `compileResult` was that `stdoutString` and `stderrString` could be properties in the Python sense, i.e. there are getter functions which perform the conversion on demand, not up-front.

frerich:
@Un1oR I now implemented your suggestion in a slightly different manner, in #221 -- can you give that one a try?

Un1oR:
@frerich I'll try it with joy, but only after 2nd of september.

frerich:
#221 was merged by now, I think that obsoletes this PR. Thanks for the patch and research, it's what made #221 possible in the first place! I never knew about that environment variable.

PR #193, opened 2016-08-04
Compilerartifacts

frerich:
This introduces (and makes use of) a new `CompilerArtifacts` type which is now returned (and expected) in `CompilerArtifactsSection.setEntry` and `CompilerArtifactsSection.getEntry`. I think this the last piece of implement the naming discussed in #187 .

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -274,6 +274,11 @@ class CompilerArtifactsSection(object):
             with open(self.cachedCompilerStderrName(key), 'wb') as f:
                 f.write(compilerStderr.encode(CACHE_COMPILER_OUTPUT_STORAGE_CODEC))
 
+    def getEntry(self, key):
+        if not os.path.exists(self.cacheEntryDir(key)):
+            return None
+        return CompilerArtifacts(self.cachedObjectName(key), self.cachedCompilerOutput(key), self.cachedCompilerStderr(key))
+
 
 class CompilerArtifactsRepository(object):
     def __init__(self, compilerArtifactsRootDir):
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -56,6 +56,7 @@
 #   value: key in the cache, under which the object file is stored
 Manifest = namedtuple('Manifest', ['includeFiles', 'includesContentToObjectMap'])
 
+CompilerArtifacts = namedtuple('CompilerArtifacts', ['objectFilePath', 'compilerStdout', 'compilerStderr'])
```
webmaster128: Can the two fields be simplified to?

CompilerArtifacts.stdout
CompilerArtifacts.stderr

I mean, all compiler artifacts come from the compiler.

frerich: Sure - I was under the (apparently misguided) assumption that `stdout` and `stderr` are reserved identifiers or so, but apparently not so!

That reminds me though that using `dir` is a builtin function (i.e. it's imported by default since `'dir` in dir(__builtins__)` is true), maybe not a good choice for the `Cache.dir` field.

webmaster128: > That reminds me though that using dir is a builtin function (i.e. it's imported by default since 'dir' in dir(__builtins__) is true), maybe not a good choice for the Cache.dir field.

Looks like it is not an issue when used as a member instead of a free symbol. There is not naming clash between Cache.dir and dir.

However, Cache.directory might be equally good.

frerich:
I think this closes  #187 .

PR #194, opened 2016-08-04
Drop support for the CL environment variable in jobCount()

webmaster128:
`CL` and `_CL_` are never respected or tested in the entire project, making
it basically impossible to rely on them. This adds a note on that.

If someone wants to get `CL` or `_CL_` support, it has to be implemented in
a much more general way.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -140,6 +140,12 @@ CLCACHE_OBJECT_CACHE_TIMEOUT_MS::
     getting ObjectCacheLockExceptions with return code 258 (which is the
     WAIT_TIMEOUT return code).
 
+Known limitations
+~~~~~~~~~~~~~~~~~
+
+* https://msdn.microsoft.com/en-us/library/kezkeayy.aspx[CL and +_CL_+] environment
+  variables are not supported.
+
 How clcache works
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


frerich:
I opened #196 now to track progress on adding support for the mentioned environment variables.

As for the 'known limitations' section in the README: I recently started some effort to document such things our wiki (right now, there's just a [Caveats page](https://github.com/frerich/clcache/wiki/Caveats)) and then referencing that from the README since the README is getting a bit long for my test. What do you think about having a 'Known Issues' section?

webmaster128:
Good, thanks.

I don't mind where it is documented. Feel free to copy the content to some other place and I'll remove it from this commit. But I would not necessarily call the section "issues". It is just one thing the product does not support, like maybe thousands of other features. Nobody noticed this feature missing for years. I would call it an issue when we claim to support it but it is not working.

In Github it is an issue of cause, where "issue" is used synonym for "ticket".

frerich:
Well I think it's a bug in either case, since not supporting those environment variables can make clcache both incorrectly restore wrong .obj files as well as miss opportunities for restoring matching object files. However I agree that not every bug needs to be fixed right away, let's keep track of it and see how popular it is (I didn't know about `_CL_` myself...).

webmaster128:
> since not supporting those environment variables can make clcache both incorrectly restore wrong .obj files as well as miss opportunities for restoring matching object files.

You are right, I misjudged this thing.

PR #195, opened 2016-08-04
Fixed /MP causing a lot of cache misses

frerich:
I noticed that when compiling two source files individually and then
compiling both of them in one go while specifying /MP would not cause
any cache hits. It turns out that this is because the /MP switch
contributes to the hashed command line, and when reinvoking itself
recursively, clcache does not strip the /MP switch for the individual
child processes.

Let's fix this by stripping the flag in both direct as well as indirect
mode. I think stripping this flag even in direct mode should be safe
since it doesn't influence the preprocessor. I think that technically,
we could strip a couple more flags in direct mode, though we need to
keep all those relevant to the preprocessor.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -173,7 +173,11 @@ class ManifestRepository(object):
 
         # NOTE: We intentionally do not normalize command line to include
         # preprocessor options. In direct mode we do not perform
-        # preprocessing before cache lookup, so all parameters are important
+        # preprocessing before cache lookup, so all parameters are important.
+        # One of the few exceptions to this rule is the /MP switch, which only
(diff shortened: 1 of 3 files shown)


clcache.py
```diff
@@ -173,7 +173,11 @@ def getManifestHash(compilerBinary, commandLine, sourceFile):
 
         # NOTE: We intentionally do not normalize command line to include
         # preprocessor options. In direct mode we do not perform
-        # preprocessing before cache lookup, so all parameters are important
+        # preprocessing before cache lookup, so all parameters are important.
+        # TOne of the few exceptions to this rule is the /MP switch, which only
```
frerich: 👍

PR #197, opened 2016-08-04
Adding a simple 'performance test'

frerich:
This test benchmarks compiling N equal source files (the exact number is
configurable via a NUM_SOURCE_FILES constant in the test code)
sequentially with a cold cache, then with a hot cache (i.e. 100% hit
rate) and then with a hot cache and concurrent compilation (which should
be even faster assuming that there is no global locking).

I'm not sure whether it's possible to actually verify some timing values
automatically, these timings depend very much on the particlar machine
and system load. So for now, I just verify that the cache contents
change as expected and then print some timings at the end.

We do however ask Appveyor to run pylint on the test code automatically
to help ensuring it's in a tidy state.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -48,6 +48,7 @@ build_script:
   - pylint --rcfile=.pylintrc clcache.py
   - pylint --rcfile=.pylintrc unittests.py
   - pylint --rcfile=.pylintrc integrationtests.py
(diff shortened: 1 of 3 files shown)


frerich:
The point of this is to have a benchmark to run manually while working on #160 .

PR #198, opened 2016-08-06
Analyze environment

webmaster128:
This is the first step of dealing with the `CL` and `_CL_` environment variables.

Tests are about to follow.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -487,6 +487,7 @@ class Configuration(object):
 
 
 class Statistics(object):
+    CALLS_WITH_UNSUPPORTED_ENVIRONMENT = "CallsWithUnsupportedEnvironment"
     CALLS_WITH_INVALID_ARGUMENT = "CallsWithInvalidArgument"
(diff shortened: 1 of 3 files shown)


webmaster128:
Check, I think this is it for CL and _CL_

frerich:
I am currently on vacation without Access to my development environment, so I cannot test this: wouldn't it be easier to pass CL and _CL_ to the function which also parses response files, and then prepend resp. append the parsed argument vector to the effective command line returned by expandCommandLine?

The only complication I can think of is that the environment variables should get un-set in case clcache invokes itself for multiple source file arguments.

webmaster128:
Sure, it is possible to implement parsing the variables. But it's not going to be easier. A lot of tests need to be written and a lot of code need to be touched potentially, since we need to keep our modified environment, which also conflicts with #192.

This just aims to ensure correctness as a short to mid-term solution until someone needs `CL` and `_CL_` working.

webmaster128:
Closing this one as I guess it is easier to implement support for `CL` and `_CL_` than counting misses.

PR #200, opened 2016-08-11
Calls to non-existing includes should force a cache miss to display compiler error

hubx:
At the moment clcache will fail when parsing a non-existing header to generate `includesContentHash` for `ManifestRepository.getIncludesContentHashForFiles(listOfIncludes)` with a non-existing include in `listOfincludes`
>          Traceback (most recent call last):
           File "clcache.py", line 1518, in <module>
           File "clcache.py", line 1409, in main
           File "clcache.py", line 1439, in processCompileRequest
           File "clcache.py", line 1476, in processDirect
           File "clcache.py", line 188, in getIncludesContentHashForFiles
           File "clcache.py", line 188, in <listcomp>
           File "clcache.py", line 688, in getFileHash
         FileNotFoundError: [Errno 2] No such file or directory: 'missing.h'

The proposed is not a elegant solution. I don't see a clever way to raise a ForceCacheMissException which could be handled. Please feel free adapt as you wish.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -186,6 +186,8 @@ class ManifestRepository(object):
     @staticmethod
     def getIncludesContentHashForFiles(listOfIncludesAbsolute):
         listOfIncludesHashes = [getFileHash(filepath) for filepath in listOfIncludesAbsolute]
+        if None in listOfIncludesHashes:
+            return None
         return ManifestRepository.getIncludesContentHashForHashes(listOfIncludesHashes)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


webmaster128:
See also https://github.com/frerich/clcache/pull/179

The problem is: a missing include does mean that it is a miss. It can also be a header change and the include might be obsolete.

webmaster128:
Could you please check, if #179 does the job?

hubx:
@webmaster128 Technically #179 sounds much more profound what I did here. I saw that you are still committing to that PR just now. Ready to test it when you are. :)

webmaster128:
> I saw that you are still committing to that PR just now. Ready to test it when you are. :)

I am happy with the current state in terms of correctness, thus I pinged you here. Please ignore the coverage fails in #179. Those happen because integrationtests are not counted properly.

PR #203, opened 2016-08-13
Handle CL and _CL_ environment variables

webmaster128:


frerich:
Looks good - I had a couple of thoughts while reading the patch, but nothing which really warrants breching a Diskussion from the Zaun. ;-) The only thing I'm really missing: mentioning this fix in the  ChangeLog. Can you add that?

webmaster128:
Sure. Added the changelog entry and tweaked the README a bit.

@frerich Could you check the coverage tool again? I get a lot of uncovered code reported around reinvokePerSourceFile() and runJobs() even though calls with multiple sources are now explicitly added to the integrationtests.

frerich:
Merged, thanks! I'll have a look at the code coverage stuff, maybe I can figure out what's going on.

PR #204, opened 2016-08-15
Fix race condition between hasEntry and addObjectToCache

Jimilian:
This commit removes inconsistency between checking `hasEntry` and creating new one. Now both actions are implemented with lock.

Fix for https://github.com/frerich/clcache/pull/179#issuecomment-239515475:
```python
           File "clcache.py", line 1568, in <module>
           File "clcache.py", line 1447, in main
           File "clcache.py", line 1477, in processCompileRequest
           File "clcache.py", line 1547, in processDirect
           File "clcache.py", line 1530, in <lambda>
           File "clcache.py", line 1358, in postprocessHeaderChangedMiss
           File "clcache.py", line 1295, in addObjectToCache
           File "clcache.py", line 282, in setEntry
           File "clcache.py", line 777, in copyOrLink
         FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'C:\\clcache\\objects\\4c\\4cdbbffca1bd7564d2756447347f0389\\object.tmp' -> 'C:\\clcache\\objects\\4c\\4cdbbffca1bd7564d2756447347f0389\\object'
```


webmaster128:
This destroys all concurrency since only one compile process can run a a time. The idea behind `postProcessing` is that we unlock the cache during the compile process.

The more interesting question I guess is: why would a buildsystem build the same object file multiple times in parallel, which is when happens in this stacktrace, as far as I can see.

Jimilian:
Sorry, I thought that `cache.lock` is related only to one cache object. Not to entire cache.

Question is pretty good. I can imagine two reasons:
1) Two different objects have same hash. Chance is pretty small, but it exists. In this case this fix is not able to help.
2) Your PR - #179 :)

So, if it breaks concurrency I will close this PR and just create for you.

webmaster128:
>  2) Your PR - #179 :)

Good idea, I'll try to check that. Let me know if that stacktrace happens in non-#179 setups as well.

I think you can close this in any case.

PR #211, opened 2016-08-22
Profiling

frerich:
This PR, inspired by #207, adds support for a new `CLCACHE_PROFILE` environment variable. If this variable is set, each clcache invocation generates coverage information in external files. The generated data can be processed into a final report using a little `showprofilereport.py` script.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1581,7 +1581,7 @@ def processNoDirect(cache, objectFile, compiler, cmdLine, environment):
 
 if __name__ == '__main__':
     if 'CLCACHE_PROFILE' in os.environ:
-        invocationHash = getStringHash(','.join(sys.argv))
-        cProfile.run('main()', filename='clcache-{}.prof'.format(invocationHash))
+        INVOCATION_HASH = getStringHash(','.join(sys.argv))
+        cProfile.run('main()', filename='clcache-{}.prof'.format(INVOCATION_HASH))
     else:
         sys.exit(main())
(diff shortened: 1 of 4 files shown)


clcache.py
```diff
@@ -7,6 +7,7 @@
 # root directory of this project.
 #
 from ctypes import windll, wintypes
+import cProfile
```
webmaster128: Do you know how much time this import takes? I think it was better placed in the CLCACHE_PROFILE block (plus ` # pylint: disable=wrong-import-position`)

frerich: I tried it locally, and got

```
C:\Users\Frerich\src\clcache>python -m timeit -- "import cProfile"
1000000 loops, best of 3: 0.418 usec per loop
```

...so it seemed to be a premature optimisation.

It's 1-2 µs in a slow-io environment. The strange thing is that this is much faster than reading a single file – but well, Python seems to take care of that, which is great.

Jimilian: Probably, it because after first several iterations it goes to cache (system or disk cache).

frerich: The disk cache may have something to do with it, but it's also the fact that the Python interpreter stores the byte code of the module to disk and reuses that. Try creating e.g. a file `a.py` with

```python
import b
```

and `b.py` with e.g.

```python
def f(): pass
```

and then run `python a.py`. Notice how `b.pyc` is created - it's a binary version of `b.py`.

PR #213, opened 2016-08-22
Cleanup cleaning and removing of compiler artifacts

webmaster128:


clcache.py
```diff
@@ -313,26 +313,17 @@ def section(self, key):
...
-        for o in removedObjects:
-            dirPath = self.section(o).cacheEntryDir(o)
-            if not os.path.exists(dirPath):
-                continue  # May be if object already evicted.
-            objectPath = os.path.join(dirPath, "object")
-            if os.path.exists(objectPath):
-                # May be absent if this if cached compiler
-                # output (for preprocess-only).
-                fileStat = os.stat(objectPath)
-                stats.unregisterCacheEntry(fileStat.st_size)
```
frerich: Where did this call go? If an entry is removed, surely the statistics should reflect that by reporting a lower number of entries and a smaller cache size?

webmaster128: removeEntry is only used in clean() and in Cache.clean() the number of entries as well as the cache size is recalculated.

frerich: Ah, so

1. The method could be "private" now (i.e. get a leading underscore) and
2. It's worth researching why `removedItems` in `postprocessHeaderChangeMiss` is unused. It looks like this happened in 39765c24 which apparently also renderes the `MAX_MANIFEST_HASHES` constant unused. Does this mean that the size of manifest is no longer bounded? I didn't notice that... but I guess it's actually a good thing (otherwise two invocations might be considered equal if the first 100 includes are the same) and shouldn't hurt.

webmaster128: I was thinking about making it private but then thought that there should be a public interface to remove an entry from CompilerArtifactsRepository. This is something that should be tested and can be reused later (i.e. some follow-up of 2.)

2.: This is about the list in `manifest.includesContentToObjectMap`, not the includes list. The idea to have a maximum number of entries in this list is not too bad in general. But the implementation was not good: Why pop the last item of a dict? Is this a random element? Is it a recent entry? Is it an old entry? 39765c24fd indeed removed this semi-clever magic, mostly because it (temporarily) removed the ability to change existing manifests. So right now `MAX_MANIFEST_HASHES` is unused. The question is, if this ever needs to be re-implemented, because it will only be used, when the same source file is compiled with more than 100 different include values. And even then there is no reason to stop at 100. Allowing 1000 entries is just fine as well.

frerich:
Ok, you convinced me. Merged, thanks!

PR #214, opened 2016-08-23
Cachekey symmetry

webmaster128:
This creates some symmetry in naming as recently found in #155.


frerich:
Looks good, merged. Thanks!

PR #216, opened 2016-08-24
Simplify hashing included code

frerich:


webmaster128:
While this change makes sense for the current code base, I think it harms the work towads updating a manifest. For example, the includes map is likely to change from a path->hash map to a path->list_of_hashes map, in order to be able to deal with multiple include contents.

Removing the redundant reading of includes when creating a manifest is cool. But I think that part will be generalized and changed a lot.

So I would just open an issue pointing out the problem and try to implement it with/after a solution for #212

frerich:
I see, I don't know what you're planning to do with #212 exactly, so I'll just keep this PR around.

frerich:
This got obsoleted now that #222 was merged.

PR #219, opened 2016-08-25
Use scandir instead of listdir if scandir exists

Jimilian:
`scandir` could be found in scandir module or in `os` (since Python 3.5)
According python [docs](https://docs.python.org/3.5/library/os.html#os.listdir) it's much better than old `listdir`.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -25,7 +25,8 @@ VERSION = "3.2.0-dev"
 HashAlgorithm = hashlib.md5
 
 # try to use os.scandir or scandir.scandir
-# fall back to os.walk if not found
+# fall back to os.listdir if not found
+# same for scandir.walk
 try:
     import scandir # pylint: disable=wrong-import-position
     WALK = scandir.walk
@@ -33,8 +34,8 @@ try:
 except ImportError:
     WALK = os.walk
     try:
-        LIST = os.scandir
-    except:
+        LIST = os.scandir # pylint: disable=no-name-in-module
+    except AttributeError:
         LIST = os.listdir
 
 # The codec that is used by clcache to store compiler STDOUR and STDERR in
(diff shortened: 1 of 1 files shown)


clcache.py
```diff
@@ -29,8 +29,13 @@
 try:
     import scandir # pylint: disable=wrong-import-position
     WALK = scandir.walk
+    LIST = scandir.scandir
 except ImportError:
     WALK = os.walk
+    try:
+        LIST = os.scandir
+    except:
+        LIST = os.listdir
```
frerich: When would `os.scandir` not be available? Note that our minimum Python version is 3.3.

clcache.py
```diff
@@ -74,10 +79,15 @@ def filesBeneath(path):
 
 
 def childDirectories(path, absolute=True):
-    for entry in os.listdir(path):
-        absPath = os.path.join(path, entry)
-        if os.path.isdir(absPath):
-            yield absPath if absolute else entry
+    for entry in LIST(path):
+        if type(entry) == str:
```
frerich: Even if it turns out that we still need to support `os.listdir` for some reason (see my question above), then it's wasteful to check the type of every entry. We could just do the test once outside the loop via e.g. `if LIST == os.listdir`.

frerich:
Good finding!

clcache.py
webmaster128: os.scandir is documented to be "New in version 3.5."

frerich: Ah oops, sorry. I mixed up the documentation for `os.scandir` and `os.listdir`.

clcache.py
```diff
@@ -74,10 +80,15 @@ def filesBeneath(path):
...
-            yield absPath if absolute else entry
+    supportsScandir = (LIST != os.listdir)
+    for entry in LIST(path):
+        if supportsScandir:
+            if entry.is_dir():
+                yield entry.path if absolute else entry.name
+        else:
+            absPath = os.path.join(path, entry)
+            if os.path.isdir(absPath):
+                yield absPath if absolute else entry
```
Jimilian: `LIST` is more than just a boolean flag. If we want to support `scandir` module (`LIST` could be reference to `scandir.scandir` or `os.scandir`) . If we don't want, I can simplify code a little bit.

So, I can move check outside of `for` loop, I don't think that it's expensive to check boolean variable, but I can do it if you want.

`supportsScandir = (LIST != os.listdir)` could be outside of `childDirectories`, of course. I can do it in such way:
```python
supportsScandir = True
try:
    import scandir # pylint: disable=wrong-import-position
    WALK = scandir.walk
    LIST = scandir.scandir
except ImportError:
    WALK = os.walk
    try:
        LIST = os.scandir # pylint: disable=no-name-in-module
    except AttributeError:
        LIST = os.listdir
        supportsScandir = False
```

frerich:
Yikes, this is getting really messy. So we have *three* functions to choose from for listing directory entries, and only two of them share the same API? 8-[

frerich:
Let's merge this - it clearly improves things and there's no reason to keep this from going in. We can always improve this code lateron.

Thanks!

PR #217, opened 2016-08-24
Per-section locking

frerich:
This PR attempts to improve concurrency by making clcache not lock the entire cache but only affected sections of it.

At the same time, it actually *increases* the duration for which a section is locked in case of a cache miss: we will hold the lock even while the compiler is running, something which seems more correct but which was not viable before because it would greatly impact concurrency.

This PR is meant to supersede #208. It resolves #160 .

PR #222, opened 2016-08-26
Atomic manifest updates and multiple entries per manifest

siu:
A bit late to the party now that #217 is ongoing and will conflict in a few places, nevertheless this fixes a few bugs and concurrency issues.

This PR tries to solve the following issues:
 * Handle alternating include files #212.
 * Fix race condition updating manifest files in postprocess functions (from From https://github.com/frerich/clcache/pull/208#issuecomment-241473339):
> postprocess functions should not assume that the object or manifest cache are in the same state as before. I.e. postprocessNoManifestMiss is run because the manifest didn't exist but by the time it is run the manifest file may exist already. The same applies to postprocessHeaderChangedMiss. That means that the read-update-write of the manifest files needs to be atomic and re-tried by the postprocess functions for each change in the manifests.

If the whole operation of clcache is performed under a lock as proposed in #217 the race condition described above disappears and this PR makes some extra operations (re-reading the manifest file) but still fix the problems with alternating headers.

There are still a few things to consider:
 * In the general case the order of includes in C/C++ is important. There are no tests checking if the file is re-compiled when the order of includes change. For the same reason if a file is included more than once the includesContentHash should be affected.
 * I did not push the release of the lock while computing the current hash of the include files as in #208 because all this is going to change with #217 but it is easy to apply.
 * Reduce the integration tests to the minimum subset (help?).
 * When a matching ManifestEntry is found the entry can be pushed to the top of the Manifest entries to implement LRU.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -129,7 +129,14 @@ class Manifest(object):
         return self._entries
 
     def addEntry(self, entry):
-        self._entries.append(entry)
+        """Adds entry at the top of the entries"""
+        self._entries.insert(0, entry)
+
+    def touchEntry(self, entryIndex):
+        """Moves entry in entryIndex position to the top of entries()"""
+        entry = self._entries[entryIndex]
+        self._entries.pop(entryIndex)
+        self._entries.insert(0, entry)
 
 
 class ManifestSection(object):
@@ -1569,7 +1576,7 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
                 cache, objectFile, manifestSection, manifestHash, sourceFile, compiler, cmdLine,
                 Statistics.registerSourceChangedMiss)
 
-        for entry in manifest.entries():
+        for entryIndex, entry in enumerate(manifest.entries()):
             # NOTE: command line options already included in hash for manifest name
             try:
                 includesContentHash = ManifestRepository.getIncludesContentHashForFiles(
@@ -1578,6 +1585,9 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
                 if entry.includesContentHash == includesContentHash:
                     cachekey = entry.objectHash
                     assert cachekey is not None
+                    # Move manifest entry to the top of the entries in the manifest
+                    manifest.touchEntry(entryIndex)
+                    manifestSection.setManifest(manifestHash, manifest)
 
                     return getOrSetArtifacts(
                         cache, cachekey, objectFile, compiler, cmdLine, Statistics.registerEvictedMiss)
--- a/clcache.py
+++ b/clcache.py
@@ -1401,9 +1401,7 @@ def createManifestEntry(manifestHash, includePaths):
 
 
 def createOrUpdateManifest(manifestSection, manifestHash, entry):
-    manifest = manifestSection.getManifest(manifestHash)
-    if manifest is None:
-        manifest = Manifest([])
+    manifest = manifestSection.getManifest(manifestHash) or Manifest()
     manifest.addEntry(entry)
     manifestSection.setManifest(manifestHash, manifest)
     return manifest
--- a/clcache.py
+++ b/clcache.py
@@ -58,11 +58,11 @@ MAX_MANIFEST_HASHES = 100
 # to use it as mark for relative path.
 BASEDIR_REPLACEMENT = '?'
 
+# ManifestEntry: an entry in a manifest file
 # `includeFiles`: list of paths to include files, which this source file uses
-# `includesContentToObjectMap`: dictionary
-#   key: cumulative hash of all include files' content in includeFiles
-#   value: key in the cache, under which the object file is stored
-Manifest = namedtuple('Manifest', ['includeFiles', 'includesContentToObjectMap'])
+# `includesContentsHash`: hash of the contents of the includeFiles
+# `objectHash`: hash of the object in cache
+ManifestEntry = namedtuple('ManifestEntry', ['includeFiles', 'includesContentHash', 'objectHash'])
 
 CompilerArtifacts = namedtuple('CompilerArtifacts', ['objectFilePath', 'stdout', 'stderr'])
 
@@ -108,10 +108,6 @@ class IncludeNotFoundException(Exception):
     pass
 
 
-class IncludeChangedException(Exception):
-    pass
-
-
 class CacheLockException(Exception):
     pass
 
@@ -123,6 +119,17 @@ class LogicException(Exception):
 
     def __str__(self):
         return repr(self.message)
+
+
+class Manifest(object):
+    def __init__(self, entries):
+        self._entries = entries
+
+    def entries(self):
+        return self._entries
+
+    def addEntry(self, entry):
+        self._entries.append(entry)
 
 
 class ManifestSection(object):
@@ -140,7 +147,9 @@ class ManifestSection(object):
         ensureDirectoryExists(self.manifestSectionDir)
         with open(self.manifestPath(manifestHash), 'w') as outFile:
             # Converting namedtuple to JSON via OrderedDict preserves key names and keys order
-            json.dump(manifest._asdict(), outFile, sort_keys=True, indent=2)
+            entries = [e._asdict() for e in manifest.entries()]
+            jsonobject = {'entries': entries}
+            json.dump(jsonobject, outFile, sort_keys=True, indent=2)
 
     def getManifest(self, manifestHash):
         fileName = self.manifestPath(manifestHash)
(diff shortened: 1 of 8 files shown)


webmaster128:
> In the general case the order of includes in C/C++ is important. There are no tests checking if the file is re-compiled when the order of includes change. For the same reason if a file is included more than once the includesContentHash should be affected.

Since the source code as well as all include contents are hashed, the oder is implicitly fixed.

siu:
@webmaster128 oh yes, that even works for transitive includes, still it would be good to add tests.

frerich:
@siu Good stuff! Indeed, I'm afraid half of this will conflict once #217 is in, and the 'alternating headers' stuff kind of overlaps with #212 . My understanding is that @webmaster128 has something in the pipe for #212 (which is why I'm not merging #216 yet), so I can't quite tell what to do with this PR just yet.

My hope is that we can get #217 ready real soon, so that this PR can be rebased and then we can compare what's left with #212.

All these PR numbers make me dizzy. %-)

webmaster128:
@frerich I did not start working on anything apart the test in #212 and am happy to see you both working on it. I think no matter how #212 is solved, #216 becomes obsolete.

I personally would not touch #212 before #155 is solved, but that might be a matter of taste or use case.

frerich:
@webmaster128 Ah, thanks for the clarification. Since @siu already picked the test from #212 -- doesn't that render #212 obsolete in favour of this PR?

siu:
That works for me, let's wait until #217 is merged and then rebase this one. In the meanwhile I will probably add more tests. You can ignore the changes until then, I am also learning how clcache works with this PR.

As @webmaster128 said, #216 will be obsolete in any case.

@webmaster128 In addition to the two reasons that you point in #155 not updating the Manifest file but creating a new one was also triggering #155. It was the case in your integration test and that is why it is passing now. You pointed that one way to trigger #155 is by cleaning the manifests but not removing the cached objects, I haven't looked at how cleaning the cache works so I am not able to create a test that triggers that. Maybe adding such a test would be a good way to proceed with #155.

frerich:
#217 was now merged, so I think this one could be rebased.

siu:
Updated, I'll wait for your reviews.

I think it can still be improved by pushing the manifest entry to the top of the list when there is a hit. I will add more tests too.

clcache.py
```diff
@@ -219,26 +229,19 @@ def getManifestHash(compilerBinary, commandLine, sourceFile):
...
+        for path in includes:
             try:
-                fileHash = getFileHash(path)
-                if fileHash != includes[path]:
-                    raise IncludeChangedException()
-                listOfIncludesHashes.append(fileHash)
+                listOfHashes.append(getFileHash(path))
             except FileNotFoundError:
-                includeMissing = True
+                raise IncludeNotFoundException
```
frerich: Hmm, does this change mean that in case the statistics are counted slightly differently in case some included file has a different hash, and another included does not exist anymore?

clcache.py
```diff
@@ -1371,24 +1379,24 @@ def processCacheHit(cache, objectFile, cachekey):
...
 
-def createManifest(manifestHash, includePaths):
-    baseDir = normalizeBaseDir(os.environ.get('CLCACHE_BASEDIR'))
-
-    includes = {path:getFileHash(path) for path in includePaths}
-    includesContentHash = ManifestRepository.getIncludesContentHashForFiles(includes)
+def createManifestEntry(manifestHash, includePaths):
+    includesWithHash = [[path, getFileHash(path)] for path in includePaths]
+    includesContentHash = ManifestRepository.getIncludesContentHashForHashes(
+        [hash for include, hash in includesWithHash])
```
frerich: If you keep `includesWithHash` (aka `includes` in the old version) a dictionary, you can simplify the `includesContentHash` definition (as well as `safeIncludes` further down) a bit by using `keys()` and `values()`, i.e. you don't need to unpack both values and then ignore one of them.

Not sure this actually improves anything, but a dictionary seems like a fairly natural data structure in this case to me, doesn't it?

clcache.py
```diff
@@ -1371,24 +1379,24 @@ def processCacheHit(cache, objectFile, cachekey):
         return 0, cachedArtifacts.stdout, cachedArtifacts.stderr, False
 
 
-def createManifest(manifestHash, includePaths):
-    baseDir = normalizeBaseDir(os.environ.get('CLCACHE_BASEDIR'))
```
frerich: Any particular reason you moved this code into `collapseBasedirToPlaceholder` (and `expandBasedirPlaceholder`)? That seems to repeat the `normalizeBaseDir(os.environ.get(..))` call for every included files, doesn't it? I.e. the older version seems more efficient to me on first glance.

clcache.py
```diff
@@ -1371,24 +1379,24 @@ def processCacheHit(cache, objectFile, cachekey):
...
-    return manifest, cachekey
+    safeIncludes = [collapseBasedirToPlaceholder(path) for path, contentHash in includesWithHash]
+    return ManifestEntry(safeIncludes, includesContentHash, cachekey)
+
+
+def createOrUpdateManifest(manifestSection, manifestHash, includePaths):
+    entry = createManifestEntry(manifestHash, includePaths)
+    manifest = manifestSection.getManifest(manifestHash)
+    if manifest is None:
+        manifest = Manifest([])
```
frerich: I think a nice way to shorten this code is to use the `or` operator in Python:

```python
manifest = manifestSection.getManifest(manifestHash) or Manifest([])
```

frerich:
Looks promising! I hope it wasn't too much of a pain to update this PR after #217 was merged.

I only did a first pass through the code now, I think I'll need to do a second pass after reading the tests very carefully since I must admit that I'm not quite sure which issue this PR tries to fix exactly (and how). :-]

siu:
I updated the pull request with (most of) your suggestions. I switched to lists in createManifestEntry because I was worried about the ordering and duplicity of includes but as @webmaster128 was pointing it is already handled by the hashing of the contents.

clcache.py
siu: I found confusing that the handling of the baseDir had to be done outside of the collapseBasedirToPlaceholder or expandBasedirPlaceholder making the surrounding code more complicated. Even more passing the baseDir as argument was not used during the tests as the whole setup is done through the environment variable.

Regarding the performance, it can be extracted to a constant variable at the top of the file.

siu:
I also added the code for reordering the manifest entries when there is a hit in b7cac4c3cb74842a0d6f8644e9140fc0f45545ed

clcache.py
```diff
@@ -125,6 +121,26 @@ def __str__(self):
...
+
+    def addEntry(self, entry):
+        """Adds entry at the top of the entries"""
+        self._entries.insert(0, entry)
+
+    def touchEntry(self, entryIndex):
+        """Moves entry in entryIndex position to the top of entries()"""
+        entry = self._entries[entryIndex]
+        self._entries.pop(entryIndex)
+        self._entries.insert(0, entry)
```
frerich: `pop()` already returns the entry which was popped, so you could simplify to

```python
self._entries.insert(0, self._entries.pop(entryIndex))
```

clcache.py
frerich: Fair enough. For what it's worth, I was just curious, I didn't actually benchmark it. :-)

frerich:
To me, it looks like it's good to go! There are just smaller stylistic things which came to my mind, but nothing really which warrants blocking this improvement from being incorporated.

The main piece missing, to me: an entry for the `ChangeLog` file. However, even after reading the test code and the description of this PR as well as the description of #212 I'm not totally sure what "alternating headers" are. So I don't even know what to write. :-)

What I gathered is that this PR improves things such that a cache hit is performed in some cases where we'd previously unnecessarily resulted in a cache miss, and this is somehow related to two compilations using... 'alternating headers'?

siu:
I tried my best to update the changelog.

Just to rephrase it: these changes try to improve the hit rate when changing between branches.

Let's say I am in branch1 which contains:
 * A.cpp
 * B.h (version 1)

Compiling with clcache (as in master) will create a manifest file for A.cpp with the contents hash of the current includes. Changing then to branch2 with the following contents:
 * A.cpp
 * B.h (version 2)

clcache will find the previous manifest but the contents hash of the include files will not match. This produces a cache miss and the contents of the manifest will be overwritten with the new includesContentHash -> objectHash.

Switching back to branch1 will generate another cache miss as the manifest does not match the content of the includes. This could be avoided as the object is already stored in the cache.

This PR solves that by having multiple entries per manifest file. Each manifest entry contains the previous contents of the manifest: the path to the includes, the hash of the contents of the includes and the hash of the cached object. Note that having the full list of includes in the manifest entries is needed because the full list may have changed (through transitive includes) and will be needed during the validation phase.

frerich:
Aaah, thanks for the explanation! Makes sense. Thanks a lot for the contribution, merged it now!

PR #223, opened 2016-08-29
Don't print "finished" message in processDirect/processNoDirect

webmaster128:
We can do so centrally, on the calling side.


frerich:
Merged, thanks!

PR #230, opened 2016-09-24
Optimized reading compiler output in invokeRealCompiler()

### LINKED ISSUE #229: Lock Timeout even with 60s
Before starting to work on the lockfile based locking again, I wanted to see how clcache has evolved since 3.0.3. Also I wanted to get some reference timing data for later comparison with another locking strategy and also wanted to use it on our internal main project.
We have a build driven by ninja (generated by cmake). I am testing on a 4 core VM so there should be at max 4 clcache instances running at the same time. I am using the 3.3.0 release. A complete build generates about 6500 cache entries.
During the build I get a Lock Timeout Error with the 10s default timeout value and even with a 60s timeout.
I then activated profiling to see where the time might be spent. But with profiling on I am not getting any timeouts. The profiling result of a completed run is attached. [prof.txt](https://github.com/frerich/clcache/files/486913/prof.txt)
Any hints on how to diagnose this further? I started looking at the clcache code but it seams that much has changed since I last looked at the code.

frerich:
As it turns out, using two PIPEs with Popen() makes it fork two threads
internally for reading stdout/stderr, and synchronising those threads
seems to take very long. Some Google research suggests that the threads
can be avoided by writing to temporary files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ clcache changelog
    of leaving the cache in a defective state.
  * Internal: Fixed running integration tests when the clcache source code is
    stored in a path with spaces (GH #206).
+ * Improvement: Optimized communicating with real compiler.
 
 ## clcache 3.3.0 (2016-09-07)
 
--- a/clcache.py
+++ b/clcache.py
@@ -21,6 +21,7 @@ import re
 import signal
 import subprocess
 import sys
+from tempfile import TemporaryFile
 
 VERSION = "3.3.0-dev"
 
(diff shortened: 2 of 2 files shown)


siu:
Just some feedback: these changes don't change the timings on my side.

webmaster128:
> Just some feedback: these changes don't change the timings on my side.

Do you have compiler output (like /showIncludes)?

When cl only prints the source file name on one output stream stream, it is unlikely that this helps.

frerich:
For the record, this PR was meant to address #229 but didn't help in that case either -- however, the original profile data (I summed things up in https://github.com/frerich/clcache/issues/229#issuecomment-249355242) changed considerably. Waiting for mutexes was no longer high up on the list.

siu:
@webmaster128 I tried with cold cache in direct mode which should add "/showIncludes". Apart from that the compiler output is very minimal in my test case, so it is maybe not relevant for benchmarking.

I guess it can be merged if it helps in #229 with a different workload, at least it didn't hurt in my use case.

PR #233, opened 2016-10-05
Ignore SIGINT and SIGTERM

siu:
Currently clcache does not attempt to handle user signals. It may happen that the user presses Ctrl-C while a manifest file is being written or the stats are being updated, etc. in that case the cache may get corrupted.

With this change SIGINT and SIGTERM are ignored avoiding that problem. The signals are ignored for the whole execution of clcache but in the future it can be improved and ignore them in smaller sections.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -19,6 +19,7 @@ import json
 import multiprocessing
 import os
 import re
+import signal
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


siu:
I reordered the includes and had to change the style to make appveyor happy, I can remove those commits from the PR if you don't want them.

frerich:
I think this makes sense, and the patch looks good. Merged, thanks!

PR #237, opened 2016-10-26
Fix crash when a manifest was evicted but the object is in the cache

siu:
Add tests that check if the evicted objects and manifests are handled properly, attempt to reproduce #155.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1432,7 +1432,7 @@ def postprocessUnusableManifestMiss(
     section = cache.compilerArtifactsRepository.section(cachekey)
     with section.lock, cache.statistics.lock, cache.statistics as stats:
         reason(stats)
-        if returnCode == 0 and os.path.exists(objectFile):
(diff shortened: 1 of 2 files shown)


siu:
Finally, it is crashing!

frerich:
Indeed, removing a manifest will not remove all referenced artifacts (because those artifacts are still usable if you don't use the 'direct' mode at all). So it may be that when adding an entry in direct mode, a manifest is missing but the referenced artifact is already there.

I'm curious to see whether the patch I mentioned in https://github.com/frerich/clcache/pull/236#issuecomment-256027348 makes the test pass.

siu:
Yes, it fixes it, all tests are passing.

frerich:
Great job, I think this resolves #155 after all!

siu:
@frerich I think you closed the issue instead of merging it.

frerich:
Ooops @siu, sorry! I seem to have hit the wrong button - that was not intentional. Thanks for catching!

PR #238, opened 2016-10-26
Handle base dir in get manifest hash

siu:
CLCACHE_BASEDIR is not used to cleanup the command line when computing the manifest hash. This is a problem because invocations of clcache from different basedirs will create different manifests for equivalent object files.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -232,9 +232,21 @@ class ManifestRepository(object):
         # One of the few exceptions to this rule is the /MP switch, which only
         # defines how many compiler processes are running simultaneusly.
         commandLine = [arg for arg in commandLine if not arg.startswith("/MP")]
+        arguments, inputFiles = CommandLineAnalyzer.parseArgumentsAndInputFiles(commandLine)
+        collapseBasedirInCmdPath = lambda path: collapseBasedirToPlaceholder(os.path.normcase(os.path.abspath(path)))
+
+        commandLineArgs = []
+        projectSpecificArgs = ("AI", "I", "FU")
+        for k in sorted(arguments.keys()):
+            if k in projectSpecificArgs:
+                commandLineArgs.extend(["/" + k + collapseBasedirInCmdPath(arg) for arg in arguments[k]])
+            else:
+                commandLineArgs.extend(["/" + k + arg for arg in arguments[k]])
+
+        commandLineArgs.extend(collapseBasedirInCmdPath(arg) for arg in inputFiles)
 
         additionalData = "{}|{}|{}".format(
-            compilerHash, commandLine, ManifestRepository.MANIFEST_FILE_FORMAT_VERSION)
+            compilerHash, commandLineArgs, ManifestRepository.MANIFEST_FILE_FORMAT_VERSION)
         return getFileHash(sourceFile, additionalData)
 
     @staticmethod
--- a/clcache.py
+++ b/clcache.py
@@ -97,8 +97,8 @@ def childDirectories(path, absolute=True):
 def normalizeBaseDir(baseDir):
     if baseDir:
         baseDir = os.path.normcase(baseDir)
-        if not baseDir.endswith(os.path.sep):
-            baseDir += os.path.sep
+        if baseDir.endswith(os.path.sep):
+            baseDir = baseDir[0:-1]
         return baseDir
     else:
         # Converts empty string to None
--- a/clcache.py
+++ b/clcache.py
@@ -227,10 +227,12 @@ class ManifestRepository(object):
         compilerHash = getCompilerHash(compilerBinary)
 
         # NOTE: We intentionally do not normalize command line to include
-        # preprocessor options. In direct mode we do not perform
-        # preprocessing before cache lookup, so all parameters are important.
-        # One of the few exceptions to this rule is the /MP switch, which only
-        # defines how many compiler processes are running simultaneusly.
+        # preprocessor options.  In direct mode we do not perform preprocessing
+        # before cache lookup, so all parameters are important.  One of the few
+        # exceptions to this rule is the /MP switch, which only defines how many
+        # compiler processes are running simultaneusly.  Arguments that specify
+        # the compiler where to find the source files are parsed to replace
+        # ocurrences of CLCACHE_BASEDIR by a placeholder.
         commandLine = [arg for arg in commandLine if not arg.startswith("/MP")]
         arguments, inputFiles = CommandLineAnalyzer.parseArgumentsAndInputFiles(commandLine)
         collapseBasedirInCmdPath = lambda path: collapseBasedirToPlaceholder(os.path.normcase(os.path.abspath(path)))
--- a/clcache.py
+++ b/clcache.py
@@ -235,18 +235,18 @@ class ManifestRepository(object):
         arguments, inputFiles = CommandLineAnalyzer.parseArgumentsAndInputFiles(commandLine)
         collapseBasedirInCmdPath = lambda path: collapseBasedirToPlaceholder(os.path.normcase(os.path.abspath(path)))
 
-        commandLineArgs = []
-        projectSpecificArgs = ("AI", "I", "FU")
+        commandLine = []
+        argumentsWithPaths = ("AI", "I", "FU")
         for k in sorted(arguments.keys()):
-            if k in projectSpecificArgs:
-                commandLineArgs.extend(["/" + k + collapseBasedirInCmdPath(arg) for arg in arguments[k]])
+            if k in argumentsWithPaths:
+                commandLine.extend(["/" + k + collapseBasedirInCmdPath(arg) for arg in arguments[k]])
             else:
-                commandLineArgs.extend(["/" + k + arg for arg in arguments[k]])
+                commandLine.extend(["/" + k + arg for arg in arguments[k]])
 
-        commandLineArgs.extend(collapseBasedirInCmdPath(arg) for arg in inputFiles)
+        commandLine.extend(collapseBasedirInCmdPath(arg) for arg in inputFiles)
 
         additionalData = "{}|{}|{}".format(
-            compilerHash, commandLineArgs, ManifestRepository.MANIFEST_FILE_FORMAT_VERSION)
+            compilerHash, commandLine, ManifestRepository.MANIFEST_FILE_FORMAT_VERSION)
         return getFileHash(sourceFile, additionalData)
 
     @staticmethod
--- a/integrationtests.py
+++ b/integrationtests.py
@@ -1073,37 +1073,102 @@ class TestNoDirectCalls(unittest.TestCase):
 
 
 class TestBasedir(unittest.TestCase):
-    def testBasedir(self):
-        with cd(os.path.join(ASSETS_DIR, "basedir")), tempfile.TemporaryDirectory() as tempDir:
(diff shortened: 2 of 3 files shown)


clcache.py
```diff
@@ -232,6 +232,9 @@ def getManifestHash(compilerBinary, commandLine, sourceFile):
         # One of the few exceptions to this rule is the /MP switch, which only
         # defines how many compiler processes are running simultaneusly.
         commandLine = [arg for arg in commandLine if not arg.startswith("/MP")]
+        baseDir = os.environ.get('CLCACHE_BASEDIR')
+        if baseDir:
+            commandLine = [arg.replace(baseDir, BASEDIR_REPLACEMENT) for arg in commandLine]
```
frerich: You could reuse the existing `collapseBasedirToPlaceholder` here and use

```python
commandLine = [collapseBasedirToPlaceholder(arg) for arg in commandLine if not arg.startswith("/MP")]
```

in line 234.

siu: That wouldn't work with the current implementation of collapseBasedirToPlaceholder as it assumes that the BASEDIR appears in the first position of the argument and that the argument is in normcase.

Here arg can be of the form -IC:\basedir\... which doesn't satisfy any of those conditions. We shouldn't change the case either.

collapseBasedirToPlaceholder could be changed like this to make it work:

```
 def collapseBasedirToPlaceholder(path):
     baseDir = normalizeBaseDir(os.environ.get('CLCACHE_BASEDIR'))
-    if baseDir is None:
-        return path
-    else:
-        assert path == os.path.normcase(path)
-        assert baseDir == os.path.normcase(baseDir)
-        if path.startswith(baseDir):
-            return path.replace(baseDir, BASEDIR_REPLACEMENT, 1)
-        else:
-            return path
+    if baseDir:
+        replacement = re.compile(re.escape(baseDir), re.IGNORECASE)
+        return replacement.sub(BASEDIR_REPLACEMENT, path)
+    return path
```

frerich: Ah, true - as it is, `collapseBasedirToPlaceholder` is not applicable. While thinking about whether it makes sense to adjust it as you suggested, it occurred to me that maybe clcache shouldn't always blindly replace the basedir value.

For instance, should it really replace the basedir in preprocessor definitions passed via `-D` on the command line?

frerich: @siu -- ping? :)

siu: Yes, I don't think it is correct to apply the replacement blindly. I can instead add a list of parameters and apply the replacement only to those.

In that case, which arguments should be in the white list? I had a look here (https://msdn.microsoft.com/en-us/library/fwkeyyhe.aspx), these are the arguments that seem good candidates:

 - "/I" https://msdn.microsoft.com/en-us/library/73f9s62w.aspx
 - "/AI" https://msdn.microsoft.com/en-us/library/x1x72k9t.aspx
 - "/FU" https://msdn.microsoft.com/en-us/library/81ex1b0a.aspx
 - "/Yu" https://msdn.microsoft.com/en-us/library/z0atkd6c.aspx

frerich: Indeed, the first three sound like good candidates. I suggest the replacement should be applied to their argument -- we can extend (or shorten) that list later as we learn about new cases.

`/Yu` doesn't seem important right now since clcache doesn't support invocations making use of precompiled headers anyway.

siu:
Summary of the changes in this PR:
 * Added integration tests to show how clcache should behave with /I and /D arguments that include CLCACHE_BASEDIR
 * Apply collapseBasedirToPlaceholder to arguments that begin with /I, /AI, /FU
 * Change collapseBasedirToPlaceholder to find the basedir at any position and ignore the capitalization
 * Change normalizeBaseDir to remove trailing "/" and adapt unit tests. I am not sure about this one, will this break anything? Should I increase the MANIFEST_FILE_FORMAT_VERSION again?
 * Cleanup in TestBasedir
 * Make ASSETS_DIR in integration tests an absolute path (to simplify TestBasedir)

I also noticed that the integration tests have a lot of duplicated code, this could be simplified if all of them inherited from a common base class. Anyway that is for another PR.

Note: github shows the commit in a different order than they were applied.

clcache.py
```diff
@@ -97,8 +97,8 @@ def childDirectories(path, absolute=True):
 def normalizeBaseDir(baseDir):
     if baseDir:
         baseDir = os.path.normcase(baseDir)
-        if not baseDir.endswith(os.path.sep):
-            baseDir += os.path.sep
+        if baseDir.endswith(os.path.sep):
+            baseDir = baseDir[0:-1]
```
frerich: Why did you need to change this?

clcache.py
```diff
@@ -781,15 +788,10 @@ def expandBasedirPlaceholder(path):
...
-        assert path == os.path.normcase(path)
-        assert baseDir == os.path.normcase(baseDir)
-        if path.startswith(baseDir):
-            return path.replace(baseDir, BASEDIR_REPLACEMENT, 1)
-        else:
-            return path
+    if baseDir:
+        replacement = re.compile(re.escape(baseDir), re.IGNORECASE)
+        return replacement.sub(BASEDIR_REPLACEMENT, path)
+    return path
```
frerich: Any reason you dropped the `assert` statements? They seem to express that the function expects that both the given `path` as well as the `baseDir` are normalized, i.e. have not only the same upper/lowercase but also forward slashes converted to backslashes.

integrationtests.py
```diff
@@ -24,7 +24,7 @@
 
 PYTHON_BINARY = sys.executable
 CLCACHE_SCRIPT = os.path.join(os.path.dirname(os.path.realpath(__file__)), "clcache.py")
-ASSETS_DIR = os.path.join("tests", "integrationtests")
+ASSETS_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "tests", "integrationtests")
```
frerich: I have a suspicion, but I'm not quite sure - why did you change this?

siu:
That is true, some changes look pretty random without an explanation. I will try to explain myself better. I can reply you in the PR this time, but just to be clear, in the future do you prefer to have the explanation in each of the commits?

clcache.py
siu: This was needed to make the integration test `TestBasedir.testBasedirIncludeArg` pass. There the include path is generated by `os.getcwd()` which doesn't include the trailing separator. As `normalizedBasedir` is used in `collapseBasedirToPlaceholder` the version with the trailing slash would not match.

frerich:
Thanks - yes, it's very much appreciated if the commits themselves explain the reasoning behind the change. This is very useful when debugging things later on, where you typically check the history to see why some changes were done -- and it might be in a time or place where GitHub is not even available, so all the things discussed here are inaccessible.

I very much recommend [this page](http://chris.beams.io/posts/git-commit/) for some good guidelines on how to write commit logs, all future contributors (and your future self :-)) will be grateful! :-)

For what it's worth, it's very easy to update your commits with logs. With the command line client, you can run `git rebase --interactive <commit_on_which_you_based_your_branch>` and then change the action for each commit to perform (the word in the first column) to `reword`. Git will then replay all your commits, but on each commit it will pause and open an editor which allows you to edit the commit log.

clcache.py
siu: Indeed I applied this change as a quick test and didn't give it too much thought. With the proposed implementation the `assert path == os.path.normcase(path)` does not hold. I can add back the other assert `assert baseDir == os.path.normcase(baseDir)`

integrationtests.py
siu: This was changed only to simplify the code in TestBasedir. The cwd is changed with `cd()` and the relative path to the assets wouldn't work. If I rearrange the code there a bit this change is not needed.

siu:
Thanks for the clarification. I will amend the commits once the review is done.

clcache.py
frerich: I first wanted to insist on rather adjusting the test such that it appends a path separator (if needed) to the `os.getcwd()` return value instead of adjusting this function, but I guess if `os.getcwd()` does not return a trailing slash then that's a good indicator that we shouldn't do so either. So I'm okay with this. 👍

clcache.py
frerich: Why would the `assert path == os.path.normcase(path)` not hold anymore? Ah, I guess it's because this function is not given paths anymore, but command line arguments, e.g. `/IC:\Foo\Bar`. That suggests that either the argument should be renamed (`path` does not fit anymore), or maybe the newly introduced `collapseBasedirInCmd` should not pass arguments to `collapseBasedirToPlaceholder` but rather the actual paths.

To extract the paths from the arguments, the code in `CommandLineAnalyzer.parseArgumentsAndFiles` could maybe get refactored to be more reusable (it already solves the issue of getting the values for command line arguments).

I guess for now, simply renaming the `path` argument to e.g. `argument` or maybe even just `s` would be fine, at least it's not misleading anymore.

integrationtests.py
frerich: I see, makes sense! 👍

clcache.py
frerich: I first wanted to insist on rather adjusting the test such that it appends a path separator (if needed) to the os.getcwd() return value instead of adjusting this function, but I guess if os.getcwd() does not return a trailing slash then that's a good indicator that we shouldn't do so either. So I'm okay with this. 👍

siu:
@frerich I rewrote the PR with your suggestions.

 * Thanks for pointing to CommandLineAnalyzer.parseArgumentsAndFiles, it can be used from getManifestHash as it is and makes the changes in collapseBasedirToPlaceholder unnecessary.
 * The trailing separator still must be removed from normalizeBaseDir as the include arguments may not have it and it should match.
 * I kept the absolute ASSETS_DIR as it makes writing tests easier.

clcache.py
```diff
@@ -232,9 +232,21 @@ def getManifestHash(compilerBinary, commandLine, sourceFile):
         # One of the few exceptions to this rule is the /MP switch, which only
         # defines how many compiler processes are running simultaneusly.
         commandLine = [arg for arg in commandLine if not arg.startswith("/MP")]
+        arguments, inputFiles = CommandLineAnalyzer.parseArgumentsAndInputFiles(commandLine)
+        collapseBasedirInCmdPath = lambda path: collapseBasedirToPlaceholder(os.path.normcase(os.path.abspath(path)))
+
+        commandLineArgs = []
+        projectSpecificArgs = ("AI", "I", "FU")
+        for k in sorted(arguments.keys()):
```
frerich: Why do you use `sorted` here?

siu: Actually the sorted is very important to have repeatable manifest hashes. As dictionaries in python are unordered the only way to guarantee that we generate the same hash every invocation is to iterate over the keys in the same order. It is the same case that #235 solved.

frerich: Ah, so this basically improves the manifest cache hit rate, i.e. it's an improvement but not directly related to the BASEDIR feature? Ok.

siu: It is going to have that side effect but at same time it is needed for correctness since the implementation changed to use a dictionary.

frerich:
Thanks for being so patient and explaining all the changes to me, I don't see any reason not to merge this nice improvement - thanks a lot! :-)

siu:
I am happy to iterate and give explanations if things are not clear, the end result is better ;) Thanks for merging.

Issue #239, opened 2016-10-31
Performance improvement unexpectedly small


akleber:
After resolving my last issues I got back to do some performance tests on my main project and I see a surprisingly small performance improvement: I am building with ninja on a 24 core machine on a SSD. A build with empty cache takes about 48 min. The rebuild with a full cache takes about 30min. Here are the cache statistics for a cold and warm build.

```
clcache statistics:
  current cache dir         : D:\j2/clcache
  cache size                : 4,354,861,756 bytes
  maximum cache size        : 10,073,741,824 bytes
  cache entries             : 6476
  cache hits                : 6394
  cache misses
    total                      : 6476
    evicted                    : 0
    header changed             : 40
    source changed             : 6436
  passed to real compiler
    called w/ invalid argument : 0
    called for preprocessing   : 762
    called for linking         : 8
    called for external debug  : 22
    called w/o source          : 0
    called w/ multiple sources : 0
    called w/ PCH              : 0
```

I profiled both runs: [prof4-cold.txt](https://github.com/frerich/clcache/files/561879/prof4-cold.txt) and [prof4-warm.txt](https://github.com/frerich/clcache/files/561881/prof4-warm.txt)
If I read the 'warm' one correctly, most of the time is spend in computing file hashes, over 2 million times. Are these all the headers from all files? If this is correct, what I do not understand then, why the 'warm' case has a small amount of more calls to getFileHash() and why it takes more than double the time.
Any ideas or input how I might improve clcaches performance improvements?

FYI: My still unfinished lockfile mode does not change these numbers significantly. Also using xcopy instead of shutil.copyfile() or using readinto() instead of read() in getFileHash() does not change the situation.

frerich:
Thanks for your thoughtful post! I agree with your initial analysis of the 'warm' profile: most of the time is spent in computing the hashes of about 2.1 million include files. The numbers you got match those you mentioned in an [earlier comment of yours](https://github.com/frerich/clcache/issues/229#issuecomment-249560531) -- after working around what seems to be an inefficient usage of threads in the `subprocess` module, a large proportion of the time is spent in the `getFileHash` function.

[...]

PR #240, opened 2016-11-01
Introduce memoization to getFileHash

frerich:
Analysis of profile reports indicates that a large proportion of the
time is spent in getFileHash(), and chances are that this is because
getFileHash() is called repeatedly for the same files. Let's test that
theory by introducing a cache.


frerich:
This didn't really help with anything, certainly not the suboptimal performance with warm caches reported in #239, so closing this.

PR #241, opened 2016-11-02
Very simple fast test of hashing performance for moderate sized files

inorton:
I noticed the conversation about hashing in one of the other issues. Perhaps this would be useful?

It creates a directory tree of 1000 files each of 256K in size and then simply hashes each.  As files are fresh each run the misleading figures due to warm/cold disc caches should be consistent.  The whole test only takes a 1-2 sec for me.

### DIFF
--- a/hashingtests.py
+++ b/hashingtests.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+#
+# This file is part of the clcache project.
+#
+# The contents of this file are subject to the BSD 3-Clause License, the
+# full text of which is available in the accompanying LICENSE file at the
+# root directory of this project.
+#
+
+import os
+import unittest
+import shutil
+import clcache
+import uuid
+
+THIS_DIR = os.path.dirname(__file__)
+ASSETS_DIR = os.path.join(THIS_DIR, "tests", "hashtests")
+
+
+class TestFileHashing(unittest.TestCase):
+    """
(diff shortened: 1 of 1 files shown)


frerich:
Thanks, I think it's a good idea to have some sort of standard benchmark. I suppose instead of creating a new file, this could be part of the existing `performancetests.py` test case?

Alas (?), the discussion in #239 suggests that the issue with slow hashing appears to be related to *concurrent* clcache instances, at least for @akleber 's setup -- so I'm not sure the test code as it is reproduces that issue

In any case, I very much agree that some sort of performance test for this functionality would be good -- it's not clear to me though which scenarios to benchmark.

inorton:
I started this with a theory that we could compute hashes in parallel using concurrent.futures if we had multiple cpus (some doing IO some doing hashing).  My tests here showed that it actually just made things worse by quite some considerable margin.

frerich:
Indeed, it matches @akleber 's observation that _concurrent_ hashing of files is substantially slower than _sequential_ hashing.

Maybe this is another argument in favor of some sort of server process which acts as the sole instance to sequentially hash (and potentially cache) hashes.

frerich:
I think a performance test to check how fast cache hits and cache misses are (both concurrently as well as sequentially) would be a nice thing to have, but that should probably go into `performancetests.py`.

PR #242, opened 2016-11-11
Memcache support (PoC), based on fe83618 

hubx:
Hi guys,

As mentioned in #78 I have been working on memcache support for clache based on fe83618. However I was not able to submit this in time and the project moved significantly in the meantime. Before I reimpliment this feature once again I would like to collect your feedback: What to you think about the general approach and intrusiveness of the feature? Would you require a fallback to disk based cache before considering this PR? 

/cc @heremaps/cci

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -104,8 +104,8 @@ CLCACHE_BASEDIR::
     move your project to different directory, clcache will produce cache hits as
     before.
 CLCACHE_OBJECT_CACHE_TIMEOUT_MS::
-    Overrides the default ObjectCacheLock timeout (Default is 10 * 1000 ms).
-    The ObjectCacheLock is used to give exclusive access to the cache, which is
+    Overrides the default ObjectCacheFileLock timeout (Default is 10 * 1000 ms).
+    The ObjectCacheFileLock is used to give exclusive access to the cache, which is
     used by the clcache script. You may override this variable if you are
     getting ObjectCacheLockExceptions with return code 258 (which is the
     WAIT_TIMEOUT return code).
--- a/clcache.py
+++ b/clcache.py
@@ -298,9 +298,67 @@ class ObjectCacheFileStrategy(object):
         return ''
 
 
+class ObjectCacheMemcacheStrategy(object):
+    MANIFEST_PREFIX = "manifests_"
+    OBJECT_PREIFX = "objects_"
+
+    def __init__(self, server):
+        self.lock = ObjectCacheDummyLock()
+        self.local_cache = {}
+        from memcache import Client
+
+        self.mc = Client(servers=[server])
+
+    def getManifest(self, manifestHash):
+        return self.mc.get(self.MANIFEST_PREFIX + manifestHash)
+
+    def setManifest(self, manifestHash, manifest):
+        return self.mc.set(self.MANIFEST_PREFIX + manifestHash, manifest)
+
+    def fetchEntry(self, key):  #XX todo return data
+        data = self.mc.get(self.OBJECT_PREIFX + key)
+        if data is not None:
+            self.local_cache[key] = data
+            return True
+        else:
+            self.local_cache[key] = None
+            return None
+
+    def hasEntry(self, key):
+        return self.fetchEntry(key)
+
+    def getEntry(self, key, outputFile):
+        if key not in self.local_cache:
+            self.fetchEntry(key)
+        if self.local_cache[key] is None:
+            raise ValueError  # a fallback to FileCache can be implemented here or by handling this exception
+        data = self.local_cache[key]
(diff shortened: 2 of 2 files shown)


hubx:
Okay looks like the team notification doesn't work // cc @sschuberth @Jimilian directly

frerich:
Thanks a lot for the PR! It's indeed a fairly hefty change, but I greatly appreciate you approach of first gradually refactoring the code base in order to introduce the notion of 'strategies' and then only actually adding memcached support in the very last step. That made things manageable.

I don't think a fallback would be *required*; instead, I imagine something like a `CLCACHE_BACKEND` or `CLCACHE_STORAGE` environment variable which can assume values like `filesystem` or `memcache` or maybe something like `postgresql` would be sufficient.

I have no experience with memcache myself though, so I'd very much need tests which exercise the functionality. Other than that, I see no reason not to include this assuming you can update the PR for the current code base.

Can you maybe also provide a writeup about what Memcache is, how to set it up and how clcache benefits from using it as the backend (and when it might be interesting for users to use it)? That might be a nice addition to the Wiki, I imagine.

frerich:
After having read up on memcached a bit, I wonder: is it really a replacement for the file system based storage? How does it ensure persistence? It seems that it's more for "short-term" caching of data?

hubx:
So my short summary for memcached is: It stores hash =>object pairs in RAM. It manages concurrent request, but also cache invalidation for you (in contrast to postgresql). If objects are hashed consistently memcache can be distributed horizontally into multiple memcached servers. 

The real benefit is that multiple build nodes can access the same synchronized cache. Ideally its enough to compute a cached object file once in you CI system.

But as you mentioned this comes with a trade-off AFAIK there is no persistence in memcache. If you restart memcache, your cache is cleared. This why many applications (e.g. [ccache/ccache#58](https://github.com/ccache/ccache/pull/58#issuecomment-221005073)) employs a mixed mode where they first query the faster cache. In ccache linked PR's you can see that ccache does  `ccache -> query local_storage - not found? -> query memcache -> save to local_storage -> ccache output`. This is what I meant to ask with "Would you require a fallback to disk based cache before considering this PR?" - I see now how this can misinterpreted as *one decision* during clcache startup, but what I meant to ask would you require this mixed mode in cache lookup/saving or is memcache only fine for now?

Jimilian:
Small note: if something supports Memcache as a storage, it also supports a lot of other stuff (with persistency/replication/clustering/etc.). 
For example, in our setup `ccache` connects to `moxi` (proxy with persistent connections) and `moxi` is connected to memcache-friendly cluster in Amazon - `ElastiCache`.  Huge amount of storage systems support memcache-friendly API (i.e. [MySQL](https://dev.mysql.com/doc/ndbapi/en/ndbmemcache-overview.html) and [PostgreSQL with plugin](https://github.com/Ohmu/pgmemcache), [CouchBase](https://en.wikipedia.org/wiki/Couchbase_Server).

So, in this PR @hubx adds support for memcache-protocol.

frerich:
@hubx Thanks for the explanations, I think I'm slowly getting a grasp of it :-)

I see that the PR as it is does not implement the 'fallback' logic, and the strategy of using memcache only as a fallback suggests that querying memcache is slower than querying a local cache (but faster than recompiling from scratch), so I wonder: doesn't the patch proposed in this PR make cache misses less likely (by sharing a cache across multiple machines) at the expensive of making any cache hit slower (since querying memcache is slower than the local disk)?

My main concern with this really is that I personally have no use case, and no test infrastructure -- so the code would effectively be unmaintained and whenever I have to touch it, I wouldn't know how to verify its effect. So at least some sort of test suite would be much appreciated. :-)

hubx:
>  doesn't the patch proposed in this PR make cache misses less likely (by sharing a cache across multiple machines) at the expensive of making any cache hit slower (since querying memcache is slower than the local disk)

For serial cache hits this is true. If you have more workers e.g. with `msbuild /maxcpucount n+1` other nodes can use CPU while some wait for I/O. But the mixed mode definitely reduces unnecessary network load on a central memcached instance. 

A local memcache (in Linux VM) as cache was faster this implementation than using the local disk (but for our code base, we also need higher timeouts like #229). Some the mentioned protocol compatible alternatives, still are worthwhile trying.

About the test suites, sure. I think we could even rerun the full integration test suite with a memcache backend if need be (start a memcache server in the integration test runner). I'm not 100% sure about this, because there is some code like removing objects which would be solely implemented for testing purposes. In production memcache handles cache invalidation itself (in contrast the a file backend) by least recently used principle. So we might need to untangle the storage tests (e.g. `TestClearing) there and run only a subset of these against the memcache backend.

hubx:
superseded by #268

PR #244, opened 2016-11-14
Optimistic locking to improve concurrent cache misses

frerich:
This PR attempts to address the issue observed in issue #229 . The theory is that lock timeouts were triggered because -- in direct mode -- clcache holds *two* locks while invoking the real compiler, one manifest section and one artifact section lock. So despite the fact that there are 256 of each category, if a compiler execution takes sufficiently long, it's very well possible that a concurrent clcache invocation tries to access the same manifest *or* the same artifact section -- but times out.

The patch supplied by this PR attempts to address this by implementing a more optimistic locking strategy: in case a cache miss is detected, clcache releases all locks, rebuilds, and then acquires all locks again and checks whether the cache still needs updating. This means that there's a miniscule chance that clcache will compile the same object twice (in case two concurrent invocations are exactly equivalent), but it probably optimizes the common code of concurrent invocations being independent.

This behaviour implemented by basically inlining a lot of code in `processDirect` and `processNoDirect` and then reshuffling things such that the locks are not acquired while calling the real compiler.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1602,33 +1602,32 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
     manifestSection = cache.manifestRepository.section(manifestHash)
     with manifestSection.lock:
         manifest = manifestSection.getManifest(manifestHash)
-        if manifest is None:
+        if manifest:
+            for entryIndex, entry in enumerate(manifest.entries()):
+                # NOTE: command line options already included in hash for manifest name
+                try:
+                    includesContentHash = ManifestRepository.getIncludesContentHashForFiles(
+                        [expandBasedirPlaceholder(path) for path in entry.includeFiles])
+
+                    if entry.includesContentHash == includesContentHash:
+                        cachekey = entry.objectHash
+                        assert cachekey is not None
+                        # Move manifest entry to the top of the entries in the manifest
+                        manifest.touchEntry(entryIndex)
+                        manifestSection.setManifest(manifestHash, manifest)
+
+                        return getOrSetArtifacts(
+                            cache, cachekey, objectFile, compiler, cmdLine, Statistics.registerEvictedMiss)
+                except IncludeNotFoundException:
+                    pass
+
             return postprocessUnusableManifestMiss(
                 cache, objectFile, manifestSection, manifestHash, sourceFile, compiler, cmdLine,
-                Statistics.registerSourceChangedMiss)
-
-        for entryIndex, entry in enumerate(manifest.entries()):
-            # NOTE: command line options already included in hash for manifest name
-            try:
-                includesContentHash = ManifestRepository.getIncludesContentHashForFiles(
-                    [expandBasedirPlaceholder(path) for path in entry.includeFiles])
-
-                if entry.includesContentHash == includesContentHash:
-                    cachekey = entry.objectHash
-                    assert cachekey is not None
-                    # Move manifest entry to the top of the entries in the manifest
-                    manifest.touchEntry(entryIndex)
-                    manifestSection.setManifest(manifestHash, manifest)
-
-                    return getOrSetArtifacts(
-                        cache, cachekey, objectFile, compiler, cmdLine, Statistics.registerEvictedMiss)
-            except IncludeNotFoundException:
-                pass
+                Statistics.registerHeaderChangedMiss)
 
         return postprocessUnusableManifestMiss(
             cache, objectFile, manifestSection, manifestHash, sourceFile, compiler, cmdLine,
-            Statistics.registerHeaderChangedMiss)
-
+            Statistics.registerSourceChangedMiss)
 
 def processNoDirect(cache, objectFile, compiler, cmdLine, environment):
     cachekey = CompilerArtifactsRepository.computeKeyNodirect(compiler, cmdLine, environment)
--- a/clcache.py
+++ b/clcache.py
@@ -1644,11 +1644,12 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
     with manifestSection.lock:
         if artifactSection is not None:
             with artifactSection.lock:
-                with cache.statistics.lock, cache.statistics as stats:
-                    Statistics.registerEvictedMiss(stats)
-                    if returnCode == 0 and os.path.exists(objectFile):
-                        artifacts = CompilerArtifacts(objectFile, compilerOutput, compilerStderr)
-                        cleanupRequired = addObjectToCache(stats, cache, artifactSection, cachekey, artifacts)
+                if not artifactSection.hasEntry(cachekey):
+                    with cache.statistics.lock, cache.statistics as stats:
+                        Statistics.registerEvictedMiss(stats)
+                        if returnCode == 0 and os.path.exists(objectFile):
+                            artifacts = CompilerArtifacts(objectFile, compilerOutput, compilerStderr)
+                            cleanupRequired = addObjectToCache(stats, cache, artifactSection, cachekey, artifacts)
 
             return compilerResult + (cleanupRequired,)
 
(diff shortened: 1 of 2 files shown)


siu:
This helps in my use case. The version on master requires a huge timeout (600000 ms), with these changes it compiles successfully with the default (10000ms). 👍 

I will have a look at the code too.

clcache.py
```diff
@@ -1600,57 +1574,89 @@ def processCompileRequest(cache, compiler, args):
...
         manifest = manifestSection.getManifest(manifestHash)
-        if manifest is None:
-            return postprocessUnusableManifestMiss(
-                cache, objectFile, manifestSection, manifestHash, sourceFile, compiler, cmdLine,
-                Statistics.registerSourceChangedMiss)
+        if manifest:
+            for entryIndex, entry in enumerate(manifest.entries()):
+                # NOTE: command line options already included in hash for manifest name
+                try:
+                    includesContentHash = ManifestRepository.getIncludesContentHashForFiles(
```
siu: It would be interesting to move this computation out of the critical section as it usually shows in the profiling as one of the hotspots. It can be done by moving the if outside of the `with manifestSection.lock` but will complicate the code around touchEntry and setManifest.

clcache.py
```diff
@@ -1600,57 +1574,89 @@ def processCompileRequest(cache, compiler, args):
...
-                    assert cachekey is not None
-                    # Move manifest entry to the top of the entries in the manifest
-                    manifest.touchEntry(entryIndex)
-                    manifestSection.setManifest(manifestHash, manifest)
+    with manifestSection.lock:
+        if artifactSection is not None:
+            return ensureArtifactsExist(cache, artifactSection, cachekey, unusableManifestMissReason,
+                                        objectFile, compilerResult)
+
+        entry = createManifestEntry(manifestHash, includePaths)
```
siu: This line computes the hashes for all the includes and may be slow too. It would be nice to move it outside of the critical section too.

siu:
Here is the result of a run with CLCACHE_PROFILE enabled: 
[coldcache.zip](https://github.com/frerich/clcache/files/595504/coldcache.zip)

inorton:
Nice idea.  About your thought that it "might compile the same thing twice".  I doubt that should be a worry since even without clcache invoking cl on the same files with the same settings and arguments would cause one to fail anyway.

clcache.py
frerich: Dito.

frerich:
@siu Thanks a lot for your thorough review, despite the fact that this patch is not exactly trivial.

Also, thanks a lot for actually trying it on a real project and providing the `CLCACHE_PROFILE` output. I was mostly worried that I somehow missed a case where two cache updates might class, but that appears to not be the case? Alas these concurrency related tickets are very hard for me to debug since the strongest Windows machine I have access to has only two cores, so there's only very limited 'real' concurrency.

The profile you uploaded was quite interesting:

* It see that 1214 compile requests were processed over 9904 seconds.
* 9461 seconds (95.5%) of the time was spent in `WaitForSingleObject`, and the exclusive caller of that function is `subprocess.wait`, i.e. 95.5% of the time was spent waiting for the compiler. I wonder whether this allows claiming that clcache currently incurs ~5% overhead for cache misses.
* 1.9% of the time was spent in `createManifestEntry`, which is dominated by the calls to `getFileHash`

My impression is that there is room for improvement -- but not much, at least not for cache misses. I think it's worthwhile to start working on a background process which takes care of hashing files and memorising those hashes; the cache misses would probably benefit a little bit, but it would also be a major improvement to cache hits.

clcache.py
```diff
@@ -1600,57 +1574,89 @@ def processCompileRequest(cache, compiler, args):
...
+        cachekey = entry.objectHash
 
-                    return getOrSetArtifacts(
-                        cache, cachekey, objectFile, compiler, cmdLine, Statistics.registerEvictedMiss)
-            except IncludeNotFoundException:
-                pass
+        def addManifest():
+            manifest = createOrUpdateManifest(manifestSection, manifestHash, entry)
+            manifestSection.setManifest(manifestHash, manifest)
```
siu: Yes, it is a trivial change and can be done afterwards.

siu:
Note that `WaitForSingleObject` is not only called from subprocess.wait, it is the call executed to acquire the locks too. In this case though that entry in the profile seems to be dominated by the subprocess call.

In fact I tried applying some the changes I was suggesting locally and I could not see any relevant change in the profiler. The time spent reading manifests and deciding what to do is negligible compared to the time it takes to compile and it didn't help with a hot cache either.

I think we should profile other use cases before drawing any conclusion. Maybe a mixed execution with only 50% of the files in the cache will lead to more cases of manifest or artifact section clashes and could benefit more from reducing the time that the manifest section is locked. Anyway all this can be done later, for the moment it would just add complexity.

I had a second look at the code. It looks safe in my opinion.

clcache.py
```diff
@@ -1600,57 +1574,89 @@ def processCompileRequest(cache, compiler, args):
...
+
 
-    return compilerResult + (cleanupRequired,)
+def ensureArtifactsExist(cache, section, cachekey, reason, objectFile, compilerResult, extraCallable=None):
+    cleanupRequired = False
+    returnCode, compilerOutput, compilerStderr = compilerResult
+    with section.lock:
+        if not section.hasEntry(cachekey):
+            with cache.statistics.lock, cache.statistics as stats:
+                reason(stats)
```
siu: This is very subtle: even if the object was added in the cache by another process it makes sense to log the reason, after all it already did the work. I think that is how it was in the original code.

frerich: I'm not sure about this - in the old code, the statistics would only get updated with a cache miss if there was no entry in the cache yet. For instance, two concurrent clcache invocations which both produce a cache miss for the same same object would only cause one object to get added to the cache, and the stats only to get bumped once. I.e. the number of added entries would match the number of logged cache misses. This is because clcache would first lock the section and then test whether there's an entry yet.

In that sense, the proposed patch is faithful to the previous behaviour. However, I tend to agree that maybe this is not what we actually want. I'm not sure, it requires more consideration.

frerich:
It's true that `WaitForSingleObject` is used in multiple places. However, in the profile report you generated, I see these lines further down in the report (in the section which depicts from where each function was called):

```
Function                                                        was called by...
                                                                    ncalls  tottime  cumtime
[...]
subprocess.py:1154(wait)                                        <-    1214    0.078 9460.795  clcache.py:1159(invokeRealCompiler)
{built-in method WaitForSingleObject}                           <-    1214 9460.709 9460.709  subprocess.py:1154(wait)
```

I.e. in this case, it appears that the *only* caller of `WaitForSingleObject` was `subprocess.py:1154(wait)`. And the only caller of that was `invokeRealCompiler`. Quite frankly, it's not clear why the `WaitForSingleObject` calls done by `CacheLock.acquire` don't show up. In fact, there's no mentioning of the `CacheLock` calls at all. Maybe they were just so fast that the profiler decided to not list them? Strange...

siu:
Ah, yes, that is true. The missing calls in the profile may be a limitation of cProfile?

frerich:
Glad to hear that there's mostly positive feedback so far. I'm curious how it works for @akleber who reported #229 in the first place. :-)

clcache.py
siu: Oh, I understand now, we are looking at different parts of the original code.

In the original the code is duplicated around https://github.com/frerich/clcache/pull/244/files#diff-c09a754611b438b1dd2362d7cc35c53dL1447 and https://github.com/frerich/clcache/pull/244/files#diff-c09a754611b438b1dd2362d7cc35c53dL1647 and the two cases behave differently.

But it is true, none of the two approaches is right, one fails to update the reason counters and the other bumps the entry and size counters wrongly. An improvement would be to increase the reason counters independently of the entry counters. If we have to pick one of the wrong approaches I think it is better to keep it as in this PR.

frerich: Indeed, there is some inconsistency - good catch! I agree that for now, it's plausible to keep the PR as it is -- but this might be something one might want to address in the future...e

akleber:
Thank you very much @frerich for this PR which I am eager to test but unfortunately I will not be able to test it before 28.11.

frerich:
@akleber Sure, no rush. For what it's worth, I also just submitted #248 which hopefully is an effective approach to tackling #239 :-)

ivaigult:
Before this changes I wasn't able to compile our project due to the timeout.  I played around with CLCACHE_OBJECT_CACHE_TIMEOUT_MS but it didn't help at all. With this changes everything works fine and I've got 4x build speedup :-). Thank you so much!

sasobadovinac:
I can also confirm good results with this and the latest master for both #155 and #229 https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.377

akleber:
@frerich This PR solves my issue reported in #229. It did not improve the overall build time by a significant amount, but it gets rid of the time out. So a big thumbs up for this PR. Thanks. Looking into #248 now...

akleber:
@frerich I am sorry to report that I messed up my tests with this PR. I by error did not use a clcache version with this change. Sorry about that. I re-run my tests and still get a timeout. However this PR improves my situation as my build passes with a timeout of "only" 20s instead of the 60s I previously had to use. I did not have time to analyze what the timeout situation is yet, but I will look into it.

frerich:
@akleber No worries, it's of course perfectly possible that there are still other scenarios in which timeouts are triggered unexpectedly. The reports from you and @TiloW in #229 suggest that. I think there is some consensus that this PR implemented a more gentle locking strategy, so it moves things into the right direction. For other issues, we can create separate issues.

PR #249, opened 2016-11-21
Minor code deduplication in Cache.removeObjects()

frerich:
We already have a CacheSection here, so we can ask that for the object
file name.


frerich:
The proposed change was obsoleted by commit 49a96f80f103150482a26e84a3133d8084b9137d which removed the `removeObjects` method completely.

Issue #232, opened 2016-10-04
Executable generated with pyinstaller is slower than cx_freeze


siu:
We found that cx_freeze generates faster executables.

For example, with a hot cache of 1248 targets:
- executable generated with "pyinstaller --onefile clcache.py" takes 99 seconds
- executable generated with "C:\Python34\Scripts\cxfreeze clcache.py" takes 56 seconds

The downside is that cx_freeze cannot generate single file executables but the speedup is well worth it.

frerich:
That's very surprising - but very interesting! I'd have expected the performance to be exactly the same.

Can you please generate a build with both PyInstaller as well as cx_freeze and then perform a rebuild with a hot cache while having `CLCACHE_PROFILE=1` set? Afterwards, please run `showprofilereport.py` in the build directory to generate a profile report and safe that.

This should explain why there's such a huge difference in execution speed.

For what it's worth, there's also [Nuitka](http://nuitka.net/) which seems interesting. I wonder how that works for you (never tried it myself).

siu:
Here are the results of the profiling, direct mode with 268 targets:

[cx_freeze.txt](https://github.com/frerich/clcache/files/510390/cx_freeze.txt) took 8.62s
[pyinstaller.txt](https://github.com/frerich/clcache/files/510389/pyinstaller.txt) took 17.52s

An interesting fact is that the python code executes faster with pyinstaller. It would also make sense that the start up time dominates the timings for clcache.

frerich:
Thanks, interesting results!
You mention that the cx_freeze build took 8.62 seconds and the one with PyInstaller took 17.52 seconds. In `cx_freeze.txt` however, it says 49.663 seconds and in `pyinstaller.txt` it says  40.581 seconds.

Did you maybe mix up the file names? Based on the log files, it seems that the cx_freeze build is actually slower. Interestingly, the time difference is about the same as what you mention (9 seconds) but the absolute values are much larger. Maybe you are measuring CPU time whereas the profiling output is measuring wall time?

siu:
I am quite sure I didn't mix with the names and it matches with my timings. I can explain a bit more: I am using ninja with 12 jobs in parallel, I am measuring wall clock time and the profiler measures CPU time.

I think the difference in timings between the profiler and the wall time relies in the startup time. Maybe pyinstaller has more overhead before it starts executing the python code?

frerich:
[...]

PR #255, opened 2016-12-06
Use multi threading instead of re-invoking clcache

TiloW:
This idea was coined when we oberserved [rather large increases in build time when comparing `clcache` on cold caches to compiling without `clcache`](https://github.com/frerich/clcache/issues/239#issuecomment-264699884).

One thing supposedly contributing to this increase is that `clcache` re-invokes itself for every source file to be compiled. This is particularly costly when using `pyinstaller`'s `--onefile` flag, as reported in #232.

This PR proposes to switch from re-invoking `clcache` to multi-threading. Some changes had to be made to where output is performed and the command line analyzer was modified to return a list of object files when given multiple source files (this is why the unit tests did change).

Initially, I wanted to also invoke the real compiler at most once per invokation of `clcache` but this is not contained in this PR.

Integration tests failed on my machine even without changing anything on the current master. I did not bother adjusting them. If this is an error on my side, please let me know.

## What does this change do performance-wise?

The following times are obtained on a machine operating on windows 10 where everything takes place on the same SSD and builds are performed in parallel using 4 Cores @ 3.3 GHz.
We are using `CLCACHE_HARDLINK`.

| clcache | this PR | `--onefile` | warm cache | build time |
|---------|------|-----|---------|--------------|
| 0           |-| - |-         | 2:10       |
| 1           |0| 0|0          | 3:40       |
| 1           |0| 0|1          |  1:10     |
| 1           |0| 1|0          | 2:57       |
| 1           |0| 1|1          |  0:32     |
| 1           |1| 0|0          | 2:27       |
| 1           |1| 0|1          |  0:17     |

Edit: I just realized that this table is somewhat confusing: `--onefile = 1` means that I've used `pyinstaller` *without* the flag.

Note that with these changes the compiler is invoked only twice during our build. Thus, it does no longer matter to us whether `--onefile` is used. Apart from multithreading being faster, this would clearly be the more convenient solution for users that do invoke the compiler with many source files (e.g., `msbuild` invokes the compiler once for every target).

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -11,6 +11,7 @@ from ctypes import windll, wintypes
 from shutil import copyfile, rmtree
 import cProfile
 import codecs
+import concurrent.futures
 import contextlib
 import errno
 import hashlib
@@ -1268,56 +1269,6 @@ def jobCount(cmdLine):
         # not expected to happen
         return 2
 
-
-# Run commands, up to j concurrently.
-# Aborts on first failure and returns the first non-zero exit code.
-def runJobs(commands, environment, j=1):
-    running = []
-
-    while len(commands):
-
-        while len(running) >= j:
-            thiscode = waitForAnyProcess(running).returncode
-            if thiscode != 0:
-                return thiscode
-
-        thiscmd = commands.pop(0)
-        running.append(subprocess.Popen(thiscmd, env=environment))
-
-    while len(running) > 0:
-        thiscode = waitForAnyProcess(running).returncode
-        if thiscode != 0:
-            return thiscode
-
-    return 0
-
-
-# re-invoke clcache.py once per source file.
-# Used when called via nmake 'batch mode'.
-# Returns the first non-zero exit code encountered, or 0 if all jobs succeed.
-def reinvokePerSourceFile(cmdLine, sourceFiles, environment):
-    printTraceStatement("Will reinvoke self for: {}".format(sourceFiles))
-    commands = []
-    for sourceFile in sourceFiles:
-        # The child command consists of clcache.py ...
-        newCmdLine = [sys.executable]
-        if not hasattr(sys, "frozen"):
-            newCmdLine.append(sys.argv[0])
-
-        for arg in cmdLine:
-            # and the current source file ...
-            if arg == sourceFile:
-                newCmdLine.append(arg)
-            # and all other arguments which are not a source file
-            elif arg not in sourceFiles:
-                newCmdLine.append(arg)
-
-        printTraceStatement("Child: {}".format(newCmdLine))
-        commands.append(newCmdLine)
-
-    return runJobs(commands, environment, jobCount(cmdLine))
-
-
 def printStatistics(cache):
     template = """
 clcache statistics:
@@ -1548,25 +1499,28 @@ def processCompileRequest(cache, compiler, args):
     try:
         sourceFiles, objectFiles = CommandLineAnalyzer.analyze(cmdLine)
 
-        if len(sourceFiles) > 1:
-            return reinvokePerSourceFile(cmdLine, sourceFiles, environment), '', ''
-        else:
-            assert objectFile is not None
-            if 'CLCACHE_NODIRECT' in os.environ:
-                returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processNoDirect(cache, objectFiles[0], compiler, cmdLine, environment)
-            else:
-                returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processDirect(cache, objectFiles[0], compiler, cmdLine, sourceFiles[0])
-            printTraceStatement("Finished. Exit code {0:d}".format(returnCode))
+        baseCmdLine = []
+        for arg in cmdLine:
+            if arg not in sourceFiles:
+                baseCmdLine.append(arg)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=jobCount(cmdLine)) as executor:
+            jobs = []
+            for i, srcFile in enumerate(sourceFiles):
+                jobCmdLine = list(baseCmdLine)
+                jobCmdLine.append(srcFile)
+                jobs.append(executor.submit( \
+                        processActualCompileRequest, \
+                        cache, compiler, jobCmdLine, srcFile, objectFiles[i], environment))
+            for future in concurrent.futures.as_completed(jobs):
+                exitCode, out, err = future.result()
+                printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
+                printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
 
-            if cleanupRequired:
-                with cache.lock:
-                    cleanCache(cache)
+                if exitCode != 0:
+                    return exitCode
 
-            printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
-            printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
-            return returnCode
+        return 0
+
     except InvalidArgumentError:
         printTraceStatement("Cannot cache invocation as {}: invalid argument".format(cmdLine))
         updateCacheStatistics(cache, Statistics.registerCallWithInvalidArgument)
@@ -1590,15 +1544,32 @@ def processCompileRequest(cache, compiler, args):
     except CalledForPreprocessingError:
         printTraceStatement("Cannot cache invocation as {}: called for preprocessing".format(cmdLine))
         updateCacheStatistics(cache, Statistics.registerCallForPreprocessing)
-    except IncludeNotFoundException:
-        pass
 
-    exitCode, err, out = invokeRealCompiler(compiler, args[1:])
+    exitCode, out, err = invokeRealCompiler(compiler, args[1:])
     printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
     printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
     return exitCode
 
-    return returnCode, compilerOutput, compilerStderr
+def processActualCompileRequest(cache, compiler, cmdLine, sourceFile, objectFile, environment):
+    try:
+        assert objectFile is not None
+        if 'CLCACHE_NODIRECT' in os.environ:
+            returnCode, compilerOutput, compilerStderr, cleanupRequired = \
+            processNoDirect(cache, objectFile, compiler, cmdLine, environment)
+        else:
+            returnCode, compilerOutput, compilerStderr, cleanupRequired = \
+            processDirect(cache, objectFile, compiler, cmdLine, sourceFile)
+
+        printTraceStatement("Finished. Exit code {0:d}".format(returnCode))
+
+        if cleanupRequired:
+            with cache.lock:
+                cleanCache(cache)
+
+        return returnCode, compilerOutput, compilerStderr
+
+    except IncludeNotFoundException:
+        return invokeRealCompiler(compiler, cmdLine, environment=environment)
 
 def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
     manifestHash = ManifestRepository.getManifestHash(compiler, cmdLine, sourceFile)
--- a/clcache.py
+++ b/clcache.py
@@ -1526,10 +1526,7 @@ clcache.py v{}
     if "CLCACHE_DISABLE" in os.environ:
         return invokeRealCompiler(compiler, sys.argv[1:])[0]
     try:
-        exitCode, compilerStdout, compilerStderr = processCompileRequest(cache, compiler, sys.argv)
-        printBinary(sys.stdout, compilerStdout.encode(CL_DEFAULT_CODEC))
-        printBinary(sys.stderr, compilerStderr.encode(CL_DEFAULT_CODEC))
-        return exitCode
+        return processCompileRequest(cache, compiler, sys.argv)
     except LogicException as e:
         print(e)
         return 1
@@ -1566,7 +1563,9 @@ def processCompileRequest(cache, compiler, args):
                 with cache.lock:
                     cleanCache(cache)
 
-            return returnCode, compilerOutput, compilerStderr
+            printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
+            printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
+            return returnCode
     except InvalidArgumentError:
         printTraceStatement("Cannot cache invocation as {}: invalid argument".format(cmdLine))
         updateCacheStatistics(cache, Statistics.registerCallWithInvalidArgument)
@@ -1593,8 +1592,12 @@ def processCompileRequest(cache, compiler, args):
     except IncludeNotFoundException:
         pass
 
-    return invokeRealCompiler(compiler, args[1:])
+    exitCode, err, out = invokeRealCompiler(compiler, args[1:])
+    printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
+    printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
+    return exitCode
 
+    return returnCode, compilerOutput, compilerStderr
 
 def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
     manifestHash = ManifestRepository.getManifestHash(compiler, cmdLine, sourceFile)
--- a/clcache.py
+++ b/clcache.py
@@ -254,7 +254,6 @@ class ManifestRepository(object):
         # compiler processes are running simultaneusly.  Arguments that specify
         # the compiler where to find the source files are parsed to replace
         # ocurrences of CLCACHE_BASEDIR by a placeholder.
-        commandLine = [arg for arg in commandLine if not arg.startswith("/MP")]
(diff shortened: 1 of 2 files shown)


TiloW:
Integration tests seem to be failing. I'll have to see whether I can get them to execute on my local machine before fixing this.

clcache.py
```diff
@@ -69,6 +69,7 @@
 
 def printBinary(stream, rawData):
     stream.buffer.write(rawData)
+    stream.flush()
```
frerich: Why?

clcache.py
```diff
@@ -1547,7 +1544,9 @@ def processCompileRequest(cache, compiler, args):
                 with cache.lock:
                     cleanCache(cache)
 
-            return returnCode, compilerOutput, compilerStderr
+            printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
+            printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
+            return returnCode
```
frerich: This code path no longer returns a 3-tuple, is that intended?

clcache.py
```diff
@@ -1574,8 +1573,12 @@ def processCompileRequest(cache, compiler, args):
     except IncludeNotFoundException:
         pass
 
-    return invokeRealCompiler(compiler, args[1:])
+    exitCode, err, out = invokeRealCompiler(compiler, args[1:])
+    printBinary(sys.stderr, err.encode(CL_DEFAULT_CODEC))
+    printBinary(sys.stdout, out.encode(CL_DEFAULT_CODEC))
+    return exitCode
 
+    return returnCode, compilerOutput, compilerStderr
```
frerich: This second `return` statement is dead code now.

clcache.py
```diff
@@ -1507,10 +1507,7 @@ def main():
     if "CLCACHE_DISABLE" in os.environ:
         return invokeRealCompiler(compiler, sys.argv[1:])[0]
     try:
-        exitCode, compilerStdout, compilerStderr = processCompileRequest(cache, compiler, sys.argv)
-        printBinary(sys.stdout, compilerStdout.encode(CL_DEFAULT_CODEC))
-        printBinary(sys.stderr, compilerStderr.encode(CL_DEFAULT_CODEC))
-        return exitCode
+        return processCompileRequest(cache, compiler, sys.argv)
```
frerich: Why did you need to do this? `processCompileRequest` now duplicates the code for printing the output in two places.

clcache.py
```diff
@@ -1526,18 +1527,18 @@ def processCompileRequest(cache, compiler, args):
...
         else:
             assert objectFile is not None
             if 'CLCACHE_NODIRECT' in os.environ:
                 returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processNoDirect(cache, objectFile, compiler, cmdLine, environment)
+                    processNoDirect(cache, objectFiles[0], compiler, cmdLine, environment)
             else:
                 returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processDirect(cache, objectFile, compiler, cmdLine, sourceFiles[0])
+                    processDirect(cache, objectFiles[0], compiler, cmdLine, sourceFiles[0])
```
frerich: What is the benefit of doing this patch - it appears that the code always only uses the first list element? If this is a preparation for a subsequent patch, please explain how this helps implementing the next commit. As it is, the commit by itself does not seem very useful.

clcache.py
```diff
@@ -1529,25 +1480,25 @@ def processCompileRequest(cache, compiler, args):
...
+        baseCmdLine = []
+        for arg in cmdLine:
+            if arg not in sourceFiles:
+                baseCmdLine.append(arg)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=jobCount(cmdLine)) as executor:
+            jobs = []
+            for i,srcFile in enumerate(sourceFiles):
+                jobCmdLine = list(baseCmdLine)
+                jobCmdLine.append(srcFile)
+                jobs.append(executor.submit(processActualCompileRequest, cache, compiler, jobCmdLine, srcFile, objectFiles[i], environment))
```
frerich: Instead of scheduling a `processActualCompileRequest` here, which then uses most (but not all!) arguments depending on whether `CLCACHE_NODIRECT` is set, how about testing `CLCACHE_NODIRECT` just once up front here and then either scheduling a call to `processDirect` (passing the environment) or `processNoDirect` (passing the source file)?

Also, I guess it might work to factor out the final (optional) cache cleaning step so that it's performed just once at the end of all compiles?

sasobadovinac:
Here are two builds from me using this, I do see a bit faster builds, will try with no `--onefile` next...

https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.481
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.482

clcache.py
TiloW: I've placed this in a separate method. It's needed since we don't want to print the aggregated output of all compiler invokations in one go but want to print information as soon as it is available (i.e., we have processed the cache hit or the compiler invokation has returned).

clcache.py
TiloW: It now also guards the call in case an exception is thrown. This new `try`-`catch` needs to be added to both methods for the change you proposed. Otherwise, I'd think this would be a good idea (even though I don't think there would be any visible performance gain).

TiloW:
@frerich thanks for the review! Integration tests are working now.

clcache.py
frerich: I see -- it would be great if you could include this in the commit log (i.e. explain why you needed to do it, what would happen if this was *not* done) to simplify future debugging in case someone complains about e.g. `printBinary` being too slow because it flushes unnecessarily ;-)

clcache.py
frerich: I see, sounds plausible. It would be great to have this in the commit log :-)

frerich:
Great, thanks for getting the tests going! I'll have another careful look at the code later today. I'm a little scared by using multiple threads since nothing in the code base is prepared for that, e.g. I wonder whether there could be concurrent accesses to the Python dictionary holding the statistics or so. :-}

TiloW:
> I wonder whether there could be concurrent accesses

That is indeed something that should be double-checked. Are there any global variables?
We could perform deep copies before passing arguments to `processActualCompileRequest` to avoid concurrent access. I believe most values are only read and it would suffice to copy the `cache`-parameter.

TiloW:
By the way: [the error that just occured on the CI](https://ci.appveyor.com/project/frerich/clcache/build/1.0.952/job/an7d80dvhw8wf93v/tests) is also happening on master on my machine.

TiloW:
> Here are two builds from me using this, I do see a bit faster builds, but this was still with --onefile

Interesting, thanks for trying this. `--onefile` should not matter with this change unless you are invoking the compiler *many* times with few source files. Could you provide some numbers, i.e., how much faster exactly is this patch for you?
How often do you invoke the compiler duing your build and with how many files per invokation?
I would also be interested in build times for warm caches.

For example, during a typical build I have only two compiler invokations and about 360 source files in total.

frerich:
> Interesting, thanks for trying this. --onefile should not matter with this change unless you are invoking the compiler many times with few source files.

For the record, this is indeed one of the two common usage patterns:

* Some build systems (apparently most commonly those which are portable to non-Microsoft platforms) appear to follow a very traditional GNU Make style, in which dependency tracking and concurrency is not managed by the compiler but rather by the build system: e.g. a library depends on many individual object files, each of which depending on a single source file. Generating the object file means invoking the compiler with a single source file. In fact, the only portable build tool I'm aware of which exploits the concurrency built into Visual Studio is [qmake](http://doc.qt.io/qt-5/qmake-manual.html). For what it's worth, this is the scenario for my own personal use case which initially triggered the development of clcache. :-)

* Build systems which are more specialized on Windows seem to rely on the depend on the concurrency built into `cl.exe` a lot more, e.g. anything built on top of msbuild.

frerich:
> By the way: the error that just occured on the CI is also happening on master on my machine.

Interesting - I did not see that before. Can you reproduce it reliably? I wonder whether there's a race condition by `subprocess.check_output` returning earlier than expected, i.e. the `main.exe` executable is still in use somehow and thus it cannot be deleted. If so, this could maybe be reproduced with a sample loop copying some `.exe` file, then launching it via `check_output` and then deleting it.

clcache.py
```diff
@@ -1528,25 +1459,38 @@ def processCompileRequest(cache, compiler, args):
...
-                returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processNoDirect(cache, objectFile, compiler, cmdLine, environment)
-            else:
-                returnCode, compilerOutput, compilerStderr, cleanupRequired = \
-                    processDirect(cache, objectFile, compiler, cmdLine, sourceFiles[0])
-            printTraceStatement("Finished. Exit code {0:d}".format(returnCode))
+        baseCmdLine = []
+        for arg in cmdLine:
+            if arg not in sourceFiles:
+                baseCmdLine.append(arg)
```
frerich: Testing for membership in lists in lists [is O(n)](https://wiki.python.org/moin/TimeComplexity), so this loop scales with O(n^2) with respect to the number of source files on the command line.

However, a quick experiment suggests that this really only starts to hurt for numbers of source files which may not be realistic (for 1500 source files, the loop takes 77ms here, with 15000 source files it's 1567ms). It may not be worth it, but using a `set` would improve this to O(n). I'm not even sure whether you could plausibly invoke `cl.exe` with a few thousand source files, maybe via response files?

clcache.py
```diff
@@ -1528,25 +1459,38 @@ def processCompileRequest(cache, compiler, args):
...
+
+        exitCode = 0
+        cleanupRequired = False
+        with concurrent.futures.ThreadPoolExecutor(max_workers=jobCount(cmdLine)) as executor:
+            jobs = []
+            for i, srcFile in enumerate(sourceFiles):
+                jobCmdLine = list(baseCmdLine)
+                jobCmdLine.append(srcFile)
```
frerich: A shorter way to write this would be just

```python
jobCmdLine = baseComdLine + [srcFile]
```

clcache.py
```diff
@@ -1528,25 +1459,38 @@ def processCompileRequest(cache, compiler, args):
...
+                jobs.append(executor.submit( \
+                        processActualCompileRequest, \
+                        cache, compiler, jobCmdLine, srcFile, objectFiles[i], environment))
```
frerich: These backslashes are not needed I think.

clcache.py
```diff
@@ -1528,25 +1459,38 @@ def processCompileRequest(cache, compiler, args):
...
+        baseCmdLine = []
+        for arg in cmdLine:
+            if arg not in sourceFiles:
+                baseCmdLine.append(arg)
+
+        exitCode = 0
+        cleanupRequired = False
+        with concurrent.futures.ThreadPoolExecutor(max_workers=jobCount(cmdLine)) as executor:
+            jobs = []
+            for i, srcFile in enumerate(sourceFiles):
```
frerich: Instead of using `enumerate` here, you could use

```python
for srcFile, objectFile in zip(sourceFiles, objectFiles):
```

such that you don't need `objectFiles[i]` further down.

TiloW:
> I did not see that before. Can you reproduce it reliably?

Not really, it happens occasionally and doesn't seem to be tied to a specific invokation of `_clean()`.

TiloW:
> the global stdout stream. I don't know whether it can safely be used from multiple threads

Neither do I. 

> calls `sys.exit`

I've changed this to raising an exception.

> Did you already try how printBinary and print behave when invoked concurrently?

Not explicitly. My build log looks fine though.

Thanks for your remarks. It's been a while since I've last written something in python.

TiloW:
Regarding #248, for builds that invoke the compiler few times with many sources, it is very easy to cache the computed file hashes during that invokation with this change. I've tested this and experienced a minor speedup (~5s) that is independent of warm/cold caches. This decreases runtime for warm caches to `12s` for me.

There are still some things that I consider discussion-worthy regarind my changes:
 * `ThreadPoolExecutor` is used even for invokation with a single source file and for invokations without `/MP`. This might increase runtime in such scenarios (depending on whether it actually creates a new thread when there is no need for it).
 * The way the `jobCmdLine` is constructed, we do not preserve the order of arguments. I am not entirely sure whether this might cause any issues.
 * Tests for invoking the compiler with multiple sources and `/MP` should be added.

frerich:
> ThreadPoolExecutor is used even for invokations with a single source file and for invokations without /MP. This might increase runtime in such scenarios (depending on whether it actually creates a new thread when there is no need for it).

Right, this is something which simply should be benchmarked.

> The way the jobCmdLine is constructed, we do not preserve the order of arguments. I am not entirely sure whether this might cause any issues.

The only difference is that the `/MP` switch is omitted and the source file argument goes last, if I'm reading the code correctly? That should be fine.

> Tests for invoking the compiler with multiple sources and /MP should be added.

There are some integration tests already in `integrationtests.py`, namely `TestRunParallel` an `TestMultipleSources`. Maybe they would be good places to add any test cases you have in mind?

TiloW:
I've added a test for the command line output in #258. This PR should be merged afterwards.

TiloW:
> The only difference is that the `/MP` switch is omitted and the source file argument goes last, if I'm reading the code correctly? That should be fine.

Right, everything else remains the same.

> Right, this is something which simply should be benchmarked.

Will do.

TiloW:
I've got some more numbers for the current version of the PR.
As mentioned by @akleber, `ninja` calls the compiler with a single source file each.
Thus, comparing `ninja` builds with this PR to the master should give a good estimate of the runtime-impact of using threads even when not needing it. I've run the tests on a different (non SSD) drive.
The numbers are averaged across 3 builds each.

| system  | clcache | warm | this PR | time |
|---------|---------|------|---------|------|
| ninja   | 0       | -    | -       | 2:27 |
| msbuild | 0       | -    | -       | 2:12 |
| ninja   | 1       | 0    | 0       | 2:59 |
| msbuild | 1       | 0    | 0       | 2:51 |
| ninja   | 1       | 1    | 0       | 0:41 |
| msbuild | 1       | 1    | 0       | 0:34 |
| ninja   | 1       | 0    | 1       | 2:56 |
| msbuild | 1       | 0    | 1       | 2:28 |
| ninja   | 1       | 1    | 1       | 0:43 |
| msbuild | 1       | 1    | 1       | 0:22 |

`ninja` seems to be slower than `msbuild`, independently of `clcache`.
It is not clear from these numbers whether there is a runtime increase to be expected with this PR when using single file invocation. I've read that multithreading doesn't perform as well because of [CPython's global interpreter lock](https://wiki.python.org/moin/GlobalInterpreterLock). This is claimed to affect threads that are cpu bound more than threads limited by IO. While this would be in-line  with the above numbers, it's rather speculative.

frerich:
Thanks for the numbers!

> It is not clear from these numbers whether there is a runtime increase to be expected with this PR when using single file invocation.

Maybe it's not crystal clear, but I think the timings do allow some qualitative statement about this PR if you regroup them a bit. Here are the timings for cold caches (i.e. dominated by cache misses):

| system  | clcache | warm | this PR | time |
|---------|---------|------|---------|------|
| ninja   | 1       | 0    | 0       | 2:59 |
| msbuild | 1       | 0    | 0       | 2:51 |
| ninja   | 1       | 0    | 1       | 2:56 |
| msbuild | 1       | 0    | 1       | 2:28 |

In the case of cold caches, single-file invocations (represented by `ninja`) perform as fast as before. Certainly no worse. msbuild, passing multiple files, benefits by being about 14% faster -- which was the whole point of this PR.

Here are the timings for warm caches, i.e. mostly cache hits:

| system  | clcache | warm | this PR | time |
|---------|---------|------|---------|------|
| ninja   | 1       | 1    | 0       | 0:41 |
| msbuild | 1       | 1    | 0       | 0:34 |
| ninja   | 1       | 1    | 1       | 0:43 |
| msbuild | 1       | 1    | 1       | 0:22 |

Again, single-file invocations are basically unchanged (I think it's safe to say that two seconds are within the margin for measurement errors), but msbuild benefits - the build being a nice 36% faster.

Based on this, I think the PR looks really good and I really have to think hard whether I can think of compelling reasons not to merge it right away. It's really just the "The code was never written with multiple threads in mind" insight which makes me be very cautious. :-]

TiloW:
> I really have to think hard whether I can think of compelling reasons not to merge it right away

Now that #248 and #258 are merged a rebase is in order.
I would also like to have a closer look at the runtime-impact of the newly introduced locks and instantiation of `Cache`-objects for each thread. This could of course also be done in a subsequent PR.

> It's really just the "The code was never written with multiple threads in mind" insight which makes me be very cautious. 

:+1:

frerich:
> Now that #248 and #258 are merged a rebase is in order.

Really? The PR still seems to merge cleanly (i.e. not cause any conflicts) according to GitHub.

TiloW:
> The PR still seems to merge cleanly

It does but we don't know whether the tests introduced in #258 are passing with these changes.

TiloW:
I'd say this is ready to be merged. The issue described in #261 is not affected by this PR.

frerich:
I tend to agree, this looks good to me. Let's merge it and see how it goes.

Thanks a lot for your great work and constructive communication - would you mind me adding you to the `README` file in the section of who gets the blame (or fame?) for the program?

TiloW:
> would you mind me adding you to the README file in the section of who gets the blame (or fame?)

Sure, glad I could contribute. (:

PR #258, opened 2016-12-11
Integration tests

TiloW:
This reduces duplicated code in the integration tests and adds a tests for invoking the compiler with many source files in parallel. In particular, the command line output is tested.

### DIFF
--- a/integrationtests.py
+++ b/integrationtests.py
@@ -806,6 +806,17 @@ class TestRunParallel(unittest.TestCase):
                 self.assertEqual(stats.numCacheMisses(), 2)
                 self.assertEqual(stats.numCacheEntries(), 2)
 
+    def testOutput(self):
+        with cd(os.path.join(ASSETS_DIR, "parallel")), tempfile.TemporaryDirectory() as tempDir:
+            sources = glob.glob("*.cpp")
+            clcache.Cache(tempDir)
+            customEnv = dict(os.environ, CLCACHE_DIR=tempDir)
+            cmd = CLCACHE_CMD + ["/nologo", "/EHsc", "/c"]
+            mpFlag = "/MP" + str(len(sources))
+            out = subprocess.check_output(cmd + [mpFlag] + sources, env=customEnv).decode("ascii")
+
+            for s in sources:
+                self.assertEqual(out.count(s), 1)
 
 # Compiler calls with multiple sources files at once, e.g.
 # cl file1.c file2.c
--- a/integrationtests.py
+++ b/integrationtests.py
@@ -689,8 +689,7 @@ class TestHeaderMiss(unittest.TestCase):
             self.assertEqual(misses3, misses2)
             self.assertEqual(hits3, hits2+1)
 
-
-class TestRunParallel(unittest.TestCase):
+class RunParallelBase:
     def _zeroStats(self):
         subprocess.check_call(CLCACHE_CMD + ["-z"])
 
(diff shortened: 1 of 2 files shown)


integrationtests.py
```diff
@@ -578,25 +579,33 @@ def testSampleproject(self):
 
 class TestHeaderChange(unittest.TestCase):
     def _clean(self):
+        # It seems that subprocess.check_output() occansionally returns before
```
frerich: Typo here: "occansionally" => "occasionally".

integrationtests.py
```diff
@@ -578,25 +579,33 @@ def testSampleproject(self):
...
+        # windows fully releases the respective executable.
+        # This pause prevents failing tests because of missing permissions to remove the file.
+        time.sleep(.1)
+
```
frerich: This is an improvement, but it suffers from two downsides: .1 might still not be enough for some cases, so the test remains flaky (though less so), and now the tests *always* waits .1 no matter whether it's necessary or not.

A common alternative (retrying and ignoring the 'Access denied' error) would be more code though, so for now let's just go for this and see if/when it breaks.

integrationtests.py
```diff
@@ -719,11 +709,22 @@ def _buildAll(self):
...
             cxxflags = ["/c", "/nologo", "/EHsc"]
             cmd = CLCACHE_CMD + cxxflags + [sourceFile]
-            processes.append(subprocess.Popen(cmd))
+            processes.append(subprocess.Popen(cmd, env=self._createEnv()))
 
         for p in processes:
             p.wait()
 
+    def _enableNoDirect(self):
+        return False
```
frerich: The implementations of this method never return a value which changes during the objects lifetime, so how about just passing the flag in the constructor and use it via a member variable?

In fact, since the variable is only used in `_createEnv`, how about passing a "base environment" in the constructor (and dropping the flag completely) and only making `_createEnv` add the `CLCACHE_DIR` variable if needed?

integrationtests.py
```diff
@@ -806,6 +807,20 @@ def testHitsViaMpConcurrent(self):
                 self.assertEqual(stats.numCacheMisses(), 2)
                 self.assertEqual(stats.numCacheEntries(), 2)
 
+    def testOutput(self):
+        with cd(os.path.join(ASSETS_DIR, "parallel")), tempfile.TemporaryDirectory() as tempDir:
+            sources = [f for f in glob.glob("*.cpp")]
```
frerich: This line is the same as

    sources = list(glob.glob("*.cpp"))

...but for me, `glob.glob` already returns a list, so maybe you can just do

    sources = glob.glob("*.cpp")

integrationtests.py
```diff
@@ -707,8 +698,7 @@ def testObsoleteHeaderDisappears(self):
             self.assertEqual(misses3, misses2)
             self.assertEqual(hits3, hits2+1)
 
-
-class TestRunParallel(unittest.TestCase):
+class TestRunParallelBase:
```
frerich: Uh-oh... defining base classes purely for the sake of code reuse is a bit of a code smell, but let's see how this works out...

integrationtests.py
TiloW: I like the idea. But note that this would mean that there is only a single environment for the entire class. In particular, all tests will work on the same cache directory. In any case we should use the constructor to pass the nodirect-parameter.

integrationtests.py
```diff
@@ -606,34 +615,16 @@ def testDirect(self):
...
-                header.write("#define VERSION 2")
+    def testDirect(self):
+        self._performTest(False)
 
-            self._compileAndLink(testEnvironment)
-            cmdRun = [os.path.abspath("main.exe")]
-            output = subprocess.check_output(cmdRun).decode("ascii").strip()
-            self.assertEqual(output, "2")
+    def testNoDirect(self):
+        self._performTest(True)
```
TiloW: Good idea, will do.

frerich:
Merging this, the amount of code duplication removed is just too good to pass up. Thanks!

PR #259, opened 2016-12-13
adds the possibility to exclude directories from being watched.

akleber:
Watched directories can not be deleted and this might be problematic
in some environments.


server/clcachesrv.py
```diff
@@ -5,13 +5,16 @@
...
 import pickle
 import signal
+import argparse
+import re
 
 import pyuv
 
 class HashCache(object):
-    def __init__(self, loop):
+    def __init__(self, loop, args):
```
frerich: I think `args` is a bit too generic and way more than `HashCache` needs. How about just passing `excludedDirs` here?

server/clcachesrv.py
```diff
@@ -46,6 +49,16 @@ def _onPathChange(self, handle, filename, events, error):
...
 
+    def isExcluded(self, dirname):
+        # as long as we do not hate more than _MAXCACHE regex we can
+        # rely on the internal cacheing of re.match
+        for pattern in vars(self._args)['exclude']:
+            if re.search(pattern, dirname):
+                logging.debug("NOT watching %s", dirname)
+                return True
+
+        return False
```
frerich: You could shorten this to

```
excluded = any(re.search(pattern, dirname) for pattern in vars(self._args)['exclude']))
if excluded:
    logging.debug("NOT watching %s", dirname)
return excluded
```

...but two more important things also come to my mind:

1. Do you really mean to use `re.search` here? It will match the pattern *anywhere* in the dir, e.g. the pattern `\\build\\` would match `C:\build\mydata` as well as `C:\foo\bar\build\yoyodyne` etc.. Maybe `re.match` would be more like what one would expect?

2.  Windows paths are case insensitive, so I imagine that the regular expression should match in a case insensitive way, too -- it appears you can achieve this by passing `re.IGNORECASE` as the third argument to `re.search` (or `re.match`).

Having written this and noticing that backslashes are both very common for Windows paths *and* need escaping, I wonder whether regular expressions are maybe overkill. Maybe it would be sufficient to allow specifying prefixes to match against and then we just call `os.path.normcase` to normalise the paths and see whether one starts with another?

server/clcachesrv.py
```diff
@@ -46,6 +49,16 @@ def _onPathChange(self, handle, filename, events, error):
             logging.debug("invalidating cached hashsum for %s", os.path.join(handle.path, filename))
             del watchedDirectory[filename]
 
+    def isExcluded(self, dirname):
+        # as long as we do not hate more than _MAXCACHE regex we can
```
frerich: Typo here: "hate" => "have"

frerich:
@akleber With clcachesrv merged to master, it seems it would be good to rebase this PR (and maybe incorporate some of the suggestions and your discretion :-).

akleber:
@frerich I am going to rebase this PR and finish it up but it will unfortunately have to wait for another 2-3 weeks.

frerich:
@akleber Sounds good -- no hurry. Happy new year! :-)

akleber:
Seems I have done something wrong with this PR when I tried to rebase it. So I created a new one: #272

PR #260, opened 2016-12-14
Release manifest sections while computing manifest hashes

siu:
As discussed in #244, it should be possible to release the manifest sections while computing the manifest hash to allow higher concurrency. In this PR the manifest section is released while searching for a match in the manifest and when a new ManifestSection is created. It is later locked again before performing any operation on the artifacts repository to keep the correct order of locks. These two cases should cover all the expensive operations performed with a manifestSection.

In my tests this does not improve compilation times but I would like to document it here and I am curious if it would help for other people use-cases. For example it would be interesting to see if this helps with the timeouts that @akleber was mentioning in https://github.com/frerich/clcache/pull/244#issuecomment-264126981.

I did not do a complete analysis checking if the code is correct but it should be done if it proves to be useful.


siu:
I won't have too much time in the following days to work on this, so feel free to take the branch and apply the necessary changes or even close it.

frerich:
I *think* (after a first glance on the code) that this introduces an issue in which two concurrent clcache processes notice that there is no manifest, then build everything, and then both try to add the manifest. If this is true, then it could probably be resolved by first checking whether the manifest exists when adding it (after invoking the real compiler).

siu:
Ah, that is true, as it is now it is possible that the same ManifestEntry is added more than once to the Manifest. Both processes may arrive to addManifest() in https://github.com/frerich/clcache/pull/260/files#diff-c09a754611b438b1dd2362d7cc35c53dR1648 with the same entry and add it twice.

To simplify the code I also removed the reordering of the entries in the manifest upon hit because the entryIndex could have changed when the manifestSection is locked again.

These two cases could be handled by turning a Manifest into an OrderedDict with includesContentHash as key and [includeFiles, objectHash] as values. The uniqueness would be guaranteed by the dictionary and the sorting would be easy with move_to_end(key).

In any case I don't want to spend the time applying these changes if this brings no benefit for anyone. The code can be used for benchmarking as it is even if it is racy when the same file is compiled at the same time from two processes. What do you think?

frerich:
@siu Would you mind providing some benchmarks which measure the effect of this patch? I suspect it won't actually buy a lot of time, but I'd love to be proven wrong. Otherwise, I'd probably rather close this PR since it clearly introduces a race condition in its current form and the (potential) benefit is yet to be demonstrated.

siu:
I don't have any example of it helping, it certainly does not help in my case. In theory it should help compiling projects with many or large header files. It should enable higher concurrency when computing manifest hashes and reduce the possibility of timeouts.

But that is just a theory, I wanted to ask others to benchmark the branch to see if it helps with other real-world scenarios.

One way to benchmark it is to lower the timeout until the compilation starts failing, with and without this change. If what I was saying is right this branch should be less prone to timeouts.

sasobadovinac:
I made a few builds for this and latest release...

This commit:
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.494 msbuild time 37:57.01
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.495 msbuild time 10:48.15

Last commit from #255 
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.496 msbuild time 36:17.08
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.497 msbuild time 08:48.06

4.0.0 release
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.500 msbuild time 35:22.66
https://ci.appveyor.com/project/sasobadovinac/freecad/build/1.0.501 msbuild time 08:46.78

There are a few more builds in #248 that I have made in the past and can be compared to...

siu:
@sasobadovinac thanks a lot for benchmarking! It not only does not help but it hurts performance, I guess that reacquiring the lock is too expensive. The only remaining benefit that this PR could provide is reducing the chance of timeouts.

As this is looking I don't think it is worth to continue the work on this branch. If a specific project produces timeouts because of a large number of headers the workaround is simple: increase the timeout.

Let's archive this PR for now.

PR #262, opened 2016-12-19
Finer grained signal blocking

### LINKED ISSUE #261: Respect termination signals
[`clcache` ignores termination signals](https://github.com/frerich/clcache/blob/4b0ebec4de5dd55f19c1597a555c4f111e001d29/clcache.py#L1456-L1459).
This is bad because a single call to `cl` may take rather long. For example, when using large header-only libraries or invoking the compiler for many source files at once.

frerich:
In order to avoid that corrupt data (e.g. half of a JSON file) is written to disk when terminating clcache via a `SIGTERM` or `SIGNINT` handler, the script ignored those signals.

However, this means that calls to clcache can no longer be terminated via Ctrl+C. In particular, if a cache miss is encountered and a long-running call to `invokeRealCompiler()` is performed, there's no wayto abort this. This regressed when merging PR #233 .

This patch adjusts the code such that the signals are only blocked during cache-modifying phases, namely:

  * a cache miss in direct mode, after the real compiler finished
  * a cache miss in indirect mode, after the real compiler finished
  * cleaning the cache

By just blocking signals for these (comparatively short) phases, we ensure that no broken object files, no incomplete stdout output and no malformed JSON files are written to disk when pressing Ctrl+C. Furthermore, by blocking the signals on this 'high' level, we ensure that we don't get interrupted between populating the cache and updating the statistics.

Of course, terminating the process by force still may leave the cache in an undefined state so more work is needed to make clcache recover gracefully.

This resolves #261 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ clcache changelog
 
 ## Upcoming release
 
- * Nothing yet!
+ * Bugfix: Permit interrupting clcache again via Ctrl+C again (GH #261).
 
 ## clcache 4.0.0 (2016-12-15)
 
--- a/clcache.py
+++ b/clcache.py
@@ -177,10 +177,10 @@ class ManifestSection(object):
         manifestPath = self.manifestPath(manifestHash)
         printTraceStatement("Writing manifest with manifestHash = {} to {}".format(manifestHash, manifestPath))
         ensureDirectoryExists(self.manifestSectionDir)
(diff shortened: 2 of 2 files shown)


frerich:
Well, this failed rather spectacularly. It didn't occur to me that `signal.signal` only works in the main thread.

frerich:
Discarding this, I don't plan to continue in this direction.

PR #266, opened 2017-01-26
Add README section about executable installation using Chocolatey

tadeu:
I've created a project ([clcache-builder](https://github.com/ESSS/clcache-builder)) to build an executable version of `clcache.py` and to package it as a Chocolatey package, in order to make it easier to use `clcache` on environments without Python or with Python 2.

Please, feel free to review that project too. If you are interested in adding the package building to this repository, I could give some help,

PR #267, opened 2017-01-27
Add creation of Chocolatey and .zip packages with PyInstaller-generated clcache.exe

tadeu:
Trying to add the generation of these packages to this repository, as discussed in this PR https://github.com/frerich/clcache/pull/266


tadeu:
@frerich Artifacts generation is working. :)

I think this would be useful to be merged as is, but later, from here, you might want to add a `deploy` section that will automatically add the artifacts to the GitHub releases (usually only for tagged builds), something like this, but I didn't research how to do the version part, or how to do only for tagged releases:

```
deploy:
  release: $(version)
  description: 'clcache $(version)'
  provider: GitHub
  auth_token:
    secure: <the github oauth key encrypted>
  artifact: clcache.$(version).nupkg,clcache-$(version).zip
```

frerich:
Great job, seems to work for me! Just one question (this may well be just me being ignorant about how to use Chocolatey, I'm still somewhat new to it):

I installed the `clcache.4.0.0-dev.nupkg` file by downloading it and then running (with Chocolatey v0.10.3):

```sh
choco install clcache -source "%CD%" -version 4.0.0-dev
```

Does that seem okay? It causes a warning message to be printed (but then proceeds to install the package):

```
C:\Users\Frerich\Desktop>cinst clcache -source %CD% -version 4.0.0-dev
Parsing -version resulted in error (converted to warning):
 Cannot bundle unregistered option '-e'.
```

Is the `-dev` part of the version number triggering this, i.e. choco expects a certain version number format? Or is this something caused by a defect when building the `.nupkg` file?
I'm not sure whether this

tadeu:
Yes, I've found out that choco treats it specially when there's the -dev in
the version, without it, only `choco install clcache.4.0.0.nupkg` is
sufficient, but with it, you must also pass `--version=4.0.0-dev`, as in
the `appveyor.yml` file

frerich:
Merged, thanks a lot!

PR #268, opened 2017-02-05
Memcached protocol support 

hubx:
Hi,

last week I had time to follow up on #242. This add memcached protocol support for current master. This PR is not finished, but it is already quite extensive so I want to collect your feedback first before going further.

What is missing at this stage:
- We need a clever way to integrate the integrationtests into appveyor (e.g. spwaning a local memcached and injecting `CLCACHE_MEMCACHED` to ensure CI. They pass with a clear local and remote cache on my machine though
- I would like to make file strategy and memcache strategy independent. For that I would need to change how `set/getEntry` work. ATM the local cache is used as intermediary, so memcache always 'pollutes' the local cache 
`        cachedArtifacts = cache.getEntry(cachekey)  # memcache puts objectFile into local cache
         copyOrLink(cachedArtifacts.objectFilePath, objectFile)  # copy local cache to build folder`
- This is also needed to write more tests for cache logic for `CacheFileWithMemcacheFallbackStrategy` more easily
- Statistics can not distinguish between local and remote file hits/misses. If we want to keep the same approach of only displaying cache hits/miss per client, we can write into a separate `stats.txt` or store these numbers in memcache seperately. Otherwise we can just use memcache's global stats to display. 
- Memcache doesn't need the current Configuration object
- See `# XX inline comments`  

Looking forward to your feedback :)

// cc @heremaps/cci - looks like I'm doing the team notification wrong: @sschuberth @Jimilian @thoulen

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -45,10 +45,14 @@ install:
   # Install dependencies for clcachesrv such that we can lint the source code
   - pip install -r server\requirements.txt
 
+  # Install pymemcache for memcache storage
+  - pip install pymemcache
+
 build_script:
   - python clcache.py --help
   - python clcache.py -s
   - pylint --rcfile=.pylintrc clcache.py
+  - pylint --rcfile=.pylintrc storage.py
   - pylint --rcfile=.pylintrc unittests.py
   - pylint --rcfile=.pylintrc --disable=no-member integrationtests.py
   - pylint --rcfile=.pylintrc performancetests.py
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,10 @@ environment:
     - PYTHON_INSTALL: "C:\\Python35"
 
 install:
+  - appveyor DownloadFile "http://s3.amazonaws.com/downloads.northscale.com/memcached-win64-1.4.4-14.zip" -FileName memcached.zip
+  - 7z x memcached.zip -y
+  - ps: $Memcached = Start-Process memcached\memcached.exe -PassThru
+
   # Make compiler available (use MSVC 2013, 32 bit)
   - call "%ProgramFiles(x86)%\Microsoft Visual Studio 12.0\VC\vcvarsall.bat" x86
 
@@ -48,6 +52,10 @@ install:
   # Install pymemcache for memcache storage
   - pip install pymemcache
 
+cache:
+  - '%LOCALAPPDATA%\pip\Cache -> appveyor.yml'
+  - memcached -> appveyor.yml
+
 build_script:
   - python clcache.py --help
   - python clcache.py -s
@@ -64,7 +72,7 @@ test_script:
   # Run test files via py.test and generate JUnit XML. Then push test results
   # to appveyor. The plugin pytest-cov takes care of coverage.
   - ps: |
-      & py.test --junitxml .\unittests.xml unittests.py --cov=clcache
+      & py.test --junitxml .\unittests.xml unittests.py --cov=clcache --cov=storage
       $testsExitCode = $lastexitcode
       & coverage report
       & coverage xml
@@ -91,6 +99,27 @@ test_script:
       & codecov --no-color -X gcov -F integrationtests -e PYTHON_INSTALL
 
       if ($testsExitCode -ne 0) {exit $testsExitCode}
+
+  - set CLCACHE_MEMCACHED=127.0.0.1:11211
+  - ps: |
+      & py.test --junitxml .\integrationtests.xml integrationtests.py --cov=clcache --cov=storage
+      $testsExitCode = $lastexitcode
+      & coverage report
+      & coverage xml
+
+      $wc = New-Object 'System.Net.WebClient'
+      $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\integrationtests.xml))
+
+      & codecov --no-color -X gcov -F integrationtests_memcached -e PYTHON_INSTALL
+
+      if ($testsExitCode -ne 0) {exit $testsExitCode}
+
+  - coverage erase
+  - del /Q coverage.xml
+  - set "CLCACHE_MEMCACHED="
+
+on_finish:
+  - ps: Stop-Process -Id $Memcached.Id
 
 on_success:
   # Creates executable with PyInstaller
--- a/clcache.py
+++ b/clcache.py
@@ -510,6 +510,18 @@ class CacheFileStrategy(object):
              self.statistics.lock:
             yield
 
+    def lockFor(self, key):
+        assert isinstance(self.compilerArtifactsRepository.section(key).lock, CacheLock)
+        return self.compilerArtifactsRepository.section(key).lock
+
+    def getEntry(self, key):
+        return self.compilerArtifactsRepository.section(key).getEntry(key)
+
+    def setEntry(self, key, value):
+        self.compilerArtifactsRepository.section(key).setEntry(key, value)
+
+    def hasEntry(self, cachekey):
+        return self.compilerArtifactsRepository.section(cachekey).hasEntry(cachekey)
 
     def clean(self, stats, maximumSize):
         currentSize = stats.currentCacheSize()
@@ -551,10 +563,6 @@ class Cache(object):
         return self.strategy.manifestRepository
 
     @property
-    def compilerArtifactsRepository(self):
-        return self.strategy.compilerArtifactsRepository
-
-    @property
     def configuration(self):
         return self.strategy.configuration
 
@@ -564,6 +572,20 @@ class Cache(object):
 
     def clean(self, stats, maximumSize):
         return self.strategy.clean(stats, maximumSize)
+
+    @contextlib.contextmanager
+    def lockFor(self, key):
+        with self.strategy.lockFor(key):
+            yield
+
+    def getEntry(self, key):
+        return self.strategy.getEntry(key)
+
+    def setEntry(self, key, value):
+        self.strategy.setEntry(key, value)
+
+    def hasEntry(self, cachekey):
+        return self.strategy.hasEntry(cachekey)
 
 
 class PersistentJSONDict(object):
@@ -1389,12 +1411,12 @@ def parseIncludesSet(compilerOutput, sourceFile, strip):
         return includesSet, compilerOutput
 
 
-def addObjectToCache(stats, cache, section, cachekey, artifacts):
+def addObjectToCache(stats, cache, cachekey, artifacts):
     # This function asserts that the caller locked 'section' and 'stats'
     # already and also saves them
     printTraceStatement("Adding file {} to cache using key {}".format(artifacts.objectFilePath, cachekey))
 
-    section.setEntry(cachekey, artifacts)
+    cache.setEntry(cachekey, artifacts)
     stats.registerCacheEntry(os.path.getsize(artifacts.objectFilePath))
 
     with cache.configuration as cfg:
@@ -1404,15 +1426,14 @@ def addObjectToCache(stats, cache, section, cachekey, artifacts):
 def processCacheHit(cache, objectFile, cachekey):
     printTraceStatement("Reusing cached object for key {} for object file {}".format(cachekey, objectFile))
 
-    section = cache.compilerArtifactsRepository.section(cachekey)
-    with section.lock:
+    with cache.lockFor(cachekey):
         with cache.statistics.lock, cache.statistics as stats:
             stats.registerCacheHit()
 
         if os.path.exists(objectFile):
             os.remove(objectFile)
 
-        cachedArtifacts = section.getEntry(cachekey)
+        cachedArtifacts = cache.getEntry(cachekey)
         copyOrLink(cachedArtifacts.objectFilePath, objectFile)
         printTraceStatement("Finished. Exit code 0")
         return 0, cachedArtifacts.stdout, cachedArtifacts.stderr, False
@@ -1623,9 +1644,9 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
                         manifest.touchEntry(entryIndex)
                         manifestSection.setManifest(manifestHash, manifest)
 
-                        artifactSection = cache.compilerArtifactsRepository.section(cachekey)
-                        with artifactSection.lock:
-                            if artifactSection.hasEntry(cachekey):
+                        artifactSection = True
+                        with cache.lockFor(cachekey):
+                            if cache.hasEntry(cachekey):
                                 return processCacheHit(cache, objectFile, cachekey)
 
                 except IncludeNotFoundException:
@@ -1648,7 +1669,7 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
 
     with manifestSection.lock:
         if artifactSection is not None:
-            return ensureArtifactsExist(cache, artifactSection, cachekey, unusableManifestMissReason,
+            return ensureArtifactsExist(cache, cachekey, unusableManifestMissReason,
                                         objectFile, compilerResult)
 
         entry = createManifestEntry(manifestHash, includePaths)
@@ -1659,34 +1680,32 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
             manifest.addEntry(entry)
             manifestSection.setManifest(manifestHash, manifest)
 
-        artifactSection = cache.compilerArtifactsRepository.section(cachekey)
-        return ensureArtifactsExist(cache, artifactSection, cachekey, unusableManifestMissReason,
+        return ensureArtifactsExist(cache, cachekey, unusableManifestMissReason,
                                     objectFile, compilerResult, addManifest)
 
 
 def processNoDirect(cache, objectFile, compiler, cmdLine, environment):
     cachekey = CompilerArtifactsRepository.computeKeyNodirect(compiler, cmdLine, environment)
-    artifactSection = cache.compilerArtifactsRepository.section(cachekey)
-    with artifactSection.lock:
-        if artifactSection.hasEntry(cachekey):
+    with cache.lockFor(cachekey):
+        if cache.hasEntry(cachekey):
             return processCacheHit(cache, objectFile, cachekey)
 
     compilerResult = invokeRealCompiler(compiler, cmdLine, captureOutput=True, environment=environment)
 
-    return ensureArtifactsExist(cache, artifactSection, cachekey, Statistics.registerCacheMiss,
+    return ensureArtifactsExist(cache, cachekey, Statistics.registerCacheMiss,
                                 objectFile, compilerResult)
 
 
-def ensureArtifactsExist(cache, section, cachekey, reason, objectFile, compilerResult, extraCallable=None):
+def ensureArtifactsExist(cache, cachekey, reason, objectFile, compilerResult, extraCallable=None):
     cleanupRequired = False
     returnCode, compilerOutput, compilerStderr = compilerResult
-    with section.lock:
-        if not section.hasEntry(cachekey):
+    with cache.lockFor(cachekey):
+        if not cache.hasEntry(cachekey):
             with cache.statistics.lock, cache.statistics as stats:
                 reason(stats)
                 if returnCode == 0 and os.path.exists(objectFile):
                     artifacts = CompilerArtifacts(objectFile, compilerOutput, compilerStderr)
-                    cleanupRequired = addObjectToCache(stats, cache, section, cachekey, artifacts)
+                    cleanupRequired = addObjectToCache(stats, cache, cachekey, artifacts)
                     if extraCallable:
                         extraCallable()
     return returnCode, compilerOutput, compilerStderr, cleanupRequired
--- a/clcache.py
+++ b/clcache.py
@@ -500,7 +500,7 @@ class CacheFileStrategy(object):
         self.statistics = Statistics(os.path.join(self.dir, "stats.txt"))
 
     def __str__(self):
-        return "Disk cached at {}".format(self.dir)
+        return "Disk cache at {}".format(self.dir)
 
     @property
     @contextlib.contextmanager
@@ -578,7 +578,7 @@ class Cache(object):
             self.strategy = CacheFileStrategy(cacheDirectory=cacheDirectory)
 
     def __str__(self):
-        self.strategy.__str__(self)
+        return str(self.strategy)
 
     @property
     def lock(self):
@@ -1371,7 +1371,7 @@ clcache statistics:
 
     with cache.statistics as stats, cache.configuration as cfg:
         print(template.format(
-            str(cache()),
+            str(cache),
             stats.currentCacheSize(),
             cfg.maximumCacheSize(),
             stats.numCacheEntries(),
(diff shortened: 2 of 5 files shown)


sschuberth:
>  looks like I'm doing the team notification wrong

Not really, but team notification only work for repos within the same org the team belongs to :-/

storage.py
```diff
@@ -215,20 +216,19 @@ def configuration(self):
...
-    def manifestLockFor(key):
+    def manifestLockFor(_):
         return CacheDummyLock()
 
     @property
     @contextlib.contextmanager
     def lock(self):
-        with self.remoteCache.lock:
-            self.localCache.lock
+        with self.remoteCache.lock, self.localCache.lock:
```
hubx: neat, pylint found an actually error that I overlooked in a hurry.

frerich:
Looks interesting! It still looks like work in progress indeed, but I very much agree with the points you listed as to the things which would still need to be addressed. I'm a little worried about the `CacheFileStrategy` (maybe that should just be called `CacheStrategy`?) class and the set of exposed methods though. It might be hard to get this right so that some implementation doesn't end up with "dummy" implementations.

When devising it's interface, it might be interesting to have a third implementation (besides file system and memcache backed caches) in mind -- one idea I've been toying with for a long time was to try a SQL or document-based backend, hoping that it would take care of things like locking, concurrency, marking of recently used items (setting the last-access time does not usually work on Windows, which hurts automatic cache cleaning) and maybe even distribution.

server/clcachesrv.py
```diff
@@ -36,8 +36,11 @@ def getFileHash(self, path):
         return hashsum
 
     def _startWatching(self, dirname):
-        ev = pyuv.fs.FSEvent(self._loop)
-        ev.start(dirname, 0, self._onPathChange)
+        self.ev = pyuv.fs.FSEvent(self._loop)
+        self.ev.start(dirname, 0, self._onPathChange)
+        
+    def __del__(self):
+        self.ev.stop()
```
frerich: This patch seems to be worth it's own PR, i.e. it appears to be a bit out of scope for the memcached support, doesn't it?

Jimilian: `HashCache` is shared between different paths. But right now you stop to listen only last one - leak a still in place.
I think it makes sense to fill new PR and address it a little bit different way:
   Start new handle -> put result object to list -> close all of them in the loop.

hubx:
thanks for the feedback. It's true fixing clcachesrv file descriptors leaks is another PR. As for you last comment @frerich can you elaborate more. I don't know what to do so this PR gets merged? What's needs to fixed now, what can be improved later?

frerich:
@hubx My main concern with this PR is that I have no experience with memcache myself and there are no tests part of the CI which exercise this functionality. I believe that the code coverage only looks okay because the `appveyor.yml` file makes AppVeyor run `py.test` such that it only calculates the coverage for `clcache.py` (I think that's what the `--cov=clcache` argument does). If it were to also generate for `storage.py`, we would maybe see that this functionality is not exercised too much.

I.e. I trust you that it works, but I don't see how I could maintain it as it is: I don't feel I could refactor the code in the future with a good conscience (let alone fixing bugs in the code). If you could teach AppVeyor to test all this stuff, that would be much appreciated. :-)

hubx:
@frerich tests are a fair point. I'll add storage.py to be included in the code coverage. As for adding tests, I'll try to add the integration tests with memcache storage in AppVeyor as well. E.g. if available we start a pure python memcached in the integrations tests. Otherwise we can start memcached like [this](https://github.com/deepstreamIO/deepstream.io-cache-memcached/blob/master/appveyor.yml#L15) in AppVeyor. Please raise other concerns you might have. I don't want to push the project in a direction you are not comfortable with.

frerich:
@hubx Thanks a lot!

Besides the maintainability aspect (which useful tests with good coverage executed as part of the CI already help a lot with!), there's indeed another "concern" in my mind - and it's also not directly related to the memcache backend functionality but rather a more long-term goal.

The file-system based database has shown to have a couple of weaknesses over the last years, e.g.

* We needed to introduce the notion of sections in order to get locking right (which we also need to do manually).
* There is currently no good way to share a cache among multiple hosts.
* It appears that file access performance degrades quite badly if there are a lot of directories (e.g. a few thousand) in a single section directory.
* Modern windows versions have updating of the last access time disabled by default, so the 'clean' functionality is currently not working as intended in that it does not actually delete the least recently used files.

One idea which I've been toying with is to introduce a backend interface, much like what you did with `Storage`, which abstracts the filesystem-based database such that I could also provide a PostgreSQL-based backend. I hope it would perform reasonably well while taking care of all of the above points. I never really thought much about how that interface would look like, so my worry is that the interface you introduce is mostly tailored to what memcache (i.e. a particular implementation) needs as opposed to what clcache (i.e. the client) wants. In that case, I'd probably need to re-do the interface in case I ever get around to adding other backends.

However, this is all quite hand-wavy at this point. Maybe it's just fine the way it is and I'll change things once I have a more precise understanding of what I want.

frerich:
By the way, does this PR supersede #242 , i.e. that older PR can be closed?

hubx:
@frerich You are right, but I think you can adjust/refactor the "Cache interface" in the storage module later. Especially if there are tests. I decided against to move the FileStrategy to the storage module for now, since this a further step. In that step, you would have to untangle application transaction locks and file access locks.

I'm sure with hstore/JSON/JSONB Postgresql can also act as a decent cache, but check out recent specialized application caches like redis or memcached before.

I closed #242

hubx:
The test run on appveyor, [see here](https://ci.appveyor.com/project/frerich/clcache/build/1.0.1014/job/b1k53fvd6hgmhatt)

You see that it intermediately hit these application lock, which I did not want to touch in this PR ;) :

`clcache.CacheLockException: Failed to acquire lock Local\C--Users-appveyor-clcache-stats.txt after 10000ms; try setting CLCACHE_OBJECT_CACHE_TIMEOUT_MS environment variable to a larger value.`

Locally I set this to a much larger value, that is why its only happening on appveyor :)

I think this happens because extraCallable() can be addManifest() which can a bit a time with the stats being a single synchronization point.
[See](https://github.com/frerich/clcache/pull/268/commits/ebbcb56dcbf5ac5fe421b4750110c60587d729d3#diff-c09a754611b438b1dd2362d7cc35c53dR1741 ) - If this is fixing the issue, I will do this in a separate clean commit.

Alternatively, I can increase the timeout. Not sure if CLCACHE_OBJECT_CACHE_TIMEOUT_MS is considered part of the tests.

hubx:
I just saw the comment (This function asserts that the caller locked 'section' and 'stats' already and also saves them) in `addObjectToCache`. IMHO if you lock that the statistics while doing setEntry there. You still lock the whole cache in a lot of cases, splitting into sections can not use its full potential. But let's do this later. I wIll only unindent extraCallable() for now

hubx:
Okay looks like AppVeyor is slow. I meant something along: https://github.com/heremaps/clcache/commit/351d9f1be712a288e692b20bbf7d5bd2d5c3eaf3 where stats.txt is only locked when written to it. If we need to lock Statistics._stats within a process, we should do it locally not globally

hubx:
Any idea why AppVeyor keeps getting stuck? Is installing couchbase-server-community / 200mb too much? Then we need to fix https://chocolatey.org/packages/memcached

frerich:
@hubx Just wanted to give a big thumbs up for your work! I'm currently attending an exhibition though (EmbeddedWorld in Nuremberg) so I only actually get to look after my GitHub-related work when I'm back in the hotel late in the evening. I hope that I'll get around to investigating what's up with AppVeyor later this week (certainly on Friday when I'll be back home).

sschuberth:
> Any idea why AppVeyor keeps getting stuck?

@hubx Just an idea for another approach: Try using the [Start-FileDownload cmdlet](https://www.appveyor.com/docs/how-to/download-file/#start-filedownload-cmdlet) to download https://packages.couchbase.com/releases/4.5.0/couchbase-server-community_4.5.0-windows_amd64.exe directly and run it manually.

Also, you should leverage the [build cache](https://www.appveyor.com/docs/build-cache/) to avoid downloading the installer over and over again.

hubx:
So will the latest status the test run using memcached-1.4.4
- @frerich Is the dependency to http://s3.amazonaws.com/downloads.northscale.com" ok? I choose this old version rather then downloading 200mb.

I'm not sure if this always download the file or if `appveyor DownloadFile` already has a mechanism to leverage the cache - since I'm changing appveyor.yml the cache is also invalidated - but we fix this later I suppose

appveyor.yml
```diff
@@ -74,7 +86,7 @@ test_script:
   - del /Q coverage.xml
 
   - ps: |
-      & py.test --junitxml .\integrationtests.xml integrationtests.py --cov=clcache
+      & py.test --junitxml .\integrationtests.xml integrationtests.py --cov=clcache --cov=storage
```
frerich: I don't think this change is needed; the integration tests won't even import the `storage` module unless `CLCACHE_MEMCACHED` is set (and `py.test` correctly points out that no coverage data will be generated for `storage` since that file is not even used).

appveyor.yml
```diff
@@ -86,6 +98,27 @@ test_script:
...
+  - ps: |
+      & py.test --junitxml .\integrationtests.xml integrationtests.py --cov=clcache --cov=storage
+      $testsExitCode = $lastexitcode
+      & coverage report
+      & coverage xml
+
+      $wc = New-Object 'System.Net.WebClient'
+      $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\integrationtests.xml))
+
+      & codecov --no-color -X gcov -F integrationtests -e PYTHON_INSTALL
```
frerich: I think it would be nice to use a separate flag here, i.e. instead of `-F integrationtests` maybe use `-F integrationtests_memcached`. The benefit of this is that one can discriminate the memcached-specific portions more clearly in the coverage report. I think(!) that without this, you'll overwrite the coverage report for the earlier upload with the `integrationtests` flag (i.e. the run without having memcached enabled).

appveyor.yml
hubx: I thought If no test are run - it should be clear that storage.py coverage is 0% - so we can track regressions

appveyor.yml
hubx: alright, will do

hubx:
@frerich sorry for letting you wait so long - I hope I addressed your remarks. There are some XX / Todo markers remaining in the commits - Should we address them before merging?

Yes, we use this PR in production already. However only in indirect mode. In direct mode there is currently a problem that local manifests do not get updated, hence after a couple of hours we see a lot of manifest misses, which trickles down to direct compiler invocation. I think ccache tries indrect mode as fall back before invoking the compiler? We could try that as well. The problem is as follows

```python
def processDirect(..)
...
        manifest = cache.getManifest(manifestHash) 
        # will take local manifest first w/o updating it from remote
        if manifest:
              ..
             unusableManifestMissReason = Statistics.registerHeaderChangedMiss 
             # local manifest is unuseable
# invoke compiler
...
return ensureArtifactsExist(cache, cachekey, unusableManifestMissReason,
                                    objectFile, compilerResult, addManifest)
...
def ensureArtifactsExist(
    with cache.lockFor(cachekey):
        if not cache.hasEntry(cachekey): 
       # does not hit because cachekey is in remote cache
      ..
          extraCallable() # == addManifest
```

I don't how to fix it without making more substantial changes to the cacheHit logic here, e.g. if having a *registerHeaderChangedMiss* we could make updating the manifest directly in processCacheHit. I don't know why addManifest() - which is rather addManifestEntry? - is done via a callback from ensureArtifactsExist?). E.g. for Memcache specifically we could fetch the manifest from remote and see if local and remote manifest differ and than hit the manifest content checking again. This is strategy specific cache lookup code. I would like to do this in a second step later. But it shows there is a lack of tests for this scenario.

But there is also more room for more strategy specific cache lookup code. E.g. that touchEntry() and cache.setManifest(manifestHash, manifest) is triggered for every manifestHit is not optimal for memcache.

Please let me know your thoughts.

frerich:
@hubx There's no rush, of course anybody is working on this little project voluntarily. I'm happy that you bother contributing something back in the first place! 👍

I'd like to address the thoughts/ideas your raised:

> I think ccache tries indrect mode as fall back before invoking the compiler? We could try that as well.

Really, ccache does that? That's surprising! I didn't consider this idea so far but treated the two approaches to detecting repeated compilations as mutually exclusive. I would need to think about it to make up my mind.

>  I don't know why addManifest() - which is rather addManifestEntry? - is done via a callback from ensureArtifactsExist?).

Yes, maybe `addManifestEntry` is a better name. However, the method also adds a new manifest if needed. Maybe `setManifestEntry` then?

The function for updating the statistics is specified via a callback since updating the statistics should ideally happen at the same moment (i.e. while holding the same locks!) as when the cache itself is updated. Otherwise, there may be short time windows in which the statistics do not actually match the cache contents. The decision if the cache needs to be updated (and hence, if the statistics need updating) and when is done within [`ensureArtifactExists`](https://github.com/frerich/clcache/blob/master/clcache.py#L1649), but the decision *which* statistics field should be updated is done by the callers. Hence, they pass the function to invoke as an argument.

That being said, I'm afraid I still don't understand the issue with the cache misses fully. My understanding is that when using memcache as the backend, the manifest repository is not trustworthy since it only indicates the state of the local cache and hence it can happen that a manifest is unusable despite an appropriate cache entry being available -- remotely? Is that it? If so, why does it only happen after some time?

hubx:
@frerich sorry for diverging - let's focus on the manifest miss thing first. 

I think you already grasp the issue, basically a local manifest can miss an updated entry, yes. However then local manifests do not get updated at all anymore even after invoking the real compiler locally. Because `extraCallable()` / `addManifest()` is not called in `ensureArtifactsExist()` because the remote cache already has the entry that came out of the repeated compiler invocation.

So the current cache hit logic assumes, if it has the cachekey in cache - it also has an entry in the corresponding manifest, which is not the case. 

It occurs from the first run. However over time the number of misses increases - so that over 2hrs it is quite noticeable.  

Can there be actually a case where includeFiles will not match the ones from `manifestHash = ManifestRepository.getManifestHash(compiler, **cmdLine**, sourceFile)`. We would have landed in this particular manifest_hash.json in the first place? Also I don't I understand why we need to hash includeFiles for every entry in the manifest? The source of truth are the includes in our soure code checkout? We can hash them once and see if there are  includesContentHash matching in the manifest?

```json
{
  "entries": [
    {
      "includeFiles": ...
      "includesContentHash": "f8cc93d189da662236471019401e33b3",
      "objectHash": "084fec7bb5fc889cc46450ea1cd78a7b"
    },...
  ]
}

we can just have 
```json
{[{"f8cc93d189da662236471019401e33b3":"084fec7bb5fc889cc46450ea1cd78a7b"}]}
``` or what do I miss?

**Comment by gferon** [comment]
I just noticed this pull request. This is definitely an improvement for my use-case! I started to use it, and so far, so good! :+1:

**Comment by frerich** [comment]
@hubx Just a quick heads-up: alas I keep getting distracted by other stuff so I didn't really get around to thinking about the manifest-related issue you describe. It's not forgotten though! :-)

**Comment by hubx** [comment]
@frerich sure. I something like f8fbef4 in mind that specifically for the FileCacheStrategy would call `cache.updateManifest(manifestHash)`

**Comment by hubx** [comment]
@frerich any thoughts on my previous comment?

**Comment by frerich** [comment]
@hubx I finally found some quiet hours to think about f8fbef4 in detail. I think it makes sense to introduce a little 'hook' like `updateManifest` but I'm not really happy with that name. If I were to implement a new storage strategy, how would I know whether I would want to reimplement that method - and how? The semantics of that method are not clear to me.

**Comment by frerich** [comment]
@hubx To follow-up on my last commit: I think if we can make the purpose of the newly introduced `updateManifest` method more clear (i.e. what it should and should not do), then I'm fine with merging this stuff since it all appears to be hidden by the `CLCACHE_MEMCACHED` environment variable.

I think this would warrant a new minor release to get more exposure and check for bugs.

**Comment by hubx** [comment]
@frerich  pylint was upgraded from 1.6.5 to 1.7 therefore it produces new, unrelated warnings/errors. We can extract it to a requirements_ci.txt and pin the dependency there?

**Comment by frerich** [comment]
@hubx I just merged #274 which makes pylint 1.7.1 happy again - if you rebase your branch on top of current master, you should get those improvements, too.

**Comment by hubx** [comment]
@frerich done

**Comment by frerich** [comment]
@hubx Thanks! I just re-read the code and wonder: is there any benefit in having the `Cache` class at all? It seems to me is that it just forwards all calls to methods with the same name of the `strategy` class.

I suspect one could drop it completely and then, in `main`, use the code of the `Cache` constructor to decide which strategy to instantiate, i.e. the `cache` variable in `main` is one of the two strategies (I suppose the term `strategy` would be a bit misleading then, at least when using it in the classical Design Patterns sense).

**Comment by hubx** [comment]
@frerich So do I understand you correctly, are asking: "Why an interface in strategy patterns?" or more generic "Why interfaces?", or is it more specific "Why interfaces in python?" Sorry if I got this wrong, then ignore the following.

There is lots of discussion available all around the keyword _code contracts_ [1](https://softwareengineering.stackexchange.com/questions/108240/why-are-interfaces-useful), [2](https://stackoverflow.com/questions/240152/why-would-i-want-to-use-interfaces), [3](https://stackoverflow.com/questions/1686174/when-should-one-use-interfaces). 

In this specific instance I would say it allows to think about the `Cache` interface decoupled from the actual File/Database implementation. "What should a cache object be able to do". As we actually did previously in this thread.  Only now we can argue if `setManifest()` should be part of `Cache` or should it rather go to `ManifestRepository`. Why is `processCacheHit(cache, objectFile, cachekey)` not `cache.processHit(key, object)` etc. This gets harder without clear interfaces the more independent implementations are available in the code base. Let's say you want to add the Postgresql backend you mentioned, will you do it by locking at the existing strategies or can look into a single interface you need to implement? 

Secondly, for testing it would allow to mock a cache interfaces. E.g. instead of having end-to integration tests, there could be one processHit unit tests which counts the calls to the mocked Cache interface w/o calls to the real compiler.

But it is true that clcache might not need a strict architecture as bigger projects have and interfaces can play its strength with statically typed languages.

**Comment by frerich** [comment]
@hubx I really only meant to ask the question for the specific case at hand; with just two strategies implemented, there didn't seem to be a lot of extra value to me.

I concur with the benefits of having a single class documenting the interface which a storage backend has to implement, but as it is this doesn't seem to exist in the code:

* The two storage backends don't have a common base class which I'd expect to see in case there was a common interface (but maybe using inheritance for interface reuse is not common in Python?). Instead, one has to read the implementation of the `Cache` class to figure out the interface, i.e. you actually have to read the class definition to see that it indeed forwards all calls.

* Even if you do read the `Cache` definition, you still don't know which methods you *need* to implement: right now, it appears that the `lockFor` and `manifestLockFor` methods don't need to be implemented (or do they?) by the Memcache backend.

* For testing, I believe you could just as well have a mocked cache strategy object which does what you suggest (tracking calls to the methods).

Now, I'm not dogmatic about this - since the code is already there, I won't insist on removing it. I just think that with just two strategies in the backend and a language in which duck typing is dominant, the `Cache` class does not really pull its own weight anymore. I just wanted to bring this up to check whether it was maybe a mistake to leave it in (since basically all its logic was moved elsewhere except for the constructor code). :-)

With that being said - I see there are still some `XX` comments in the patch which suggest that there's something missing yet?

**Comment by frerich** [comment]
@hubx Other than these fishy `XX` comments, I'd say this stuff is good to go - or do you have anything on your mind which should prohibit using this stuff in production?

File: storage.py [hunk 5]
```diff
@@ -0,0 +1,235 @@
...
+        pass
+
+    def __exit__(self, typ, value, traceback):
+        pass
+
+
+class CacheMemcacheStrategy(object):
+    def __init__(self, server, cacheDirectory=None, manifestPrefix='manifests_', objectPrefix='objects_'):
+        self.fileStrategy = CacheFileStrategy(cacheDirectory=cacheDirectory)
+        # XX Memcache Strategy should be independent
```
hubx: - I would like to make file strategy and memcache strategy independent. For that I would need to change how `set/getEntry` work. ATM the local cache is used as intermediary, so memcache always 'pollutes' the local cache 
`        cachedArtifacts = cache.getEntry(cachekey)  # memcache puts objectFile into local cache
         copyOrLink(cachedArtifacts.objectFilePath, objectFile)  # copy local cache to build folder`

storage.py
```diff
@@ -0,0 +1,235 @@
...
+                                  serializer=python_memcache_serializer,
+                                  deserializer=python_memcache_deserializer,
+                                  timeout=5,
+                                  connect_timeout=5,
+                                  key_prefix=(getStringHash(self.fileStrategy.dir) + "_").encode("UTF-8")
+                                 )
+        # XX key_prefix ties fileStrategy cache to memcache entry
+        # because tests currently the integration tests use this to start with clean cache
+        # Prevents from having cache hits in when code base is in different locations
+        # adding code to production just for testing purposes
```
hubx: my concern is that we implement it this way just for testing, to enable
tempfile.TemporaryDirectory() as tempDir:
customEnv = dict(os.environ, CLCACHE_DIR=tempDir)
subprocess.check_call(baseCmd + ["fibonacci01.cpp", "fibonacci02.cpp"], env=customEnv)

storage.py
```diff
@@ -0,0 +1,235 @@
...
+    def __str__(self):
+        return "Remote Memcache @{} object-prefix: {}".format(self.server, self.objectPrefix)
+
+    # XX Global lock here?
+    # XX memcache has own statistics about hit/misses
+    @property
+    def statistics(self):
+        return self.fileStrategy.statistics
+
+    # XX Global lock here?
```
hubx: I think my comment is valid: `with cache.statistics.lock, cache.statistics as stats:` in `ensureArtifactsExists` is a global barrier. Where MemcachedStrategy could suffer local locking. It is fine if there are enough parallel compilations though

storage.py
```diff
@@ -0,0 +1,235 @@
...
+            self._fetchEntry(key)
+        if self.localCache[key] is None:
+            return None
+        data = self.localCache[key]
+
+        printTraceStatement("{} remote cache hit for {} dumping into local cache".format(self, key))
+
+        assert len(data) == 3
+
+        # XX this is writing the remote objectfile into the local cache
```
hubx: see line 22

clcache.py
```diff
@@ -1604,60 +1689,61 @@ def processDirect(cache, objectFile, compiler, cmdLine, sourceFile):
...
         cachekey = entry.objectHash
 
         def addManifest():
-            manifest = manifestSection.getManifest(manifestHash) or Manifest()
+
+            manifest = cache.getManifest(manifestHash) or Manifest()
             manifest.addEntry(entry)
-            manifestSection.setManifest(manifestHash, manifest)
+            # XX not locking manifests here?
+            cache.setManifest(manifestHash, manifest)
```
hubx: see line 1745 is not locking this call, but we are writing to the manifest here

hubx:
@frerich there are 3:
- Potential global locking where performance would suffer
- Currently FileStrategy and MemcacheStrategy are tightly coupled/dependent as mentioned in my initial comment here
- One potential logic error in addManifest

See my inline comments. I would address these in 3 different PRs.  We use this in non-direct mode for a couple of months in production now. So I propose the following:
- The XX comments for (II) I would keep in the code.
- The others I would remove.

hubx:
@frerich I amended my commits as stated as above. For "The XX comments for (II) I would keep in the code." I would create an issue to keep track of this once the PR is merged.

frerich:
Merged, thanks! I think it would be really cool to refactor how `getEntry`/`setEntry` work such that the memcache strategy doesn't need any local storage.

PR #271, opened 2017-03-03
Fix leaks in clcachesrv

Jimilian:
This PR tries to address two issues:
1) In case of SIGTERM clcachesrv doesn't close handlers gracefully
2) It doesn't stop event listeners in any case

Unfortunately, I didn't find any tests for clcachesrv - so, I didn't create new one.


frerich:
Looks good, merged! Thanks!

PR #272, opened 2017-03-06
adds the possibility to exclude directories from being watched

akleber:
Watched directories can not be deleted and this might be problematic
in some environments.
Watching of directories can be selectively disabled by providing one
or more exclude patterns or disabled completely.

### DIFF
--- a/server/clcachesrv.py
+++ b/server/clcachesrv.py
@@ -5,14 +5,18 @@ import logging
 import os
 import pickle
 import signal
+import argparse
+import re
 
 import pyuv
 
 class HashCache(object):
-    def __init__(self, loop):
+    def __init__(self, loop, excludePatterns, disableWatching):
         self._loop = loop
         self._watchedDirectories = {}
         self._handlers = []
+        self._excludePatterns = excludePatterns if excludePatterns else []
+        self._disableWatching = disableWatching
 
     def getFileHash(self, path):
         logging.debug("getting hash for %s", path)
@@ -28,10 +32,11 @@ class HashCache(object):
             hashsum = hashlib.md5(f.read()).hexdigest()
 
         watchedDirectory[basename] = hashsum
-        if dirname not in self._watchedDirectories:
+        if dirname not in self._watchedDirectories and not self.isExcluded(dirname) and not self._disableWatching:
             logging.debug("starting to watch directory %s for changes", dirname)
             self._startWatching(dirname)
-            self._watchedDirectories[dirname] = watchedDirectory
+        
+        self._watchedDirectories[dirname] = watchedDirectory
 
         logging.debug("calculated and stored hashsum %s", hashsum)
         return hashsum
@@ -51,6 +56,14 @@ class HashCache(object):
     def __del__(self):
         for ev in self._handlers:
             ev.stop()
+
+    def isExcluded(self, dirname):
+        # as long as we do not have more than _MAXCACHE regex we can
+        # rely on the internal cacheing of re.match
+        excluded = any(re.search(pattern, dirname, re.IGNORECASE) for pattern in self._excludePatterns)
+        if excluded:
+            logging.debug("NOT watching %s", dirname)
+        return excluded
 
 
 class Connection(object):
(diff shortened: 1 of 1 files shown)


server/clcachesrv.py
```diff
@@ -5,14 +5,18 @@
...
 
 import pyuv
 
 class HashCache(object):
-    def __init__(self, loop):
+    def __init__(self, loop, excludePatterns, disableWatching):
         self._loop = loop
         self._watchedDirectories = {}
         self._handlers = []
+        self._excludePatterns = excludePatterns if excludePatterns else []
```
frerich: Just a minor remark - this expression can be shortened to `excludePatterns or []`.

server/clcachesrv.py
```diff
@@ -28,10 +32,11 @@ def getFileHash(self, path):
             hashsum = hashlib.md5(f.read()).hexdigest()
 
         watchedDirectory[basename] = hashsum
-        if dirname not in self._watchedDirectories:
+        if dirname not in self._watchedDirectories and not self.isExcluded(dirname) and not self._disableWatching:
```
frerich: This condition makes me a bit dizzy, not only because of the number of boolean expressions involved but also because terms like `isExcluded` or `disableWatching` are negations in themselves. Not sure how to simplify it though.

server/clcachesrv.py
```diff
@@ -114,9 +127,24 @@ def onSigterm(handle, signum):
 def main():
     logging.basicConfig(format='%(asctime)s [%(levelname)s]: %(message)s', level=logging.INFO)
 
+    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser.add_argument('--exclude', metavar='REGEX', action='append', \
+                        help='Regex ( re.search() ) for exluding of directory watching. Can be specified multiple times. \
+                              Pathnames to test against are after os.path.normcase(), i.e. all lowercase with backslashes. \
```
frerich: I think there's no pressing need to mention the implementation detail of `os.path.normcase` being used in the help output. How about omitting the entire second sentence and just relying on it working as expected? Since this is on Windows, I'd certainly expect the paths to be case-insensitive and backslashes being used as the separators.

server/clcachesrv.py
```diff
@@ -114,9 +127,24 @@ def onSigterm(handle, signum):
 def main():
     logging.basicConfig(format='%(asctime)s [%(levelname)s]: %(message)s', level=logging.INFO)
 
+    parser = argparse.ArgumentParser(description='Process some integers.')
```
frerich: The `description` looks odd, probably a leftover from copying this out of some example?

server/clcachesrv.py
```diff
@@ -114,9 +127,24 @@ def onSigterm(handle, signum):
...
+    parser.add_argument('--exclude', metavar='REGEX', action='append', \
+                        help='Regex ( re.search() ) for exluding of directory watching. Can be specified multiple times. \
+                              Pathnames to test against are after os.path.normcase(), i.e. all lowercase with backslashes. \
+                              Example: --exclude \\\\build\\\\')
+    parser.add_argument('--disable_watching', action='store_true', help='Disable watching of directories which we have in the cache.')
+    args = parser.parse_args()
+
+    if args.exclude:
+        for pattern in vars(args)['exclude']:
```
frerich: Not sure I read this correctly, but why do you need to use `vars` here? It seems to me that `exclude` should be a list in itself already:

```python
>>> import argparse
>>> p = argparse.ArgumentParser()
>>> p.add_argument('--exclude', action='append')
_AppendAction(option_strings=['--exclude'], dest='exclude', nargs=None, const=None, default=None, type=None, choices=None, help=None, metavar=None)
>>> args = p.parse_args('--exclude this --exclude that --exclude something_else'.split())
>>> type(args.exclude)
<class 'list'>
>>> args.exclude
['this', 'that', 'something_else']
```

...so I'd expect that one could just use

```python
for pattern in args.exclude or []:
    logging.info("Not watching paths which match: %s", pattern)
```

server/clcachesrv.py
akleber: nice thanks

server/clcachesrv.py
akleber: yup, sorry

server/clcachesrv.py
akleber: `or` is becoming my friend.....

server/clcachesrv.py
akleber: agreed

frerich:
Looks good, merged. Thanks!

PR #274, opened 2017-05-03
Make Pylint 1.7.1 happy

frerich:
It appears that pip now installs Pylint 1.7.1 on AppVeyor, which introduced a few additional checks causing PRs to get marked as failures. This PR attempts to make the code base a bit better by either silencing Pylint warnings I do not agree with (or which yield false positives), or by actually patching the code.


frerich:
The remaining issue reported for `clcachesrv` is caused by what seems to be a bug in pylint: it does not recognise members of the `pyuv` module correctly (see PyCQA/pylint#1467).

Issue #276, opened 2017-05-23
Memcache Strategy should be independent


hubx:
As mentioned in #268 
- Make file strategy and memcache strategy independent. 
- change how `set/getEntry` work. ATM the local cache is used as intermediary, so memcache always 'pollutes' the local cache 
`        cachedArtifacts = cache.getEntry(cachekey)  # memcache puts objectFile into local cache
         copyOrLink(cachedArtifacts.objectFilePath, objectFile)  # copy local cache to build folder`

```python
class CacheMemcacheStrategy(object):
..

   def connect(self, server):
..
# XX key_prefix ties fileStrategy cache to memcache entry
# because tests currently the integration tests use this to start with clean cache
# Prevents from having cache hits in when code base is in different locations
# adding code to production just for testing purposes`
..
    def getEntry(self, key):
..
# XX this is writing the remote objectfile into the local cache
# because the current cache lookup assumes that getEntry gives us an Entry in local cache
# so it can copy it to the build destination later```

PR #278, opened 2017-05-23
Move storage classes to storage module

hubx:
While thinking about #276 I thought I makes sense that all storage related classes go into the same module. Let me know what you think

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -500,31 +500,6 @@ class PersistentJSONDict(object):
         return type(self) is type(other) and self.__dict__ == other.__dict__
 
 
-class Configuration(object):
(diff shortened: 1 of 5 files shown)


frerich:
I think moving the code related to (re-)storing cached data into a single file makes a lot of sense.

frerich:
@hubx For what it's worth, I just [did a release](https://github.com/frerich/clcache/releases/tag/v4.1.0) which ships your memcached support - thanks a lot! I hope you don't mind me giving you some public fame. :-)

hubx:
@frerich sure, no problem. Looking forward to user feedback :)

PR #280, opened 2017-07-25
Create setup.py

xoviat:
This will allow installation using pip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,20 @@
+#!/usr/bin/env python
+
+import os
+
+from setuptools import setup, find_packages
+
+setup(
+    name='clcache',
+    description='MSVC Complier cache',
+    author='frerich',
+    author_email='frerich@users.noreply.github.com',
+    url='https://github.com/frerich/clcache',
+    packages=find_packages(),
+    platforms='any',
+    keywords=[],
+    install_requires=['pypiwin32'],
+    setup_requires=[
+        'setuptools_scm',
+    ],
+    use_scm_version=True)
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,8 @@ from setuptools import setup, find_packages
 setup(
     name='clcache',
     description='MSVC Complier cache',
-    author='frerich',
-    author_email='frerich@users.noreply.github.com',
+    author='Frerich Raabe',
+    author_email='raabe@froglogic.com',
     url='https://github.com/frerich/clcache',
     packages=find_packages(),
     platforms='any',
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='clcache',
-    description='MSVC Complier cache',
+    description='MSVC compiler cache',
     author='Frerich Raabe',
     author_email='raabe@froglogic.com',
     url='https://github.com/frerich/clcache',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -0,0 +1,26 @@
...
+
+import os
+
+from setuptools import setup, find_packages
+
+setup(
+    name='clcache',
+    description='MSVC Complier cache',
+    author='frerich',
+    author_email='frerich@users.noreply.github.com',
```
frerich: Please use the same email address given in the README file, i.e. raabe@froglogic.com

setup.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+
+import os
+
+from setuptools import setup, find_packages
+
+setup(
+    name='clcache',
+    description='MSVC Complier cache',
+    author='frerich',
```
frerich: The [examples](https://docs.python.org/3/distutils/setupscript.html) seem to use the full name here for the author, I think the code should do the same.

frerich:
Please merge the two commits, I think there's not a lot of benefit in having two commits.

Some CI tests for this would be needed to make sure that 'pip install' really works, this would probably mean augmenting the appveyor.yml file.

xoviat:
> Please merge the two commits, I think there's not a lot of benefit in having two commits.

You should be able to use the 'squash' method on merging.

> 'Some CI tests for this would be needed to make sure that 'pip install' really works, this would probably mean augmenting the appveyor.yml file.'

I don't think so as this is pretty well-defined behavior.

frerich:
@xoviat Even if 'pip install' is well-defined behaviour, it would still be worthwhile to check that the `setup.py` file has the desired effect. In particular, it also serves as documentation for how to use the file. I for one never wrote such a `setup.py` file before and wouldn't even know how to use it, so seeing it in action would be good in case I document it in the Wiki and then somebody comes around and says it 'Does not work'. :-)

frerich:
@xoviat Thanks for reminding me of the 'squash' merge option. I heard of that but never tried it myself. It sounds as if it's exactly what I wanted. :-)

setup.py
```diff
@@ -7,8 +7,8 @@
 setup(
     name='clcache',
     description='MSVC Complier cache',
```
frerich: There's a typo here ("Complier") and I suppose "compiler" should be written all lower-case.

setup.py
```diff
@@ -7,8 +7,8 @@
...
-    author='frerich',
-    author_email='frerich@users.noreply.github.com',
+    author='Frerich Raabe',
+    author_email='raabe@froglogic.com',
```
frerich: 👍

xoviat:
You can see the install works from https://ci.appveyor.com/project/frerich/clcache/build/1.0.1065/job/xdow19mlnqqwmg9j#L136

frerich:
@xoviat Thanks! So

```
pip install .
```

Would be how to install clcache via pip then? Does this build an executable along the way? I'm surprised that running `clcache -help` just like that works, I thought 'pip' would just copy the script somewhere.

xoviat:
No. That command is because the git repository is already cloned. For a typical installation:

```
pip install git+https://github.com/frerich/clcache.git
```

>  Does this build an executable along the way?
Yes, but it is not a self-contained executable. It is dependent on the local python interpreter.

xoviat:
@frerich Are there any barriers to merging this?

frerich:
I just experimented a bit, and it seems to work for me. Thanks!

PR #285, opened 2017-09-18
setup: remove pypiwin32 dependency

xoviat:
pypiwin32 is quite a large dependency that isn't needed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -127,7 +127,7 @@ on_finish:
 
 on_success:
   # Creates executable with PyInstaller
-  - pip install pyinstaller
+  - pip install pyinstaller pypiwin32
   - pyinstaller clcache.py
   - for /F %%i in ('python -c "import clcache; print(clcache.VERSION)"') do set CLCACHE_VERSION=%%i
   - "echo Version: %CLCACHE_VERSION"
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,6 @@ setup(
               'clcache = clcache:main'
           ]
     },
-    install_requires=['pypiwin32'],
     setup_requires=[
         'setuptools_scm',
     ],


frerich:
Hm, looking at the [Python 3.3 build log](https://ci.appveyor.com/project/frerich/clcache/build/1.0.1069/job/f2au2cfyfmc9ksbc#L347), it sounds like pypiwin32 is needed for the `pywintypes` module -- but only for Python 3.3?

rasjani:
In same cases,  pywintypes*.dll might be imported into system path and results look something like this: works somewhere (dependency is there but its installed by someone else) and doesnt work somewhere else ..

xoviat:
This is a problem with pyinstaller. The setup.py is not the correct place to put it.

frerich:
Ah, I see. However, the [PyInstaller installation instructions](http://pyinstaller.readthedocs.io/en/stable/installation.html) say

> For Windows, PyWin32 or the more recent pypiwin32, is a prerequisite. The latter is installed automatically when you install PyInstaller using pip or easy_install.

So I take your patch (which makes the build work) suggests that the installation instructions are wrong? Or is there something I'm missing?

xoviat:
This is a bug in pyinstaller. I have worked extensively on that project.

xoviat:
Honestly though, I have moved on from the PyInstaller project because I
came to realize that the entire approach was wrong. ​PyInstaller is an
awesome and intricate beast of a project that simulates dynamic import
machinery and covers almost all of the cases that most users will see. It's
also a massive waste of time: it has massive exposure to implementation
details that constantly break, the test suite takes literally hours, and it
solves a problem (packaging) that is already solved with pip. You're almost
always better off running the Python installer (which doesn't require
administrative privileges) and then running "pip install" either from pypi
or prepackaged wheels.

frerich:
I see. Well, I guess it doesn't hurt to mention this dependency here, so I'll merge this patch. Thanks a lot for cleaning this up! 👍

PR #287, opened 2017-10-06
ENH: monkey: patch distsutils if USE_CLCACHE is set

### LINKED ISSUE #284: Feature request: monkeypatch distutils to use clcache
clcache should provide a module that when imported, monkeypatches distutils c compiler to use clcache.

xoviat:
The goal here is to patch distutils to use clcache if the environment variable is set. This currently needs a test that I will add shortly.

Closes #284.


clcache_monkey.py
```diff
@@ -0,0 +1,30 @@
...
+
+from typing import List
+
+
+def patch_distutils() -> None:
+    from distutils import ccompiler
+    
+    ccompiler_spawn = ccompiler.CCompiler.spawn
+    def msvc_compiler_spawn(self, args: List[str]) -> None:
+        if args[0] != self.cc:
```
xoviat: A major assumption here is that clcache is run on windows (eg. if this is run on linux, clcache will still be used because this doesn't patch msvccompiler, which isn't public).

xoviat:
Unmarking as WIP because this mostly done except for errors that I haven't caught. Appveyor failed for a different reason, but feels like a separate PR.

xoviat:
Follow up to appveyor failure is #288.

xoviat:
@frerich Okay to merge.

clcache/monkey.py
```diff
@@ -0,0 +1,59 @@
...
+
+    with suppress(ImportError):
+        from numpy.distutils import ccompiler
+
+    from distutils import ccompiler
+    
+    # Make some reasonable assumptions about the Python directory structure
+    clcache_path = None
+    if True:
+        clcache_path = join(dirname(sys.executable), 'Scripts', 'clcache.exe')
```
frerich: I think there's no reason anymore to keep lines 20 and 21.

xoviat: It looks nice, but will revert.

xoviat:
```
clcache\monkey.py:17: error: Name 'ccompiler' already defined
clcache\monkey.py:24: error: Too many arguments
clcache\monkey.py:27: error: Too many arguments
clcache\monkey.py:31: error: Too many arguments
clcache\monkey.py:38: error: Too many arguments
clcache\monkey.py:40: error: Cannot assign to a method
clcache\monkey.py:40: error: Incompatible types in assignment (expression has type "Callable[[Any, List[str], VarArg(Any), KwArg(Any)], None]", variable has type "Callable[[CCompiler, List[str]], None]")
```

xoviat:
@frerich Ping.

frerich:
@xoviat Thanks for pinging me, I'm currently a bit swamped. I just looked at this code again, and as far as I'm concerned, it's good to go - thanks a lot for your effort (and thanks for the test!). I'll merge this now.

Just one thing I'd like to ask for: can you maybe add a section to the [Integration](https://github.com/frerich/clcache/wiki/Integration) wiki page which explains how to enable clcache when using distutils?

Issue #288, opened 2017-10-06
all modules not included in setup.py


xoviat:
Appveyor failure is the following:
```

python : Traceback (most recent call last):
At line:3 char:1
+ python setup.py build
+ ~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Traceback (most recent call last)::String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
  File "C:\Python33\Scripts\clcache-script.py", line 9, in <module>
    load_entry_point('clcache==4.1.1.dev25+g22457ee', 'console_scripts', 'clcache')()
  File "c:\python33\lib\site-packages\clcache.py", line 1508, in main
    cache = Cache()
  File "c:\python33\lib\site-packages\clcache.py", line 575, in __init__
    from storage import CacheFileWithMemcacheFallbackStrategy
ImportError: No module named 'storage'
```

The cause of this failure is that `storage` is not included in `setup.py`. However, the problem is that `storage` is too generic of a module name. This is normally handled by moving the modules into a dedicated package to avoid namespace conflicts.

However, that brings up another issue: you're (or others) probably just copying the clcache files into a folder on the PATH. This might not work any longer if we use a dedicated package, but the general method of redistributing code is to use setup.py / the standard package layout and `pip install`.

The question is: would you be open to moving to a package named `clcache` with a `__main__.py` file inside, along with the other modules?

frerich:
I don't mind (justified) reorganisations of the source code at all. I'd prefer keeping backwards compatibility (with existing setups) though, of course.

Would it be possible to keep a `clcache.py` wrapper script which merely imports the `clcache` package and invokes a main function?

xoviat:
@frerich I don't think so because it would have the same name as the package. We could rename the package though.

PR #290, opened 2017-10-07
BUG: fix source file duplication in command line

### LINKED ISSUE #289: clcache does not handle 'tc' argument correctly
After some investigation into why clcache does not work as a substitute for the distutils compiler (#287 is just to get distutils to use clcache, not make it actually work), I received the following interesting output: 

```
running build
running build_ext
building 'demo' extension
creating build
creating build\temp.win32-3.3
creating build\temp.win32-3.3\Release
C:\Python33\Scripts\clcache.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -IC:\Python33\include -IC:\Python33\include /Tcexample.c /Fobuild\temp.win32-3.3\Release\example.obj
C:\Python33\Scripts\clcache.py Found real compiler binary at 'c:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\BIN\cl.exe'
C:\Python33\Scripts\clcache.py Arguments we care about: '['C:\\Python33\\Scripts\\clcache', '/c', '/nologo', '/Ox', '/MD', '/W3', '/GS-', '/DNDEBUG', '-IC:\\Python33\\include', '-IC:\\Python33\\include', '/Tcexample.c', '/Fobuild\\temp.win32-3.3\\Release\\example.obj']'
python : cl : Command line warning D9026 : options apply to entire command line
At line:4 char:1
+ python setup.py build
+ ~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (cl : Command li...re command line:String) [], RemoteException
 
cl : Command line error D8036 : '/Fobuild\temp.win32-3.3\Release\example.obj' not allowed with multiple source files
C:\Python33\Scripts\clcache.py Parsing given commandline '['/c', '/nologo', '/Ox', '/MD', '/W3', '/GS-', '/DNDEBUG', '-IC:\\Python33\\include', '-IC:\\Python33\\include', '/Tcexample.c', '/Fobuild\\temp.win32-3.3\\Release\\example.obj']'
C:\Python33\Scripts\clcache.py Expanded commandline '['/c', '/nologo', '/Ox', '/MD', '/W3', '/GS-', '/DNDEBUG', '-IC:\\Python33\\include', '-IC:\\Python33\\include', '/Tcexample.c', '/Fobuild\\temp.win32-3.3\\Release\\example.obj']'
C:\Python33\Scripts\clcache.py Compiler source files: ['example.c']
C:\Python33\Scripts\clcache.py Compiler object file: ['build\\temp.win32-3.3\\Release\\example.obj']
C:\Python33\Scripts\clcache.py Invoking real compiler as ['c:\\Program Files (x86)\\Microsoft Visual Studio 10.0\\VC\\BIN\\cl.exe', '/showIncludes', '/c', '/nologo', '/Ox', '/MD', '/W3', '/GS-', '/DNDEBUG', '-IC:\\Python33\\include', '-IC:\\Python33\\include', '/Tcexample.c', '/Fobuild\\temp.win32-3.3\\Release\\example.obj', 'example.c']
C:\Python33\Scripts\clcache.py Real compiler returned code 2
C:\Python33\Scripts\clcache.py Finished. Exit code 2
error: command 'clcache' failed with exit status 2
```

What's essentially going on here is that clcache is passing `example.c` twice: once as `/Tcexample.c` and once as `example.c`. Unfortunately, this is not compatible with `/Fo`.

Solution: don't pass the last argument (`example.c`).

xoviat:
Previously, if the /Tp or /Tc option were passed, clcache would
duplicate the source file in the command line, making MSVC
think that it was compiling two separate files (/Tcexample.c 
and example.c), when it was really compiling one.

The problem with this was that MSVC only allows some
options if it is compiling one file, which caused some
invocations of clcache to fail.

Closes #289.


xoviat:
Test failure seems unrelated.

clcache.py
```diff
@@ -1293,6 +1293,12 @@ def analyze(cmdline):
 
 def invokeRealCompiler(compilerBinary, cmdLine, captureOutput=False, outputAsString=True, environment=None):
     realCmdline = [compilerBinary] + cmdLine
+
+    sourceFile = realCmdline[-1]
+    if '/Tc' + sourceFile in realCmdline or '/Tp' + sourceFile in realCmdline:
+        printTraceStatement("Removing last argument becuase of /Tc (Issue #289)")
```
frerich: Small typo here (`becuase`) and I guess `/Tc` is imprecise because it seems this could also be because of `/Tp`.

clcache.py
```diff
@@ -1293,6 +1293,12 @@ def analyze(cmdline):
...
+        realCmdline = realCmdline[:-1]
```
frerich: Is it really safe to assume that the last argument is always a source file argument? I wonder whether maybe this fix should be done on the caller side - but it's not clear to me on first glance where clcache would implicitly add a source file argument to the command line.

clcache.py
xoviat: I'm going to ignore this for now because it's not relevant to the discussion.

clcache.py
xoviat: > Is it really safe to assume that the last argument is always a source file argument?

No. This is an example that fixes the problem but I had hoped that you would have a suggestion about a better fix.

> I wonder whether maybe this fix should be done on the caller side

Yes. Do you have any pointers for where to look?

xoviat:
@frerich Found it (I think):

https://github.com/xoviat/clcache/blob/f0227bb98a16a9804e2b71ef2a6d100c1c7b2719/clcache.py#L1633

Would it be okay to move the `if` statement there?

```
if '/Tc' + sourceFile in realCmdline or '/Tp' + sourceFile in realCmdline:
```

frerich:
@xoviat Good catch, that place indeed looks like it might end up constructing incorrect command lines like the one you noticed. I think an `if` statement along the lines of what you propose would be much more appropriate in `scheduleJobs` since that's where the issue is actually introduced right now.

However, maybe it would be even nicer to not use an `if` but rather reuse the knowledge about which source files are specified via `/Tp` resp. `/Tc`: the program [already knows this in `CommandLineAnalyzer.analyze`](https://github.com/xoviat/clcache/blob/f0227bb98a16a9804e2b71ef2a6d100c1c7b2719/clcache.py#L1248) but discards that knowledge. The function only yields the list of source files but does not track which language they use. This knowledge could be used in `scheduleJobs` to [construct a `baseCmdLine`](https://github.com/xoviat/clcache/blob/f0227bb98a16a9804e2b71ef2a6d100c1c7b2719/clcache.py#L1622) value which not only omits all 'plain' source files but also those specified via `/Tc` and `/Tp`.

xoviat:
Test failure is due to "too many local variables." Not sure what to do about that.

frerich:
The typical fix for the "too many local variables" error is, well, to not use as many local variables. :-) Instead of removing variables though, it's usually better to rather introduce a new function (i.e. the check is a kind of code complexity measure).

However, I must admit that I like the older version with the `if` better than the new one. My idea triggering the suggestion to reuse the knowledge from `CommandLineAnalyzer.analyze` was that there are actually two issues right now:

1. The code which [constructs the `baseCmdLine`](https://github.com/xoviat/clcache/blob/f0227bb98a16a9804e2b71ef2a6d100c1c7b2719/clcache.py#L1622) is defective: it attempts to filter all source files from the command line (such that lateron it's safe to add a source file to the command line) but fails to do so for source files which were specified via `/Tcxxx` or `/Tpxxx`. Furthermore, if an optional space is used (e.g. `/Tc xxx` or `/Tp xxx`), I believe the filtered command is even incorrect since e.g. `cl /Tp foo.cpp` will get turned into `cl /Tc foo.cpp` if I read correctly.

2. The code which [builds the `jobCmdLine`](https://github.com/xoviat/clcache/blob/f0227bb98a16a9804e2b71ef2a6d100c1c7b2719/clcache.py#L1633) is defective, too: it fails to consider any explicitly specified language (i.e. `/Tc` or `/Tp`) for the source file appended to the command line. So when invoking clcache like e.g. `cl /Tc a.x /Tp b.c` I suspect that the `jobCmdLine` values will compiler will be invoked like `cl a.x` and `cl b.c`, the former probably triggering an error and the latter (incorrectly) deducing C (instead of the specified C++) from the file extension `.c`.

My hope was that it might be viable to change `CommandLineAnalyzer.analyze` such that instead of a plain list of source files, it also returns an explicitly configured language (if any) for the source files. E.g. `cl a.cpp /Tpx.c /Tc foo.y` yields a list of tuples like `[('a.cpp', ''), ('x.c', 'Tp'), ('foo.y', 'Tc')]` or something like that. And this language information could then be used both when filtering the command line (to produce the `baseCmdLine` value) as well as when appending a single source file to the base command line (because by having the language, the code ensures to not append e.g. just `foo.y` but rather `/Tc foo.y`).

xoviat:
@frerich Before I fix up the unit tests, does that seem okay?

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
     printOutAndErr(out, err)
     return exitCode
 
+def formBaseCommandLine(cmdLine, sourceFiles):
+    # type: (List[str], List[Tuple[str, str]]) -> List[str]
+    setOfSources = set([sourceFile for sourceFile, _ in sourceFiles])
```
frerich: Minor cosmetic: you could shorten this by using a generator expression:

```python
setOfSources = set(sourceFile for sourceFile, _ in sourceFiles)
```

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
...
+    skippedArgs = ('/MP', '/Tc', '/Tp')
```
frerich: I _think_ that `-Tc` and `-Tp` work, too. At least we support either form in `CommandLineAnalyzer.analyze`, and with Visual Studio 2013 (the only compiler I have at hand right now) you can indeed use either form.

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
...
+    baseCmdLine = [
+        arg for arg in cmdLine
+        if not (arg in setOfSources or arg.startswith(skippedArgs))
+    ]
```
frerich: This list comprehension works for cases like `/Tcfoo.c` but it's also possible to have an optional(!) space, i.e. `/Tc foo.c`. In this case _both_ elements should get removed (since just `/Tc` on its own is a command line error).

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
...
+    # Filter out all source files from the command line to form baseCmdLine
+    baseCmdLine = formBaseCommandLine(cmdLine, sourceFiles)
 
     exitCode = 0
     cleanupRequired = False
     with concurrent.futures.ThreadPoolExecutor(max_workers=jobCount(cmdLine)) as executor:
         jobs = []
-        for srcFile, objFile in zip(sourceFiles, objectFiles):
-            jobCmdLine = baseCmdLine + [srcFile]
+        for srcFile, srcLanguage, objFile in zip(zip(*sourceFiles), objectFiles):
```
frerich: This double-`zip` is confusing to me. How about just

```python
for (srcFile, srcLanguage), objFile in zip(sourceFiles, objectFiles):
```

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
...
+            jobCmdLine = baseCmdLine + list(filter(None, [srcLanguage, srcFile]))
```
frerich: Since `srcLanguage` is either an empty string or one of `/Tp` or `/Tc`, couldn't you drop the composition of `list` and `filter` and just use

```python
jobCmdLine = baseCmdLine + [srcLanguage + srcFile]
```

clcache.py
```diff
@@ -1612,19 +1618,28 @@ def processCompileRequest(cache, compiler, args):
...
+    return baseCmdLine
+
 def scheduleJobs(cache, compiler, cmdLine, environment, sourceFiles, objectFiles):
-    baseCmdLine = []
-    setOfSources = set(sourceFiles)
-    for arg in cmdLine:
-        if not (arg in setOfSources or arg.startswith("/MP")):
-            baseCmdLine.append(arg)
+    # type: (???, str, List[str], ???, List[Tuple[str, str]], List[str]) -> int
+    # Filter out all source files from the command line to form baseCmdLine
```
frerich: I think this comment is misleading: the `formBaseCommandLine` function does not only filter out all source files, it also filters out `/MP`. However, I agree that the function _should_ filter out just the source files because then it could get the more descriptive name `filterSourceFiles` and the `/MP` could be filtered on the caller side:

```
baseCmdLine = [arg for arg in filterSourceFiles(cmdLine, sourceFiles) if arg != '/MP']
```

clcache.py
xoviat: Ahh, but in the second case, it would be in `setOfSources`.

clcache.py
xoviat: Will do.

clcache.py
xoviat: Will do.

clcache.py
xoviat: Will do.

xoviat:
@frerich Third round of concerns addressed.

frerich:
Thanks for your persistence! 👍 I think by now, the patch looks plausible and conceptually consistent. I believe that after fixing up the unit tests (and maybe adding four, five small tests for `filterSourceFiles`) this stuff is ready to get merged. It technically fixes two issues by now: it avoids constructing invalid command lines in which a source file is mentioned twice, and it ensures that when compiling in parallel, each source file gets the correct input language specified.

unittests.py
```diff
@@ -505,7 +505,7 @@ def _testFailure(self, cmdLine, expectedExceptionClass):
 
     def _testFull(self, cmdLine, expectedSourceFiles, expectedOutputFile):
         sourceFiles, outputFile = CommandLineAnalyzer.analyze(cmdLine)
-        self.assertEqual(sourceFiles, expectedSourceFiles)
+        self.assertEqual([s for s, _ in sourceFiles], expectedSourceFiles)
```
frerich: Wouldn't it be nicer to rather adjust the caller of `_testFull` such that `expectedSourceFiles` is actually a list of tuples, specifying the expected language (if any) specification for every source file?

xoviat:
Unfortunately, one of the integration tests failed. I'm not sure I understand this failure but the cache statistics may be omitting the fully specified source file arguments. I looked at `numCacheHits` and could not find how it was calculated..

xoviat:
In fact clcache appears not to reuse the cache files. I looked at this some more and unfortunately it appears to be a bit too dense for me.

clcache.py
```diff
@@ -1243,15 +1246,21 @@ def parseArgumentsAndInputFiles(cmdline):
...
-            inputFiles += options['Tp']
+            inputFiles.update({inputFile: '/Tp' for inputFile in options['Tp']})
             compl = True
         if 'Tc' in options:
-            inputFiles += options['Tc']
+            inputFiles.update({inputFile: '/Tc' for inputFile in options['Tc']})
             compl = True
 
+        # Now collect the inputFiles into the return format
+        inputFiles = list(inputFiles.items())
```
siu: Could it be that the failure in tests is introduced here because inputFiles.items() does not always return the items in the same order? either sorting them or using OrderedDict in line 1253 could help.

xoviat: I will try that.

xoviat:
@siu That still did not fix the problem. There are still zero cache hits on the concurrent test.

clcache.py
```diff
@@ -1612,19 +1621,26 @@ def processCompileRequest(cache, compiler, args):
...
+
 def scheduleJobs(cache, compiler, cmdLine, environment, sourceFiles, objectFiles):
-    baseCmdLine = []
-    setOfSources = set(sourceFiles)
-    for arg in cmdLine:
-        if not (arg in setOfSources or arg.startswith("/MP")):
-            baseCmdLine.append(arg)
+    # type: (???, str, List[str], ???, List[Tuple[str, str]], List[str]) -> int
+    # Filter out all source files from the command line to form baseCmdLine
+    baseCmdLine = [arg for arg in filterSourceFiles(cmdLine, sourceFiles) if arg != '/MP']
```
frerich: The `arg != /MP` test is what breaks the `TestRunParallel.testHitsViaMpConcurrent` test.

The test first compiles two source files individually using the command lines

1. `/nologo /EHsc /c fibonacci01.cpp`
2. `/nologo /EHsc /c fibonacci02.cpp`

The test then tries to compile both files again via

* `/nologo /EHsc /c /MP2 fibonacci01.cpp fibonacci02.cpp`

...expecting that this triggers two cache hits. This causes clcache to reinvoke itself recursively for each of the source files. However, due to the `arg != '/MP'` test, the recursive invocations still include the `/MP2` flag, i.e. the two nested instances use the command lines

1. `/nologo /EHsc /c /MP2 fibonacci01.cpp`
2. `/nologo /EHsc /c /MP2 fibonacci02.cpp`

Thus, the command line are different than they were when the files were originally compiled, hence the hash sum for the cache entry is different and hence there is no cache hit.

It seems this can be fixed by using `not args.startswith('/MP')` instead of `arg != '/MP'`.

clcache.py
siu: Even if it didn't fix the tests I think it still makes sense to keep the OrderedDict. The output of analyze() should return the same order across executions 1) to keep the tests passing and 2) to generate the same hash for the compilation task (if it is computed after, I am not sure about this). I don't know enough about this part of the code, to be checked with @frerich.

xoviat: The test failures have nothing to do with the order, and if I understand correctly, the hash is calculated per-file ("This causes clcache to reinvoke itself recursively for each of the source files"), making the order irrelevant. There is no reason to tell Python to preserve the order when it doesn't matter.

xoviat:
Okay, this should finally be ready to merge.

frerich:
Looks good, thanks! Just two questions which cam to my mind during a last review:

1. I see you're adding 'type specifiers' like `# type: (str) -> None` in a couple of places. Are these processed by some sort of static analysis tool? For this PR, I'd like to keep them out - but if they are processed by a tool then a separate PR might be nice which also etends the CI system such that it runs the tool.

2. You modified the `printTraceStatement` function along the way -- did you mean to include this change in this PR? It might be useful (I personally don't really need the line numbers, but I guess they also don't hurt) but a bit out of scope for this PR.

xoviat:
> You modified the printTraceStatement function along the way -- did you mean to include this change in this PR? It might be useful (I personally don't really need the line numbers, but I guess they also don't hurt) but a bit out of scope for this PR.

Done.

> I see you're adding 'type specifiers' like `# type: (str) -> None` in a couple of places. Are these processed by some sort of static analysis tool? For this PR, I'd like to keep them out - but if they are processed by a tool then a separate PR might be nice which also etends the CI system such that it runs the tool.

First, these type specifiers are the pet project of the BDFL himself (see mypy). I won't provide an in-depth explanation here, but the short version is that most developers spend time reading code to understand what is happening (as I had to do here to repair the bug, and you had to do to review my changes), and *the biggest problem with Python* is that you have to figure out what the types are of the function arguments to understand what is happening. Static typing solves that problem: it's not for computers, it's for people.

> For this PR, I'd like to keep them out

I'm going to disagree here. Of course, at the end of the day, it's your project, so if you insist, I will remove them. But what I will say is that the typing is done per-file, and it won't be useful until most of the file has it. That means that you need to build the typing over time (for example, the question marks `???` aren't even valid, but I did not know the type, so I put them there so say that) until you have something reasonable that mypy can check (you don't even need to use mypy because as I said, the type annotations are for people, but it's there as an option). Every time to work on a function, you have the typing knowledge in your mind, so you might as well spend an extra few seconds to write it down. But if you're trying to type the entire file in one go, then you don't have that knowledge, and therefore you need to learn the types by checking the caller (which is what is done without type annotations), and therefore it's not just "an extra few seconds."

frerich:
> First, these type specifiers are the pet project of the BDFL himself (see mypy). I won't provide an in-depth explanation here, but the short version is that most developers spend time reading code to understand what is happening (as I had to do here to repair the bug, and you had to do to review my changes), and the biggest problem with Python is that you have to figure out what the types are of the function arguments to understand what is happening. Static typing solves that problem: it's not for computers, it's for people.

As someone who greatly enjoys writing programs in Haskell in his free time and C++ for a living, I very much agree - the lack of static typing is one of my main griefs with Python. Thanks for mentioning mypy, I wasn't even aware of that effort - it sounds exciting! I'd love to learn how expressive it is.

After looking at the [mypy homepage](http://mypy-lang.org/), it seems to me that it's built on top of type hints as mentioned in [PEP 483](https://www.python.org/dev/peps/pep-0483/), [PEP 484](https://www.python.org/dev/peps/pep-0484/) and variable annotations as described in [PEP 526](https://www.python.org/dev/peps/pep-0526/). The [typing module](https://docs.python.org/3/library/typing.html) seems to implement this as of Python 3.5. Does that sounds accurate?

If so, what do you think about omitting the comments you wrote from the PR for now and I'll instead raise the minimum supported Python version to 3.5 such that we can use 'real' type annotations as understood by mypy (and we can then also introduce mypy to the CI builds)? My impression is that the comments are a bit of a cludge due to the fact that we currently have to support Python 3.3. Or is there maybe a way to get type hints prior to Python 3.5?

I greatly appreciate you contributing this type information, it just seems to me that 1.) the type hints are a bit 'out of scope' for this particular PR which attempts to address a very specific bug (#289) and 2.) my current understanding is that we could have 'real' type hints (not in a comment) which would require dropping support for Python < 3.5. What do you think?

xoviat:
> I greatly appreciate you contributing this type information, it just seems to me that 1.) the type hints are a bit 'out of scope' for this particular PR which attempts to address a very specific bug (#289) and 2.) my current understanding is that we could have 'real' type hints (not in a comment) which would require dropping support for Python < 3.5. What do you think?

There's no need to drop support for Python 3.5, as Python 3 in general supports static typing. There is a need to "fix" the package format (#288) before switching to the nicer inline type annotations, but removing the comments will just mean that someone has to re-discover the types the next time around. And the comments are real type annotations for Python 2 that are fully supported by mypy (just not the question marks: `???`).

frerich:
Ah, I just skimmed the [mypy examples](http://mypy-lang.org/examples.html) and noticed that mypy seems to recognise _both_ comments as in `d = {}  # type: Dict[str, int]` as well as what I thought were called type hints in `def overdrawn(self) -> bool`. And in [PEP 483](https://www.python.org/dev/peps/pep-0483/), the [pragmatics](https://www.python.org/dev/peps/pep-0483/#pragmatics) section says

> Type declaration in comments, e.g.:
> 
> lst = []  # type: Sequence[int]

...so that would kinda answer my question: this `# type: ...` notation is not something you made up yourself but it's actually a convention recognised by tools? Nice!

frerich:
The [Fundamental building blocks](https://www.python.org/dev/peps/pep-0483/#fundamental-building-blocks) section says that

> Every type is consistent with Any; and it is also consistent with every type (see above).

...so instead of `???` you could maybe use `Any`?

xoviat:
Will do.

frerich:
For the record, there's a dedicated discussion of [type comments](https://www.python.org/dev/peps/pep-0484/#type-comments) as one way to articulate type hints.

xoviat:
Updated.

xoviat:
Test failure appears unrelated.

frerich:
Yes, the test failure should be fixed in `master` already after merging @sui's PR #292 . I think we're good to go - thanks a lot for your persistence (and patience)! 👍

PR #292, opened 2017-10-11
Revert reading manifests and statistics without acquiring the lock

siu:
Bad news, one of the test runs in pull request #290 failed in a call to os.replace(), see https://ci.appveyor.com/project/frerich/clcache/build/1.0.1113/job/88r9w6ar5plwo7uc/tests. I believe that it failed because another process/thread had the file open for reading.

Unfortunately windows does not allow to replace a file that is opened
for reading by another process/thread. Reads and writes need to be
serialized by using the locks.

This reverts the locking of manifests and statistics to the state before
pull request #286.

### DIFF
--- a/clcache.py
+++ b/clcache.py
@@ -1379,7 +1379,7 @@ clcache statistics:
     called w/ multiple sources : {}
     called w/ PCH              : {}""".strip()
 
-    with cache.statistics as stats, cache.configuration as cfg:
+    with cache.statistics.lock, cache.statistics as stats, cache.configuration as cfg:
         print(template.format(
             str(cache),
             stats.currentCacheSize(),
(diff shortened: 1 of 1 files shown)


siu:
Hmmm some tests failed in the first commit in a way that I didn't expect. Something seems wrong in processNoDirect().

frerich:
I'm a little confused - the test run in #290 passes. Is there a race condition, i.e. is there an issue or not? :-)

siu:
This is all I have found until now:

Reading the files without locking is not a good idea because os.replace(...) writing to a file that is open will yield an error. This is a limitation of windows and we cannot do much about it. This is what the PR addresses, I think it is safe to merge.

There is another problem in processNoDirect that I don't understand yet. I created a branch in my fork of the repo and enabled AppVeyor to investigate https://github.com/siu/clcache/commits/revertLockFreeReadsAppveyor. The last commit makes some tests run 100 times in an attempt to trigger the race. These are the results: https://ci.appveyor.com/project/siu/clcache/build/1.0.14

What is surprising is that it only failed on Python 3.3, it could be luck or it could be something else. For the moment I could not see anything wrong in the code.

frerich:
Merged, thanks a lot for following-up so quickly! I'm somewhat embarrased that our tests did not notice this peculiar behaviour behaviour of `os.replace`. :-(

siu:
Same here, I wish the automatic tests were more intensive and showed synchronization issues early.

I will start a PR with [this commit](https://github.com/siu/clcache/commit/de862f73a1649006758987731a67e342b91c4fbf) to discuss about the issue in processNoDirect.

PR #295, opened 2017-10-12
Refactor into subpackage

xoviat:
Summary of changes:
- files are moved into a subpackage called 'clcache'
- test files are changes to the `test_*` in the `tests` folder, which is `pytest` convention
- appropriate installation requirements are added

In addition, setuptools will automatically detect the Python files with `find_packages`, so `py_modules` is removed.


xoviat:
I may need to revise if the CI comes back with a failure.

xoviat:
I cannot seem to generate the coverage report; I'll come back to this later.

PR #296, opened 2017-10-12
Replace whole cache entries atomically

siu:
This is a follow-up of #286. This time the cache entry directory is replaced atomically.

With these changes all the files written by clcache are replaced atomically on disk, preventing cache corruption when the clcache process is terminated in the middle of a write.


frerich:
Hmm, good stuff - I can't think of much when reviewing this code. I suppose it's hard to write a test for this, so I'll merge. Thanks!

PR #297, opened 2017-10-13
Refactor into subpackage

### LINKED ISSUE #288: all modules not included in setup.py

xoviat:
Summary of changes:
- files are moved into a subpackage called 'clcache'
- test files are changes to the `test_*` in the `tests` folder, which is `pytest` convention
- appropriate installation requirements are added

In addition, setuptools will automatically detect the Python files with `find_packages`, so `py_modules` is removed.

I would also recommend using the "squash and merge" option for this PR.

Closes #288.


appveyor.yml
```diff
@@ -95,22 +95,9 @@ test_script:
...
-      & coverage report
-      & coverage xml
-
-      $wc = New-Object 'System.Net.WebClient'
-      $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\integrationtests.xml))
-
-      & codecov --no-color -X gcov -F integrationtests -e PYTHON_INSTALL
-
-      if ($testsExitCode -ne 0) {exit $testsExitCode}
-
```
xoviat: This might look odd, but with `pytest` convention, we can run all of the tests at once with a single command because `pytest` already knows what the test files are called.

appveyor.yml
```diff
@@ -35,6 +35,7 @@ install:
 
   # Check pip version
   - pip --version
+  - python -m pip install --upgrade pip setuptools wheel
```
frerich: Might be nice to move this command one line up such that `pip --version` shows the new version, after pip was upgraded.

xoviat: Okay.

xoviat:
This is going to need a rebase because git is confused about the patch.

xoviat:
Need to rebase now because I can't tell what is going on.

xoviat:
@frerich I added the wrapper script for `clcachesrv.py`. As for the `clcache.py` wrapper script, I would recommend adding it in another PR because it makes the diff too large (git becomes confused).

xoviat:
@frerich Is this moving forward or not?

frerich:
I'm on vacation and only check my mail every now and then via my phone.
Will continue working on this in a couple of days.

Am 22.10.2017 10:37 nachm. schrieb "xoviat" <notifications@github.com>:

xoviat:
@frerich Ping.

frerich:
Merged, thanks!

PR #298, opened 2017-10-16
Remove signal handlers

siu:
The signal handlers were introduced to avoid cache corruption when the
clcache process is stopped in the middle of a write to the cache
(statistics, manifests or objects). See #233. Even if SIGINT and SIGTERM
were ignored the cache still had a chance to be corrupted in the event
of a SIGTERM (which cannot be ignored).

Since #233 the writing of files to the cache has been improved to
replace the files atomically, reducing the risk of storing corrupted
files in the cache. See pull requests #286, #292 and #296. Therefore
ignoring these signals is not needed anymore.

### DIFF
--- a/clcache/__main__.py
+++ b/clcache/__main__.py
@@ -20,7 +20,6 @@ import multiprocessing
 import os
 import pickle
 import re
-import signal
 import subprocess
 import sys
 import threading
@@ -1513,15 +1512,7 @@ def createManifestEntry(manifestHash, includePaths):
     return ManifestEntry(safeIncludes, includesContentHash, cachekey)
 
 
-def installSignalHandlers():
-    # Ignore Ctrl-C and SIGTERM signals to avoid corrupting the cache
-    signal.signal(signal.SIGINT, signal.SIG_IGN)
-    signal.signal(signal.SIGTERM, signal.SIG_IGN)
-
 def main():
-
-    installSignalHandlers()
-
     if len(sys.argv) == 2 and sys.argv[1] == "--help":
         print("""
 clcache.py v{}


siu:
@frerich I think it is safe to do this now but I don't know if it has any side effects when the clcachesrv or memcache is used.

frerich:
I *think* I agree that removing the signal handlers should be safe now, but I'll have to think about this some more (and remind myself of the discussions we had back when the handlers were introduced). Thanks for bringing this topic up and providing a patch!

siu:
Should close #263, #261.

frerich:
@siu Can you rebase your PR on current master? Thanks! I think that otherwise, this is safe to merge (after dropping Python 3.3 support).

siu:
Done, I have a question though, how does one run the tests locally after the refactor in modules? If I run ```py.test``` I get this error:

```
ImportError while importing test module 'C:\src\clcache\tests\test_integration.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests\test_integration.py:23: in <module>
    from clcache import __main__ as clcache
E   ImportError: No module named 'clcache'
```

xoviat:
@siu Just run `pip install -e .` before running the tests.

siu:
@xoviat, thank you, I will give it a try

siu:
@xoviat that worked. Does that mean that if I forget to run `pip install -e .` I will be running the tests against the clcache installed in the system?

xoviat:
`pip install -e .` installs your development checkout of clcache. That means if you edit any files, the changes will take immediate effect.

siu:
Cool, that makes a lot of sense, thanks for clarifying the workflow.

siu:
@frerich in my opinion this PR is ready, let me know if anything else is blocking the merge.

frerich:
Looks good - I suppose this closes #261.

siu:
#263 can be closed too.

PR #300, opened 2017-10-26
[clcache] allow specification of maxsize with environment variable

xoviat:


frerich:
When is this useful?

clcache.py
```diff
@@ -1530,8 +1530,11 @@ def main():
         print('Statistics reset')
         return 0
 
-    if len(sys.argv) == 3 and sys.argv[1] == "-M":
-        arg = sys.argv[2]
+    if (len(sys.argv) == 3 and sys.argv[1] == "-M") or "CLCACHE_MAXSIZE" in os.environ:
+        if "CLCACHE_MAXSIZE" in os.environ:
+            arg = os.environ["CLCACHE_MAXSIZE"]
+        else:
+            arg = sys.argv[2]
```
frerich: I find it strange that the environment variable overrides the command line argument. I think it's more common to have it the other way round, isn't it?

xoviat:
Actually this isn't useful. I worked around it.

PR #304, opened 2017-10-31
Find compiler: apply shutil.which to basenames

xoviat:
This handles the case where CLCACHE_CL is set to a basename rather than a full path. This is useful for setting cl to a different binary, like `clang-cl`.


frerich:
Looks good, merged. Thanks!

PR #307, opened 2018-01-19
Remove py2exe specific code

siu:
I think this code is not needed anymore since py2exe is not the recommended way
to build the package.


siu:
The appveyor tests are failing but I don't think it is related to the code changes.

jimbo1qaz:
Nope, PyInstaller reuses the same "frozen" tag. The comments should reflect that. https://pyinstaller.readthedocs.io/en/stable/runtime-information.html

siu:
Thanks for the tip. Closing as invalid.

PR #318, opened 2018-07-25
 Drop version in directory structure of chocolatey (nuget) target package

formiaczek:
Because the tool is used by some MSBuild automation / scripts, having a version in the 'target' directory is not very useful.

For example:
' nuget.exe install clcache -excludeversion -outputdirectory d:\'

Would install clcache binary is located at:

' d:\clcache\clcache-4.4.1\clcache.exe '

This is not very useful as one needs to detect this version, and having it as:

' d:\clcache\clcache\clcache.exe '

Would make more sense.
Note that dropping the '-excludeversion' option would still install the clcache as expected:
' d:\clcache-4.4.1\clcache\clcache.exe '

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ clcache changelog
    entries in case it gets terminated in the middle of copying a file (GH #263).
  * Feature: Added 'monkey' script to simplify integrating clcache with distutils-based
    workflows (GH #284).
+ * Feature: Drop version in directory structure of chocolatey (nuget) target
+   package (GH #318).
 
 ## clcache 4.1.0 (2017-05-23)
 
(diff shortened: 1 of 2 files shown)


formiaczek:
Sorry, I can't edit the file other than using web-editing tool and it looks like it also updated / changed the line with the "-<!-- Do not remove this test for UTF-8:". :/

frerich:
@formiaczek This patch looks great, could you maybe also add an entry to the `CHANGELOG.md` file and rebase it on current master, such that the CI builds work? I'd like to merge this before doing a new release. :-)

formiaczek:
Sure @frerich, I can do that :+1:

formiaczek:
Thaks @frerich, have a look now, let me know if I missed anything and I'll fix if needed.

frerich:
Looks great, merged. Thanks a lot for your contribution (and your patience) :-)

formiaczek:
no problem, thanks too for a great tool and effort in maintaining it!

PR #321, opened 2018-07-31
Fix upgrading pip/setuptools/wheel during AppVeyor CI builds

frerich:
The CI builds started failing a while ago, printing:

  pip install -U pip setuptools wheel
  ERROR: To modify pip, please run the following command:
  c:\python34\python.exe -m pip install -U pip setuptools wheel

I.e. it sounds as if one should not invoke 'pip' directly for some
reason but rather load it as a Python module. Looking at the YAML file,
it seems we already did that -- right after the problematic 'pip install
-U' call! So let's drop the direct 'pip' call in favor of the existing
'python -m pip ...' call to fix this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -54,8 +54,10 @@ install:
   # Install the package
   - pip install .
 
-  # Install tool for uploading coverage data to codecov.io
-  - pip install codecov
+  # Install tool for uploading coverage data to codecov.io; select version
+  # 2.0.10 specifically since later versions seem to have a bug due to
+  # which calling 'codecov' in a PowerShell script fails.
+  - pip install https://github.com/codecov/codecov-python/tarball/v2.0.10
   - coverage --version
 
   # Install dependencies for clcachesrv such that we can lint the source code
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,7 +27,7 @@ install:
   - set PATH=%PYTHON_INSTALL%;%PATH%
 
   # Prepend Python scripts to PATH (e.g. pip, py.test, pylint)
-  - set PATH=%PYTHON_INSTALL%\\Scripts;%PATH%
+  - set PATH=%PYTHON_INSTALL%\Scripts;%PATH%
 
   # Check Python version
   - python --version
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -96,7 +96,7 @@ test_script:
       $wc = New-Object 'System.Net.WebClient'
       $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\unittests.xml))
 
-      & codecov --no-color -X gcov -F unittests -e PYTHON_INSTALL
+      & codecov --no-color -X gcov -F unittests -e PYTHON_INSTALL -f coverage.xml
 
       if ($testsExitCode -ne 0) {exit $testsExitCode}
 
@@ -113,7 +113,7 @@ test_script:
       $wc = New-Object 'System.Net.WebClient'
       $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\integrationtests.xml))
 
-      & codecov --no-color -X gcov -F integrationtests_memcached -e PYTHON_INSTALL
+      & codecov --no-color -X gcov -F integrationtests_memcached -e PYTHON_INSTALL -f coverage.xml
 
       if ($testsExitCode -ne 0) {exit $testsExitCode}
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,10 +33,9 @@ install:
   - python --version
   
   # Upgrade the Python build tools
-  - pip install -U pip setuptools wheel
+  - python -m pip install --upgrade pip setuptools wheel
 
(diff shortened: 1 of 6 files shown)


formiaczek:
ouh, I didn't check other PRs.. and tried to addressed the same :) I think the complaint was about the pip installing on top of itself.. and in 2cc04bb  I thought that this pip install was not really needed after all, invoking it through the module will update pip itself without complaints :) I'll drop this commit / rebase once you've merged your fix :)

clcache/storage.py
```diff
@@ -156,7 +156,7 @@ def clean(self, stats, maximumSize):
                                 maximumSize)
 
 
-class CacheFileWithMemcacheFallbackStrategy(object):
```
formiaczek: wasn't worth keeping it, given you've disabled the warning in the other commit anyway? I guess apart from one line - the whole codebase was is still compatible with python 2.7 - but I guess on Windows it's never the case that there needs to be python 2.7 ?

formiaczek:
thanks for fixing this :+1:

frerich:
Alas, I don't understand why the `coverage` program can be executed as a plain shell command in the `.yml` file, but not in a PowerScript script. :-(

appveyor.yml
```diff
@@ -96,7 +97,7 @@ test_script:
       $wc = New-Object 'System.Net.WebClient'
       $wc.UploadFile("https://ci.appveyor.com/api/testresults/junit/$($env:APPVEYOR_JOB_ID)", (Resolve-Path .\unittests.xml))
 
-      & codecov --no-color -X gcov -F unittests -e PYTHON_INSTALL
+      & $($env:PYTHON_INSTALL)\Scripts\codecov.exe --no-color -X gcov -F unittests -e PYTHON_INSTALL
```
izmmisha: Probably it should looks like `${env:PYTHON_INSTALL}` instead `$($env:PYTHON_INSTALL)`
Also it should be changed at the line 117 
```
 & codecov --no-color -X gcov -F integrationtests_memcached -e PYTHON_INSTALL
```

frerich: I'll try that (I took the syntax from the line right above, where the URL to upload the results to is constructed)!

frerich: I also don't understand why it doesn't just pick up the `codecov` executable which is reachable in the` PATH` -- maybe the PowerShell script does not use the `PATH` the way I expect it to...

izmmisha: Ok, I'm able to reproduce this error on local machine. Problem is related to codecov:
```
==> Preparing upload
    Error running `cd 'C:\projects\clcache' && git ls-files`: None
```
It try to `Popen("cd 'C:\projects\clcache' && git ls-files", stdout=PIPE, shell=True)` and it fails with 'The filename, directory name, or volume label syntax is incorrect.'
Here is the code:
```
try:
    from shlex import quote
except ImportError: # pragma: no cover
from pipes import quote
[...]
 root = quote(root)
[...]
 toc = str((try_to_run('cd %s && git ls-files' % root) or
[...]
```
quote's result is `"'C:\\projects\\clcache'"`

izmmisha:
Ok, one more problems I discovered:
```
Codecov v2.0.15
==> Detecting CI provider
    AppVeyor Detected
    Fixing merge commit SHA
==> Preparing upload
    Error running `cd 'C:\projects\clcache' && git ls-files`: None
XX> Skip processing gcov
==> Collecting reports
XX> Searching for reports disabled
Error: No coverage report found
```
codecov can't find coverage report if `coverage.xml` is already exists, so need to remove `coverage xml` call, codecov will call it.

`codecov.exe : The filename, directory name, or volume label syntax is incorrect.` this exception probably can be avoided if redirect codecov stderr output to null (`2> $null`), but not sure if this helps.

izmmisha:
Well, maybe better would to move from codecov-python to codecov-exe, I'm tried it - results at 
https://ci.appveyor.com/project/izmmisha/clcache/build/1.0.3/job/qahj29815ccvi2jc

izmmisha:
@frerich the problem with
```
 codecov.exe : The filename, directory name, or volume label syntax is incorrect.
    At line:9 char:1
    + & codecov --no-color -X gcov -F unittests -e PYTHON_INSTALL
    + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        + CategoryInfo          : NotSpecified: (The filename, d...x is incorrect.:String) [], RemoteException
```
is related to codecov itself, here is my [comment](https://github.com/frerich/clcache/pull/321#discussion_r206811310) and issue https://github.com/codecov/codecov-python/issues/156

invocation of codecov is correct.

izmmisha:
Coverage report still not uploading via codecov: `Error: No coverage report found`.
according to the [code](https://github.com/codecov/codecov-python/blob/master/codecov/__init__.py#L675) 
```
            if os.path.exists(opj(os.getcwd(), '.coverage')) and not os.path.exists(opj(os.getcwd(), 'coverage.xml')):
                write('    Generating coverage xml reports for Python')
                # using `-i` to ignore "No source for code" error
                try_to_run('coverage xml -i')
                reports.append(read(opj(os.getcwd(), 'coverage.xml')))
```
`coverage.xml` will be append to reports if `coverage.xml` not exists and codecov will generate it by `coverage xml -i`

PR #322, opened 2018-08-01
Another try to fix AppVeyor CI builds

izmmisha:
* if coverage.xml exists codecov doesn't add it to reports, but if it not
exitst codecov will generate it

* redirect stderr of codecov to null, not sure is it helps, to avoid
 codecov bug https://github.com/codecov/codecov-python/issues/129


izmmisha:
Looks like AppVeyor is not start check if appveyor.yml changed :) So better to continue fixes in PR https://github.com/frerich/clcache/pull/321

Sorry.

PR #324, opened 2018-09-03
Normalize out `/[Ww]*` arguments

refack:
They should not affect the binary + cache hits will omit them anyway


refack:
@frerich thanks for the answer. This is a pain point for me ATM, as I'm trying to re-align `Node.js`'s warning exclusion list...
It's quite frustrating that these two are generate different hashes...
```cmd
cl /wd4351 /wd4355 /c foo.cpp
cl /wd4355 /wd4351 /c foo.cpp
```

Would you consider a new ENV var? Something like `CLCACHE_IGNORE_WARNINGS`?

P.S. `clcache` has been a godsend it brings down build time on my machine from 2h to ~7m (we have a ton of de-optimizations, e.g. we recreate the proj/make/ninja files almost on each build)

Issue #327, opened 2018-09-26
Option to compress cache content


benjaminfuchs:
Hi,
I was wondering if it is possible to add an option to compress the content stored in the cache like ccache has (look for CCACHE_COMPRESS here: https://ccache.samba.org/manual/latest.html).

Where in the code would be the entry point for such a feature?
If this is a good idea I would try to create a first implementation.

frerich:
#328 was just merged, so I guess this can be closed now.

PR #328, opened 2018-10-17
Adding feature to compress cache content

benjaminfuchs:
Pull request for issue #327 

I tried to implement the feature similar to ccache. It can be controlled by the following two env variables: 

CLCACHE_COMPRESS
    If true, clcache will compress object files it puts in the cache. If the cache
    was filled without compression it can't be used with compression and vice versa
    (i.e. you have to clear the cache when changing this setting). The default is false.

CLCACHE_COMPRESSLEVEL
    This setting determines the level at which clcache will compress object files.
    It only has effect if compression is enabled. The value defaults to 6, and
    must be no lower than 1 (fastest, worst compression) and no higher than 9
    (slowest, best compression).

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -71,6 +71,15 @@ CLCACHE_HARDLINK::
     will be created. This is more efficient (faster, and uses less disk space)
     but doesn't work if the cache directory is on a different drive than the
     build directory.
+CLCACHE_COMPRESS::
+    If true, clcache will compress object files it puts in the cache. If the cache
+    was filled without compression it can't be used with compression and vice versa
+    (i.e. you have to clear the cache when changing this setting). The default is false.
+CLCACHE_COMPRESSLEVEL::
+    This setting determines the level at which clcache will compress object files.
+    It only has effect if compression is enabled. The value defaults to 6, and
+    must be no lower than 1 (fastest, worst compression) and no higher than 9
+    (slowest, best compression).
 CLCACHE_NODIRECT::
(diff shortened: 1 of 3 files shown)


benjaminfuchs:
@frerich We are now using this dev version and it works stable in our env.

frerich:
@benjaminfuchs This looks interesting! Did you ever profile the effect of compressing the cache contents?

My gut reaction was that compressing cached files might be counter-productive, since it requires extra CPU time. However, maybe smaller files improve the disk cache hit rate, which may outweigh the time needed for (de-)compressing data?

Or is the main purpose maybe not to accelerate anything at all but rather to reduce the disk usage of the cache?

benjaminfuchs:
@frerich Thanks for the feedback! In our setup the cache size goes up to multiple GBs and  on our build servers IO throughput got the bottleneck. I wouldn't say the compression is  accelerating the build but it's  also not slowing it down.
We also use ccache with compression on. This way we can have multiple cache folders side by side for different builds and save lot's of disk space.
I would say it hardly depends on the use case if compression makes sense and in our case it helps a lot.

frerich:
Looks good to me, let's give it a try. Thanks!

pitrou:
I'm gonna try this on our AppVeyor builds. As a sidenote, it would be nice to make the compression scheme configurable. Using something like [zstd](https://pypi.org/project/zstandard/) should give higher compression ratios at the same (de)compression speed, or much larger speeds at similar compression ratio.

PR #335, opened 2019-01-04
Fixing pylint finding introduced by pylint 2.2

benjaminfuchs:
The unittest always run with the latest pylint version.
The latest version causes new findings in this file:
```
tests\test_unit.py:436:0: W1403: Implicit string concatenation found in list (implicit-str-concat-in-sequence)
tests\test_unit.py:437:0: W1403: Implicit string concatenation found in list (implicit-str-concat-in-sequence)
tests\test_unit.py:438:0: W1403: Implicit string concatenation found in list (implicit-str-concat-in-sequence)
tests\test_unit.py:439:0: W1403: Implicit string concatenation found in list (implicit-str-concat-in-sequence)
```
More details can be found here: http://pylint.pycqa.org/en/latest/whatsnew/2.2.html


frerich:
Thanks a lot for doing this grunt work - merged!

Issue #343, opened 2019-01-16
How to support build with two different path & arch compilers in parallel?


shijunjing:
Hello,
I have a problem to enable the clcache in my Uefi firmware build. My firmware build need both 32bit compiler and 64bit compiler, because some modules need run in 32bits at the begin and others need run in 64bits later on, and my firmware build tool will combine all these modules into one finial image. Not like gcc and clang, the MSVC has different path for these two compilers as below:
C:\Program Files (x86)\Microsoft Visual Studio 14.0\Vc\bin\cl.exe
C:\Program Files (x86)\Microsoft Visual Studio 14.0\Vc\bin\x86_amd64\cl.exe

If I replace the above two cl.exe with clcache_main.exe, how could I set the CLCACHE_CL to two real cl.exe compilers? Because my firmware build support multithread, so the modules build sequence is not predictable. How to correctly switch the CLCACHE_CL to real cl.exe?

My current thinking is to generate two different clcache_main.exe, which one is to map 32bit cl.exe and the other is to map 64bit cl.exe. However, how to easily let the clcache_main.exe know the correct compiler path?

YngveNPettersen:
This is probably the same issue as discussed in https://github.com/frerich/clcache/issues/270

Issue #270, opened 2017-02-27
Clcache should allow command line specification of compiler command


YngveNPettersen:
While having clcache map directly to cl might be useful for Visual Studio projects, in other cases, such as with the Chromium project which uses ninja project files generated by their GN system (and which have direct support for enabling ccache)  being able to specify the actual cl command on the command line would be better, especially in environments where one is both compiling with clcache and withou clcache.

frerich:
I'm not sure I understand: you can already specify the exact cl command line to invoke by setting the `CLCACHE_CL` environment variable to point to some specific compiler executable and then passing all arguments on the command line, no? Can you maybe give an example of what you mean?

YngveNPettersen:
I mean "clcache path/to/cl.exe params"

frerich:
I see - can you elaborate on when this would be useful? I'm not familiar with Chromium or Ninja really. What you describe seems to be equivalent to

```
set CLCACHE_CL=path/to/cl.exe
clcache params
```

on first glance. I suppose that's not sufficient for your use case?

YngveNPettersen:
It would be similar, but chromium specifically sets up a toolchain with all commands used by ninja, with fully qualified paths for the compiler and link executables. The GN code setting up the toolchain already have functionality to use ccache with the command path as the first positional argument, so it is easy to enable that functionality without having to change too much in the toolchain setup in our environment.

This also makes it easier to just flip the GN flag for enabling clcache for various product environments without having to perform "tricks" with the PATH variable.

frerich:
@YngveNPettersen After re-reading, I wonder how this would work exactly. I suppose the idea is that *if* the first argument to `clcache.py` is an existing file *and* that file ends in `.exe`, then we assume it's the path to the real compiler? I suppose that could work - I'm trying to think of counter-examples, but couldn't find anything convincing.

One, somewhat far-fetched, case I could think of: if you try to compile and link a source file ending in `.exe`, e.g. as in

```
cl foo.exe (or: clcache.py foo.exe)
```

Then you would get an error message about an invalid extension used for a source file. With the proposed semantics, the error message would differ in that we would print an error about not being able to execute the program 'foo.exe'.

benjaminfuchs:
I would also like this feature. We are already  using ccache with cmake via the CMAKE_C_COMPILER_LAUNCHER option and having this feature would make it easier to use clcache just like ccache in our build system.

benjaminfuchs:
[...]

Issue #269, opened 2017-02-27
Commandline parse should use argparse


YngveNPettersen:
The argparse Python module <https://docs.python.org/3/library/argparse.html> should be used to parse the clcache command line.

Using argparse allows for better parsing of the command line and more flexibility, and also allows some conditions, such as registering mutually exclusive arguments. It also automatically creates a --help message.

frerich:
I seem to recall someone tried that a few years ago (that was before Python 3 though) and it didn't actually help very much: the clcache command line API is very small and changes very rarely. Most of the command line handling is actually custom parsing code to replicate the command line parser used by the Microsoft compiler (in order to detect which include directories are used etc.).

I'd be happy to hear that `argparse` became more viable for `clcache`, but I suspect there's not a lot of benefit to reap.

YngveNPettersen:
Argparse makes it easier to add new command line arguments, especially if multiple arguments are can be used. In my case I also needed to add support for specifying the specific cl command to be used

frerich:
Closing this since there still doesn't seem to be a lot of need to change how argument parsing is done. Please don't hesitate to reopen this in case there's a concrete use case.

PR #346, opened 2019-02-17
Add support for command line specification of compiler executable, and support for clang-cl

YngveNPettersen:
This PR adds support for specifying the path to the compiler on the command line, similar to what is supported by ccache. See https://github.com/frerich/clcache/issues/270

The presumed compiler path is checked to make sure it ends with ".exe", otherwise it is assumed to be a compiler argument, with the compiler path either being cl.exe found by searching the path, or from the environment variable.

Specifying the compiler path on the command line several two major benefits:

* The project can use multiple compilers https://github.com/frerich/clcache/issues/343
* The project can use functionality already present for ccache, e.g as in the Chromium project
* The developer does not need to change the PATH or the environment variable when changing between projects that use a special compiler, e.g the Chromium project, which downloads different clang-cl versions for each branch.

Additionally, this PR adds support for a couple of clang-cl command line options and optimizes the use of the list of options, and adds a couple of new gitignore entries


clcache/__main__.py
```diff
@@ -1527,46 +1530,78 @@ def createManifestEntry(manifestHash, includePaths):
...
+                             action="store_true",
+                             help="reset cache statistics")
+    groupParser.add_argument("-M", "--set-size", dest="cache_size", type=int,
+                             default=None,
+                             help="set maximum cache size (in bytes)")
+
+    # This argument need to be optional, or it will be required for the status commands above
+    parser.add_argument("compiler", default=None, action=CommandCheckAction,
+                        nargs="?",
+                        help="Optional path to compile exectuable. If not "
```
frerich: Typo here: "exectuable".

clcache/__main__.py
```diff
@@ -1527,46 +1530,78 @@ def createManifestEntry(manifestHash, includePaths):
...
+            setattr(namespace, self.dest, values)
+
+    class RemainderSetAction(argparse.Action):
+        def __call__(self, parser, namespace, values, optional_string=None):
+            nonCommand = getattr(namespace, "non_command", None)
+            if nonCommand:
+                values.insert(0, nonCommand)
+            setattr(namespace, self.dest, values)
+
+    parser = argparse.ArgumentParser(description="clcache.py v" + VERSION)
```
frerich: I see that you scratched your own itch (#269 ), too. That's the spirit. :-)

I was quite skeptical about `argparse` actually working as intended, remembering the strange handling of quotation marks on the Windows command line. I seemed to recall that the argument parsing done by `argparse` was not bug-compatible (ahem) with what cl.exe ecpects. However, seeing that the tests still pass, maybe it's not so bad after all, phew! :-)

It doesn't seem to be a big improvement in terms of lines of code saved though, so I wonder -- any particular reason you started to use `argparse` here?

clcache/__main__.py
```diff
@@ -1527,46 +1530,78 @@ def createManifestEntry(manifestHash, includePaths):
...
-        return 0
+    # These Argparse Actions are necessary because the first commandline
+    # argument, the compiler executable path, is optional, and the argparse
+    # class does not support conditional selection of positional arguments.
+    # Therefore, these classes check the candidate path, and if it is not an
+    # executable, stores it in the namespace as a special variable, and
+    # the compiler argument Action then prepends it to its list of arguments
+    class CommandCheckAction(argparse.Action):
+        def __call__(self, parser, namespace, values, optional_string=None):
+            if values and not values.endswith(".exe"):
```
frerich: Should the test for `.exe` maybe be case-insensitive (e.g. by making `values` all lower-case first) given that file names on Windows are typically case-insensitive?

frerich (commented):
Looks good to me! Besides the smaller items which I mentioned in the in-line comments, I guess some mentioning of

* `clang-cl` now being supported
* The possibility of specifying a custom compiler command

In the `CHANGELOG.md` file would be good.

Also, the new optional custom compiler command should get mentionmed in the `README.asciidoc` file (alongside the other command line arguments), I suppose.

Now that I think about it t -- I guess a nice Wiki page which demonstrates how tot use clcache with different compilers with specific example invocations would be cool, too. This is convenient to have as a reference to point people to. :-)

clcache/__main__.py
YngveNPettersen: We have been using argparse this way since I started using clcache, although the organization was a bit different; the combination into one is just a few days old. If there were any issues with processing cl (or other commands') command line arguments, aside from what we hit wit clang-cl (which wasn't an argparse issue) we would probably have seen it long ago.

Argparse may, in this case, have some issues with space compared to the old way, but it does provide a standard way of specifying parameters (including that the query options are mutually exclusive), datatypes, and help. Additionally, with this organization it may be possible to add extra command line arguments, in addition to the optional compiler path I added. It may also be possible to add processing of the compiler arguments (what is done in the command line analyzer) , but I haven't investigated that, not sure it is worth the extra work.

clcache/__main__.py
YngveNPettersen: A point

clcache/__main__.py
frerich: Fair enough! I don't mind using `argparse` at all really (it just never seemed important enough for me to tackle it myself) as long as it doesn't break stuff. I was just curious whether maybe there's some other benefit I didn't notice.

The command line analyzer is a good point -- I think somebody _did_ try to use `argparser` at first but the noticed that it might be easier to reverse-engineer to peculiar command line handling of `cl.exe` manually, and that then became the command line analyzer.

YngveNPettersen:
I have locally done the suggested changes on our own production branch, but will need to pick them over to this branch and push them.

Regarding a Wiki page, that might be an idea. I don't see a good fit with the current pages, though, so maybe you should add a general page for running clcache using the current configuration, and then I can add text later, when this PR has been merged?

frerich:
I also don't know a good place to add such a page, but I was occasionally thinking that -- given how many ways there are to install, integrate and run clcache -- some sort of 'How To' section would be good where people could be referred to for things like 'How to use clcache with different compiler executables' or so.

YngveNPettersen:
That could be a way to do it.

BTW, I'll squash down the commits later before you merge

clcache/__main__.py
YngveNPettersen: Done

frerich:
Merged, thanks a lot!

This resolves #343 .

PR #348, opened 2019-02-18
Add CLCACHE_RECACHE feature

oktal3700:
Just like in ccache, this puts clcache into a "write-only" mode.
Objects will be stored in the cache, but objects already in the cache
will not be reused. The real compiler is always used.

This is useful in CI when you have a release branch that you want to
always use the real compiler, and feature branches that you want to
reap the benefit of the cache generated from the release branch.
In that case you would enable CLCACHE_RECACHE only when building from
the release branch.


frerich:
Looks good to me -- just a test seems to be missing. :-)

oktal3700:
Added a testcase.

PR #349, opened 2019-02-21
Support CLCACHE_BASEDIR in nodirect mode

oktal3700:
Do a simple case-insensitive find-and-replace to transform absolute
paths into relative paths within the preprocessor output that is used
to compute the hash. This makes CLCACHE_NODIRECT mode usable in the
presence of the `__FILE__` macro.

Fixes a bug in _normalizedCommandLine that caused the source filename
to be included in the hash computation.

### DIFF
--- a/clcache/__main__.py
+++ b/clcache/__main__.py
@@ -469,6 +469,11 @@ class CompilerArtifactsRepository:
         compilerHash = getCompilerHash(compilerBinary)
         normalizedCmdLine = CompilerArtifactsRepository._normalizedCommandLine(commandLine)
 
+        if "CLCACHE_BASEDIR" in os.environ:
+            baseDir = normalizeBaseDir(os.environ["CLCACHE_BASEDIR"]).replace("\\", "\\\\").encode("UTF-8")
+            newBaseDir = BASEDIR_REPLACEMENT.encode("UTF-8")
+            preprocessedSourceCode = re.sub(re.escape(baseDir), newBaseDir, preprocessedSourceCode, flags=re.IGNORECASE)
+
         h = HashAlgorithm()
         h.update(compilerHash.encode("UTF-8"))
         h.update(' '.join(normalizedCmdLine).encode("UTF-8"))
@@ -495,7 +500,7 @@ class CompilerArtifactsRepository:
         argsToStrip += ("MP",)
 
         return [arg for arg in cmdline
-                if not (arg[0] in "/-" and arg[1:].startswith(argsToStrip))]
+                if arg[0] in "/-" and not arg[1:].startswith(argsToStrip)]
 
 class CacheFileStrategy:
     def __init__(self, cacheDirectory=None):
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -86,10 +86,11 @@ CLCACHE_NODIRECT::
     key. Use this if you experience problems with direct mode or if you need
     built-in macroses like \__TIME__ to work correctly.
 CLCACHE_BASEDIR::
-    Has effect only when direct mode is on. Set this to path to root directory
-    of your project. This allows clcache to cache relative paths, so if you
-    move your project to different directory, clcache will produce cache hits as
-    before.
+    Set this to path to root directory of your project. In direct mode, this allows
+    clcache to cache relative paths, so if you move your project to different directory,
+    clcache will produce cache hits as before. When direct mode is disabled, clcache will
+    translate any absolute paths in the preprocessor output into relative paths before
+    computing the hash.
 CLCACHE_OBJECT_CACHE_TIMEOUT_MS::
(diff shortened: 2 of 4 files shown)


oktal3700:
Not sure what caused the build to fail.

clcache/__main__.py
```diff
@@ -118,6 +118,19 @@ def normalizeBaseDir(baseDir):
...
+    output = b''
+    index = 0
+    while True:
+        nextIndex = dataLower.find(oldLower, index)
+        if nextIndex < 0:
+            return output + data[index:]
+        output += data[index:nextIndex] + new
+        index = nextIndex + len(old)
+
+
```
frerich: I'm not sure you need a dedicated function for this -- maybe it would be viable to just use `re.sub` here?  I think `caseInsensitiveReplace` is basically

```python
re.sub(re.escape(old), lambda _: new, data, flags=re.IGNORECASE)
```

clcache/__main__.py
```diff
@@ -469,6 +482,11 @@ def computeKeyNodirect(compilerBinary, commandLine, environment):
         compilerHash = getCompilerHash(compilerBinary)
         normalizedCmdLine = CompilerArtifactsRepository._normalizedCommandLine(commandLine)
 
+        if "CLCACHE_BASEDIR" in os.environ:
+            baseDir = normalizeBaseDir(os.environ["CLCACHE_BASEDIR"]).replace("\\", "\\\\").encode("UTF-8")
```
frerich, line 473: Hmm, why are backslashes escaped here?

clcache/__main__.py
```diff
@@ -495,7 +513,7 @@ def _normalizedCommandLine(cmdline):
         argsToStrip += ("MP",)
 
         return [arg for arg in cmdline
-                if not (arg[0] in "/-" and arg[1:].startswith(argsToStrip))]
+                if arg[0] in "/-" and not arg[1:].startswith(argsToStrip)]
```
frerich, line 503: Hmm, did you mean to change the behaviour here? I suspect the `and` should become `or` here. [De Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) found that `not (a and b)` is equivalent to `not a or not b`.

That being said, I'm not sure the new version is any nicer than the old one, so maybe it's not worth touching this at all.

tests/test_unit.py
```diff
@@ -109,6 +109,13 @@ def testFilesBeneathRecursive(self):
             self.assertIn(r".\d\4.txt", files)
             self.assertIn(r".\d\e\5.txt", files)
 
+    def testCaseInsensitiveReplace(self):
+        replace = lambda s: clcache.caseInsensitiveReplace(s, b"PaTtErN", b"replacement")
+        self.assertEqual(replace(b"pattern may appear"), b"replacement may appear")
+        self.assertEqual(replace(b"infix paTTErn embed"), b"infix replacement embed")
+        self.assertEqual(replace(b"Pattern and PATTERN and pattern"), b"replacement and replacement and replacement")
+        self.assertEqual(replace(b"and finally PATTERN"), b"and finally replacement")
+
```
frerich: I suppose that in case you decide to drop the custom `caseInsensitiveReplace` function in favor of `re.sub`, you could also drop this test.

frerich (commented):
I think this looks like a plausible improvement, thanks a lot! However, it also invalidates all existing cache entries for people who do _not_ use the direct mode, i.e. I guess this shouldn't be part of a bugfix release but rather of a feature release.

clcache/__main__.py
frerich: Oh, I only now noticed that you _meant_ to fix a bug here. In that case, can you please add a test case which demonstrates the bug?

clcache/__main__.py
oktal3700: Rightly or wrongly, for simple things I personally like to avoid regex for performance reasons. But I'm a beginner in Python so if you prefer regex here I'm happy to oblige. And `re.escape` should protect it from any shenanigans.

clcache/__main__.py
oktal3700: It's to match the backslashes that are escaped in the preprocessor output because `__FILE__` expands to a C string literal.

clcache/__main__.py
oktal3700: `testBasedirNoDirect` would fail without this bugfix, but I can add a test that exercises _only_ this bugfix and none of the other changes, if you like.

oktal3700:
> it also invalidates all existing cache entries for people who do _not_ use the direct mode

Are you sure? Only `computeKeyNodirect` is affected. The bugfix part will invalidate cache entries for people who do not use `BASEDIR` though, is that what you meant?

oktal3700:
Applied suggested changes.

